PR #651, opened 2021-08-16
Update Readme.md: docker-compose fix

glitchcore:
# What's new

Changes wrong `docker compose`  to `docker-compose` command in build instructions.

# Verification 

Run build instructions from Readme.

# Checklist (do not modify)

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -67,13 +67,13 @@ One liner: `./flash_core1_main.sh`
    ```
 3. Prepare the container:
    ```sh
-   docker compose up -d
+   docker-compose up -d
    ```
 
 ## Compile bootloader
 
 ```sh
-docker compose exec dev make -j$(nproc) -C bootloader TARGET=f6
+docker-compose exec dev make -j$(nproc) -C bootloader TARGET=f6
 ```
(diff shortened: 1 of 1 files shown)


ReadMe.md
```diff
@@ -67,13 +67,13 @@ One liner: `./flash_core1_main.sh`
    ```
 3. Prepare the container:
    ```sh
-   docker compose up -d
+   docker-compose up -d
```
skotopes, line 70: Hi ) 

https://docs.docker.com/engine/reference/commandline/compose/

Old versions does not have it?

glitchcore: Hey!

Nice, docker now includes compose as a sub-command! But it works only since 3.4.0 as [Compose V2](https://docs.docker.com/compose/cli-command/). `docker-compose` seems more compatible.

anna-oake: I used `docker compose` intentionally to popularize new Docker versions. 

Will reconsider this.

glitchcore: Ok, nice point. If docker has this feature now, I think my proposal no longer needed.

PR #653, opened 2021-08-17
docker-compose linting and a good practice fix

nemanjan00:
# What's new

I just fixed linting issues for docker-compose.yml and made sure there is `apt upgrade` before any `adb install` commands, to make sure there is no version mismatch for new packages and dependencies.

# Verification 

- Run `docker-compose up`

- Build Dockerfile

# Checklist (do not modify)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,3 +1,4 @@
+---
 version: '3'
 services:
   dev:
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,6 +1,8 @@
 FROM ubuntu:focal
 
-RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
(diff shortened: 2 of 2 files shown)


anna-oake:
Hi! Thank you for your PR.

- We prefer `DEBIAN_FRONTEND=noninteractive` as an env var right before the command, because we don't need it anywhere else in the container
- As for the `apt upgrade`, we don't do this since it would increase image size and break the immutability with no benefits. Please check Docker best practices

nemanjan00:
Yes, `apt upgrade` increases size of image, but, you can end up with broken image if you do not run it before installing anything...

What can happen is you install latest version of something and end up with up to date package and out of date dependencies (worst case scenario, you end up with incompatible version of libc)...

Using `DEBIAN_FRONTEND=noninteractive` as env directly in command works and reduces build by one step (and saves a few kb), but, it is not a great idea as it creates space for mistakes further down the line

PR #655, opened 2021-08-17
[FL-1472, FL-1596, FL-1673] IRDA: stability improvements

albkharisov:
# What's new

- Restrict with 31 bytes length for remote and signal name
- Don't stuck for 0 PWM cycle timings
- Support timings > 65535 PWM cycles
- Fix remote file open error
- Add IRDA TX debug redirect
- Add remote parse error print, improve parsing, support tabs
- Fix stucks with uncorrect RAW signal values, long strings in remote file, etc
- Fix HAL signals capturing (save previous read value)
- Fix leak in case of failed parsing

# Verification 

- try to rename button, remote with long name
- try to send long and short timings on any frequencies in RAW format
- try different signals saving to file and send them
- check leaks in all error cases

# Checklist (do not modify)


applications/irda/irda-app-file-parser.cpp
```diff
@@ -195,34 +222,81 @@ const char* find_first_not_of(const char* str, char symbol) {
     return str;
 }
 
+static int remove_args32(std::string_view& str, size_t num) {
```
ValentiWorkLearning, line 215: There is no reason for passing sting_view by reference. See https://stackoverflow.com/questions/58410793/passing-stdstring-view-by-reference discussion

applications/irda/irda-app-file-parser.cpp
```diff
@@ -306,13 +399,15 @@ bool IrdaAppFileParser::check_errors() {
 }
 
 std::string IrdaAppFileParser::file_select(const char* selected) {
-    TextStore* filename_ts = new TextStore(128);
+    TextStore* filename_ts = new TextStore(IrdaAppRemoteManager::max_remote_name_length);
```
ValentiWorkLearning: Probably it's better to use std::unique_ptr there. There is no reason to use a raw C pointer. Also, this simplifies the mamory management here( delete will be called automatically)

applications/irda/irda-app-remote-manager.cpp
```diff
@@ -11,6 +12,8 @@
 #include "irda-app-file-parser.h"
 
 static const std::string default_remote_name = "remote";
+uint32_t const IrdaAppRemoteManager::max_button_name_length = 31;
```
ValentiWorkLearning: `inline constexpr` is more suitable here. See more detailed at the article: https://habr.com/en/company/jugru/blog/506104/

applications/irda/irda-app.cpp
```diff
@@ -48,6 +48,18 @@ int32_t IrdaApp::run(void* args) {
...
+IrdaApp::IrdaApp() {
+    furi_check(IrdaAppRemoteManager::max_button_name_length < get_text_store_size());
+    notification = static_cast<NotificationApp*>(furi_record_open("notification"));
+    irda_worker = irda_worker_alloc();
+}
+
+IrdaApp::~IrdaApp() {
+    irda_worker_free(irda_worker);
+    furi_record_close("notification");
+    for(auto& it : scenes) delete it.second;
```
ValentiWorkLearning: For range-based for loops you can use structured binding, like this:
```cpp
for(auto&[key,scene] : scenes)
    delete scene;
```
Also, it should be considerable to use the smart pointers there.

albkharisov: Didn't know that staff. Thank you for reviewing!

applications/irda/irda-app-file-parser.cpp
albkharisov: I thought about it for a while and came to this:
I understand that passing by ref has no profit in speed/size as far as sv is a pointer+size.
But I want to return length of removed symbols, so passing by ref is valid for this case.

There can be another approach - pass by value, and then move sv data ptr (remove_prefix) by value of returned value.
But this seems little bit odd as we move this ptr inside function, and then move it again by remove_prefix outside.

So for this case I think this is a valid function signature.

skotopes (approved):

PR #691, opened 2021-09-08
Fix: Correct the instruction file docker-compose command.

DF-wu:
# What's new
- I noticed the command `docker compose` is wrong in ReadMe.md. So I correct it to `docker-compose` which is right command to run.
- I am sorry to not able to finish the last task in CheckList because I don't have the device to upload and test(I am not firmware delevoper but yes I hava built it.)
- This modification should not effect the source code of firmware.

# Verification 
- ~~My experience~~
- The offical docker docs
  - [Overview of docker-compose](https://docs.docker.com/compose/)
- The result of current version is executed.
  - 圖片
- The correct version
  - 圖片

# Checklist (do not modify)


anna-oake:
It is not wrong, it just relies on the new Docker Desktop versions which have compose included as a `docker` subcommand.
Check this out:
Screenshot 2021-09-08 at 21 43 53

It's really strange that they did not include it in normal Docker Engine yet.
Will reconsider this, thanks.

DF-wu:
> 
> 
> It is not wrong, it just relies on the new Docker Desktop versions which have compose included as a `docker` subcommand.
> Check this out:
> Screenshot 2021-09-08 at 21 43 53
> 
> It's really strange that they did not include it in normal Docker Engine yet.
> Will reconsider this, thanks.

@aprosvetova 
Thank you for replying.  I understand now.
I should mention my platform(Arch x86-64) and docker engine version before.
- 圖片

Maybe a note in the Readme.md  is better.  I may not the first guy though it was weird :)

anna-oake:
Let's keep this open.

I'll research a bit more about this new integrated Compose.
If it's not going to be added to the normal Engine soon, I will merge this.

skotopes:
@aprosvetova switching PR to draft till your research complete.

anna-oake:
Well, looks like I was talking about [Docker Compose V2](https://docs.docker.com/compose/cli-command/), which is really cool and much better than old `docker-compose`, but is currently in a RC state and is included in Docker Desktop only.

**It is better to reference plain old `docker-compose` in the docs for now**, but I will keep an eye on the Compose V2 and revert the change as soon as it becomes a thing.

Thanks a lot, @DF-wu!

anna-oake (approved):

skotopes (approved):

PR #699, opened 2021-09-11
Improved reliability of shell scripts

rusdacent:
# What's new

- Added quotes for variables in shell script. Fixed problem with spaces in directories names and params.
- Added exit for cd command. Prevented run commands after incorrect change directory.

# Verification 

- Run changed scripts

# Checklist (do not modify)

### DIFF
--- a/docker/syntax_check.sh
+++ b/docker/syntax_check.sh
@@ -6,7 +6,7 @@ CLANG_FORMAT_BIN="/usr/bin/clang-format-12"
 
 PROJECT_DIR=$(pwd)
 
-cd $PROJECT_DIR
+cd "$PROJECT_DIR"
 
 echo "RUN C\C++ SYNTAX CHECK"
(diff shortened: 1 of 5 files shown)


docker/syntax_check.sh
```diff
@@ -6,7 +6,7 @@ CLANG_FORMAT_BIN="/usr/bin/clang-format-12"
 
 PROJECT_DIR=$(pwd)
 
-cd $PROJECT_DIR
+cd "$PROJECT_DIR" || exit
```
skotopes, line 9: why?

rusdacent: Prevent next commands run in script after incorrect change directory.

skotopes: why not `exit 255` in that case?

skotopes (approved):

PR #717, opened 2021-09-22
FL-1862 add return codes to ob.py

lomalkin:
# What's new

- Added ob.py tool return codes tor testing jigs purposes

# Verification 

Check this in these cases: (stlink connected, stlink disconnected, target disconnected from stlink)
- `./ob.py check; echo $?`
- `./ob.py set; echo $?`

Return code must be 0 only in case of operation success.

# Checklist (do not modify)


DrZlo13 (approved):
Works

PR #857, opened 2021-11-30
Fix readme typos

okalachev:
There is also a lot of markdown format issues (like multiple first-level headers, etc) in readme files, I can fix them as well if you'd like.


DrZlo13 (approved):
Nice

okalachev:
Should I address mentioned markdown issues?

PR #883, opened 2021-12-08
fix "Nothing to be done for upload" error

vvzvlad:
If this line is present, when you run the "make -C firmware upload" command a second time (repeatedly, after the first firmware), it gives an error "Nothing to be done for upload", which is not exactly what is expected from the firmware command (its behavior should not depend on the past runs). See https://t.me/flipperzero/3009

# What's new

- [ Describe changes here ]

# Verification 

# Checklist (do not modify)


skotopes:
Hi,

Actually that is exactly how I've designed it.

How about discussing things in issues before doing pull requests?

PR #884, opened 2021-12-08
add "reboot_to_dfu" and "reboot_to_app" cmd to make

vvzvlad:
Adds two commands to reboot into the DFU and to reboot from the DFU into the main application in the makefile. The command is supposed to be used like this:
male -C firmware reboot_to_dfu upload reboot_to_app.
I am not sure if reboot_to_dfu will work correctly on all machines, but I have not found any other way to get the flipper port address. Also, this command has not been tested on Linux.

# What's new

- [ Describe changes here ]

# Verification 

# Checklist (do not modify)

### DIFF
--- a/make/rules.mk
+++ b/make/rules.mk
@@ -84,6 +84,15 @@ flash: $(OBJ_DIR)/flash
 
 upload: $(OBJ_DIR)/upload
 
+reboot_to_dfu:
+	stty -f /dev/cu.usbmodemflip_* 115200
+	echo -en "dfu\r\n" > /dev/cu.usbmodemflip_*
+
+reboot_to_app:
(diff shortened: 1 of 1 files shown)


anna-oake:
Definitely won’t work on Linux and Windows, will also go crazy with multiple Flippers connected :(

make/rules.mk
```diff
@@ -84,6 +84,15 @@ flash: $(OBJ_DIR)/flash
 
 upload: $(OBJ_DIR)/upload
 
+reboot_to_dfu:
+	stty -f /dev/cu.usbmodemflip_* 115200
```
skotopes, line 88: It is either cross platform or it's not here.

Just keep a bash script for this purpose

PR #885, opened 2021-12-08
add vscode config for dev and debug

vvzvlad:
Suggestion: A few files for easy debugging in VS Code and a few settings for working on code.
Additionally, a non-boring coloring of the development environment in flipper colors. Additionally, need install ms-vscode.cpptools, marus25.cortex-debug, ms-vscode.makefile-tools packets

# What's new

- [ Describe changes here ]

# Verification 

# Checklist (do not modify)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+  "cortex-debug.armToolchainPath": "/usr/local/bin/",
+  "cortex-debug.openocdPath": "/usr/local/bin/openocd",
+  "cortex-debug.gdbPath": "/usr/local/bin/arm-none-eabi-gdb",
+}
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,25 @@
+{
+  "version": "2.0.0",
+  "tasks": [
+    {
+      "label": "Build",
+      "type": "shell",
+      "group": "build",
+      "command": "make",
+      "problemMatcher": []
+    },
+    {
+      "label": "Clean",
+      "type": "shell",
+      "group": "build",
+      "command": "make clean",
+      "problemMatcher": []
+    },
+    {
+      "label": "Write firmware",
+      "type": "shell",
+      "command": "make firmware_flash",
+      "problemMatcher": []
+    }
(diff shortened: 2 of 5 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "cortex-debug.armToolchainPath": "/usr/local/bin/",
+  "cortex-debug.openocdPath": "/usr/local/bin/openocd",
```
skotopes, line 3: Same issues, this is not going to be usable outside of mac os.

We've been talking about it a lot and decided not to add vscode solution if it is not working on all platforms.

skotopes:
We are open to contribution, but they must work for everyone.
If it's targeting only one platform and broken on the other, then it's not going to be merged.

Please join discussion in issues or our telegram/discord channel before doing pull requests like this one.
We value your and our time and the best choice is to discuss changes with us before implementing them.

Thank you for your understanding.

DrZlo13:
@vvzvlad 
Note that we shouldn't add IDE settings to the main repository. This will break the settings for other people who have different view point.
For example, my settings are fundamentally different from this pull request.

.vscode/settings.json
okalachev: Before it supports Linux and Windows it could be named Upload & Run Flipper (macOS), or may be implemented only for macOS, VSCode launch-files allow that. And to show something like 'not implemented yet' on other platforms.

Having ready shortcut for debugging in macOS is much better than not having any shortcuts at all. It's not easy to configure VSCode to debug this.

.vscode/tasks.json
```diff
@@ -0,0 +1,25 @@
+{
```
okalachev, line 1: This file would work on all the platforms, why not to merge that.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "cortex-debug.armToolchainPath": "/usr/local/bin/",
```
okalachev, line 2: On my machine debugging works without these settings. Maybe to remove them and this way it'll become cross-platform?

PR #889, opened 2021-12-10
Create .editorconfig file

okalachev:
[EditorConfig file](https://editorconfig.org) allows IDEs (like VSCode) to use correct indentation settings without configuration.

As for me, I'm strongly for using tabs in C/C++ for indentation (reasons: https://habr.com/ru/post/118208/). But if the project is using spaces already and doesn't plan to switch to tabs, it would be much more convenient to setup VSCode automatically for all new and existing files (instead of switching to spaces manually, like I had to do now).

Note: not all the project file comply with `insert_final_newline` rule for now.


anna-oake:
We use clang-format. It’d be pain in the ass to keep multiple configurations in sync.

There is also a clang-format VS Code plugin that loads the config file and uses it to reformat files on every save.

Could you please try installing the clang-format plugin?

okalachev:
These are about different things.

EditorConfig is very basic and about initial code editor settings with you open any file in a project. It's actually mostly about spaces/tabs setting. It works almost in all the code editors without any plugins (even on GitHub site when viewing the code!).

clang-format is about complex C++ code formatting rules. it works when you hit 'format code' and requires plugins.

EditorConfig is a really popular file and doesn't conflict with clang-format.

anna-oake:
Well, I see your point. It might be easy enough to keep this file updated and in sync with the clang-format rules.

skotopes:
Do I understand right that VSCode, SublimeText, Vim doesn't support this config without plugins?
What editors do support it?

okalachev:
VSCode support without plugins was my error. The list of editors supporting without plugins: https://editorconfig.org/#pre-installed.

skotopes (approved):

PR #898, opened 2021-12-14
I2C Scanner app

okalachev:
Hi! I made and tested a graphical I2C scanner app for Flipper.

I guess having such small debugging tools for common embedded things can make Flipper a very useful tool for developers and "hackers". No need to run an RPi with SSH access from a computer to do a quick testing of an I2C device.

For now the tool scans the I2C bus and shows the list of addresses responded:

   

If you’re ok with such an idea I can add:

* Detecting device types using a database (like https://i2cdevices.org/addresses,  https://github.com/romkey/ESP-Diagnostic-Tool/blob/master/src/i2c_scanner_devices.cpp).
* GUI for reading and writing device registers (with appropriate warnings about possible harm). Basically repeat the functionality of `i2c-tools` for Linux.
* Take out the scanning procedure to a separate thread (now the app is one-threaded for the sake of code simplicity).

# Checklist (do not modify)


anna-oake:
Looks like this belongs to the GPIO app.

okalachev:
So, should I transfer this thing into the GPIO app?

DrZlo13:
Since we already have the ability to load applications from the SD card, I would prefer that this application remain separate. It doesn't look like basic Flipper functionality.

applications/debug_tools/i2c_scanner.c
```diff
@@ -0,0 +1,101 @@
...
+#include <gui/modules/submenu.h>
+#include <gui/gui.h>
+
+static const char header[] = "I2C C0=SCL C1=SDA";
+
+typedef struct {
+    Gui* gui;
+    ViewDispatcher* view_dispatcher;
+    Submenu* submenu;
+    bool device_found[128];
```
skotopes, line 14: We actually have m-lib for C containers: it got arrays and `string_t` that suites perfectly for this task.

skotopes (changes requested):

okalachev:
@DrZlo13, it depends on what Flipper is. If it's only doors opener/rfid/nfc/radio cracker, then maybe. If it's more wide embedded "hacking" tool, then, for me, showing the list of I2C devices connected is a very basic feature, as basic as:

* UART monitor
* USB-UART adapter
* PWM generator (for testing servos, motors...)
* GPIO pins manual control

etc

DrZlo13:
@okalachev, I didn't make myself clear enough. The app is definitely useful and deserves to be included in our repository, but for now we are not going to add it to the firmware releases.

Space on the Flipper flash is very limited, and we are already experiencing a shortage of it. As soon as it becomes possible to download the application from the SD card, we will add this application to the release build.

To summarize, we can accept the pull request, but in order for it to appear in other Flippers, they will have to enable this application separately and rebuild the firmware.

okalachev:
So what should I do?

Set the `APP_I2C_SCANNER` flag to zero and/or transfer the app from debug tools to somewhere else?

By the way, the idea that it should belong to GPIO app does make sense, on my opinion.

okalachev:
@DrZlo13?

DrZlo13:
@okalachev 
Set the APP_I2C_SCANNER flag to zero and make the changes requested by @skotopes.

skotopes:
This PR is marked as inactive and will be automatically closed in 7 days.

PR #993, opened 2022-02-14
Fix missing dependency in Brewfile

Astrrra:
# What's new

`protobuf-c` dependency added to brewfile

# Verification 

`brew bundle --verbose` on a clean machine should now install `protobuf-c` and protobuf should compile properly.

# Checklist (do not modify)


skotopes:
Looks like `protobuf` is enough.

PR #1014, opened 2022-03-10
Add NTAG215 as possible device type

devn:
I noticed that `NTAG215` was missing as a possible device type in the list spit out by the writer. I've added it here.

### DIFF
--- a/applications/nfc/nfc_device.c
+++ b/applications/nfc/nfc_device.c
@@ -250,7 +250,7 @@ static bool nfc_device_save_file(
         if(!flipper_format_write_header_cstr(file, nfc_file_header, nfc_file_version)) break;
         // Write nfc device type
         if(!flipper_format_write_comment_cstr(
-               file, "Nfc device type can be UID, Mifare Ultralight, Bank card"))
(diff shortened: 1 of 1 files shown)


skotopes:
@gornekich waiting for your review

anna-oake:
There are also NTAG213 and NTAG216. Almost the same, just different memory size. I'd label it as "NTAG" or "NTAG21x" instead.

devn:
> There are also NTAG213 and NTAG216. Almost the same, just different memory size. I'd label it as "NTAG" or "NTAG21x" instead.

Would you be ok with me listing all 3? To me, it reads as list of possible choices. NTAG21x would look to me like that is the value I should enter as opposed to being specific about NTAG213, NTAG216, etc.

zhovner:
Useless commit.

devn:
> Useless commit.

As a n00b, it was not to me. Having the enumerated list of options was useful as I tried to craft a .nfc in the correct format.

I'll take your comment as a sign I shouldn't spend any additional effort trying to contribute here.

Thanks for saving me some time!

corasaurus-hex:
> Useless commit.

Useless comment.

Issue #1020, opened 2022-03-14
3v3 and 5v GPIO Output Issue


MuddledBox:
Playing around with the GPIO today I’ve noticed a few oddities, either the GPIO commands via serial are non-functional or not working as expected, and providing odd functionality.

---

Scenario Name: Turn off 3v3 voltage via Serial
Flipper Zero Firmware: 0.50.0 0.50.0 (3f164ef3 built on 03-03-2022)
Test Setup: Multi-meter plugged into Pin 11 (GND) and Pin 9 (3v3), Flipper connected via USB-C
Additional Software Used: PuTTY
Repro Steps:

Plug in multi-meter to Pin 11 and Pin 9. Connect Flipper to computer and turn on. Open PuTTY and connect to relevant COM port. Type “power 3v3 1”. Observe voltage on multi-meter as 3.30v, as expected.

Next, type “power 3v3 0”. Observe voltage on multi-meter as 2.81v, not as expected. Expected result was 0.00v reading on multi-meter.

---

Scenario Name: Turn off 5v voltage via Serial
Flipper Zero Firmware: 0.50.0 0.50.0 (3f164ef3 built on 03-03-2022)
Test Setup: Multi-meter plugged into Pin 11 (GND) and Pin 1 (5v), Flipper connected via USB-C
Additional Software Used: PuTTY
Repro Steps:

Plug in multi-meter to Pin 11 and Pin 1. Connect Flipper to computer and turn on. Open PuTTY and connect to relevant COM port. Type “power 5v 1”. Observe voltage on multi-meter as 4.96v-5.04v, as expected (within variance).

Next, type “power 5v 0”. Observe voltage on multi-meter as 4.96v-5.04v, not as expected. Expected result was 0.00v reading on multi-meter.

---

In summary, clearly something wrong here. For sanity, I also tested via the “5V on GPIO” functionality under GPIO and again, GPIO voltage was always on, never went to 0.00v as expected. I also noted some variance from the 3v3 pin where voltage reached up to 5.26v briefly, but I’m not sure why that would be or how to recreate.

skotopes:
3.3V: this line is used to power SD-Card it is highly unrecommended to turn this line off(power will be re-enabled to scan sdcard). We will remove this command in future releases.
5V: it is expected behaviour when USB connected. Also using iButton will enable 5V while reading keys.

MuddledBox:
Makes sense, thanks!

PR #1022, opened 2022-03-14
Removed power 3v3 options

MuddledBox:
# What's new

- Per skotopes, "3.3V: this line is used to power SD-Card it is highly unrecommended to turn this line off(power will be re-enabled to scan sdcard). We will remove this command in future releases." - #1020 

So I removed it here. No longer an option and should not be possible to switch on and off 3v3 via the power command. 

# Verification 

- Tested via PuTTY. Option no longer available and multi-meter does not report voltage changes when command is executed (as command is no longer valid). 

puttyTest

# Checklist (do not modify)

### DIFF
--- a/applications/power/power_cli.c
+++ b/applications/power/power_cli.c
@@ -34,6 +34,16 @@ void power_cli_5v(Cli* cli, string_t args) {
     }
 }
 
+void power_cli_3v3(Cli* cli, string_t args) {
+    if(!string_cmp(args, "0")) {
+        furi_hal_power_disable_external_3_3v();
+    } else if(!string_cmp(args, "1")) {
+        furi_hal_power_enable_external_3_3v();
+    } else {
+        cli_print_usage("power_ext", "<1|0>", string_get_cstr(args));
+    }
+}
+
 static void power_cli_command_print_usage() {
     printf("Usage:\r\n");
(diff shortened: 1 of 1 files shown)


DrZlo13:
We need to consult with QA, there is a chance they used this command for production testing.

DrZlo13:
> We need to consult with QA, there is a chance they used this command for production testing.

OK, they dont.

skotopes:
Thanks for PR.

One small thing: we have "system debug" flag that we use when do factory tests.
I think it's ok to hide 3.3 control behind this flag so it will be still available for factory testing.

MuddledBox:
That sounds like a more elegant solution! Where is the flag and example of its usage?

skotopes:
https://github.com/flipperdevices/flipperzero-firmware/blob/3b8b2e59df8bdaac9956ae7521a91ca1f39da705/applications/subghz/subghz_cli.c#L598

MuddledBox:
@skotopes Most recent commit now implemented the way you requested. Tested and working via PuTTY, when debug on working as expected, when debug off working as expected. Help menu changes too so no confusion from non-debug users.

skotopes (dismissed):

skotopes:
Please run `make format` to reformat sources

skotopes (approved):

PR #1024, opened 2022-03-14
nfc: add basic Mifare DESFire support

kevinwallace:
# What's new

- Detection of DESFire cards based on ATQA/SAK
- Reading of all non-protected card data
- UI to browse apps/files in card data
- Fixes to TextBox wrap behavior
- Adds `bool` type to flipper_format
- Saving and loading DESFire tag data

# Verification 

- Open NFC app and scan a DESFire card.  Here is an example of me scanning an ORCA (Seattle public transit) card and briefly browsing its data.

flipper_desfire_orca

https://user-images.githubusercontent.com/184975/158492956-21d614c1-8e55-452a-bb48-48a1f1f53f81.mov

# Checklist (do not modify)


kevinwallace:
Updated to include saving/loading of DESFire tag data.  Let me know if the flipper_format bool change is unwanted -- I'm happy to drop that commit and rewrite to use hex for those fields instead.

skotopes:
Hi @kevinwallace 

Great amazing job, you are awesome! )

Adding bool is ok. @gornekich will review PR and provide feedback.

kevinwallace:
UI has changed a bit since the original video, here's the current state of things

https://user-images.githubusercontent.com/184975/158708890-6bd59e9b-79b8-4e40-94dd-44851dc0bdf6.mov

(and thank **you** for Flipper -- this has been such a fun device to hack on!  :)

applications/nfc/nfc.c
```diff
@@ -163,6 +163,8 @@ int32_t nfc_app(void* p) {
     if((*args != '\0') && nfc_device_load(nfc->dev, p)) {
         if(nfc->dev->format == NfcDeviceSaveFormatMifareUl) {
             scene_manager_next_scene(nfc->scene_manager, NfcSceneEmulateMifareUl);
+        } else if(nfc->dev->format == NfcDeviceSaveFormatMifareDesfire) {
+            scene_manager_next_scene(nfc->scene_manager, NfcSceneReadMifareDesfireSuccess);
```
gornekich: Now we start nfc application with parameters if we want to launch NFC app from Archive and start UID or Mifare Ultralight emulation immediately. In your case, we will see information about selected DESFire card. I am not sure that we want to break user's expectation here, so I would delete this for now

applications/nfc/nfc_worker.c
```diff
@@ -658,6 +670,242 @@ void nfc_worker_emulate_mifare_ul(NfcWorker* nfc_worker) {
...
+                    tx_len = mf_df_prepare_read_records(tx_buff, file->id, 0, 0);
+                    break;
+                }
+                err = nfc_exchange_full(tx_buff, tx_len, rx_buff, sizeof(rx_buff), &rx_len);
+                if(err != ERR_NONE) {
+                    FURI_LOG_W(TAG, "Bad exchange reading file %d, err: %d", file->id, err);
+                    continue;
+                }
+                if(!mf_df_parse_read_data_response(rx_buff, rx_len, file)) {
+                    FURI_LOG_W(TAG, "Bad response reading file %d\r\n", file->id);
```
gornekich: No need in \r\n here

applications/nfc/scenes/nfc_scene_mifare_desfire_app.c
```diff
@@ -0,0 +1,109 @@
...
+    view_dispatcher_send_custom_event(nfc->view_dispatcher, index);
+}
+
+void nfc_scene_mifare_desfire_app_on_enter(void* context) {
+    Nfc* nfc = (Nfc*)context;
+    Submenu* submenu = nfc->submenu;
+    MifareDesfireApplication* app = nfc_scene_mifare_desfire_app_get_app(nfc);
+    if(!app) {
+        FURI_LOG_E(TAG, "Bad state. No app selected?");
+        return;
```
gornekich, line 42: There is a risk of undesired behavior. If you return here, view dispatcher will display and deliver events to the view from previous scene and execute on_event() handler of MifareDesfireApp scene. I suggest switching to a Popup view with error explanation.

gornekich (changes requested):
Hello @kevinwallace !
Thank you very much for your amazing work!
So far, I looked through your code and left a few comments. I need more time to dive into your DESFire implementation more carefully, and make detailed review.
I noticed a memory leakage after I open DESFire card and then exit NFC application. You can see this from our log console (LoaderSrv logs) or executing 'free' CLI command before and after launching NFC app and opening DESFire card. Memory leakage is a blocking issue for merging your pull request.
I suggest you fix memory issues while I will make detailed review in next couple of days.

applications/nfc/nfc.c
kevinwallace: Done!  8b85ff3f854976834b8974ca390c51ce05d47f9d

applications/nfc/nfc_worker.c
kevinwallace: Oops!  This was originally a printf.  Done.  ff1fcbf55f11f6f88d93b84f9d6158491d25536a

applications/nfc/scenes/nfc_scene_mifare_desfire_app.c
kevinwallace: If this codepath ever runs, it's a bug in the MifareDesfireData scene.  But just in case I made a nice popup anyway.  :)  f5dbd6fd781180897596584a32b5dbacb6fd6de9

error_popup

kevinwallace:
Thank you for the comments!

A few hours ago, I noticed a memory leak while reviewing this PR's diff and fixed it in 56ab13eab0703e88ea7b7e83ff61cf10cca20a5f.  I think this was the same leak you saw -- I'm now checking with the `free` CLI command and `Free heap size` is remaining at the same size whenever I return to the desktop from the NFC app.  Could you let me know if you're still seeing this, and if there's a particular set of interactions to reproduce it?

applications/nfc/scenes/nfc_scene_saved_menu.c
```diff
@@ -62,7 +74,11 @@ bool nfc_scene_saved_menu_on_event(void* context, SceneManagerEvent event) {
             scene_manager_next_scene(nfc->scene_manager, NfcSceneDelete);
             consumed = true;
         } else if(event.event == SubmenuIndexInfo) {
-            scene_manager_next_scene(nfc->scene_manager, NfcSceneDeviceInfo);
+            if(nfc->dev->format == NfcDeviceSaveFormatMifareDesfire) {
+                scene_manager_next_scene(nfc->scene_manager, NfcSceneReadMifareDesfireSuccess);
```
gornekich: I think it's better to show DESFire cards Info in NfcSceneDeviceInfo, or make another NfcSceneMifareDesfireInfo scene. If I go to  NfcSceneReadMifareDesfireSuccess scene from **Saved cards**, I don't expect to **Name and save** card that I already opened

gornekich (commented):
Perfect! The leakage I mentioned is gone. Also, thanks for following my suggestions!

Another issue I found is that if I go to saved DESFire card -> Emulate UID -> Back -> Info, there is no info about saved card and after I exit NFC ~3kB of heap leak.
Current UID emulation supports displaying reader commands on the screen. The reason behind this is that we give users an opportunity to research what kind of reader in front of them. From technical point of view, we pass commands received in **nfc worker** to **emulate uid scene** through NfcReaderRequestData structure in NfcDeviceData, which shares memory with MifareDesfireData. That's where after UID emulation we loose DESFire data and can't free allocated memory.
The quickest fix here is to disable UID emulation for DESFire cards, just like it's done for Bank Cards. However, if you want to keep opportunity for UID emulation, you should fix NfcDeviceData structure  so that NfcReaderRequestData and MifareDesfireData don't share same memory.

Let me know how you want to fix UID emulation issue and when it's done, we will merge your brilliant PR :)

applications/nfc/scenes/nfc_scene_saved_menu.c
kevinwallace: Agreed that offering "Name and save" for an already-saved card is inappropriate.  I've added support for DESFire cards to `NfcSceneDeviceInfo` and now go there instead of `NfcSceneReadMifareDesfireSuccess` when reading a saved card.  The "Data" button goes to `NfcSceneMifareDesfireData`, and `NfcSceneMifareDesfireMenu` is now only accessible when scanning a new card.

bf643267394ed3ee8988eabc2af02c64d5b6aee4

device_info

I also tried another approach, keeping `NfcSceneReadMifareDesfireSuccess` in both cases, and showing the 'More' button selectively based on scene history.  It's in my `desfire_wip` branch (https://github.com/kevinwallace/flipperzero-firmware/commit/26ef6e03d5c82c66c64935c6e0665f923eb252b0, https://github.com/kevinwallace/flipperzero-firmware/commit/a42c938b6e27c3c952836fbeb20f93e10dab9fc8) if you'd like to take a look, but after trying both, I prefer the approach in this PR.

kevinwallace:
Good catch on the broken UID emulation behavior due to memory sharing -- I didn't look closely at the UID emulation code earlier to understand this was going on.  It seems useful to still be able to emulate UID only, so I've moved `NfcReaderRequestData` out of the union so that it no longer shares memory with DESFire data.  This means it's possible to safely emulate UID of EMV cards from the saved card menu, too, so I've also re-enabled it for those.

fcc1bfa728c716a12e948819df9edea033848cd6

I'm also fine disabling it and leaving `NfcReaderRequestData` in the union if you prefer that approach -- let me know and I'll revise.

applications/nfc/scenes/nfc_scene_saved_menu.c
gornekich: I like current approach with adding DESFire to NfcSceneDeviceInfo scene

gornekich:
Thanks for fixing UID emulation. Now everything looks good!

gornekich (approved):

skotopes (approved):

gornekich:
Thanks for contribution!

PR #1042, opened 2022-03-19
Keyboard enhance shift

MuddledBox:
# What's new

- Added ability to use capital letters with long press of "ok". 
- Short press is lower case, long press is upper case. 
ShiftFunction

# Verification 

- Tested by myself and other users. 

# Checklist (do not modify)


skotopes (approved):

DrZlo13:
Fast moving by pressing and holding keys is broken.

MuddledBox:
@DrZlo13 I would like to add that back in, along with a few keyboard enhancements like not auto-capitalizing the first letter, which now serves little purpose. Also want to change canvas when button is held to change all keys to caps.

Issue #1040, opened 2022-03-17
Feature Request: "Twin Duck" usb rubber ducky emulation


Patronics:
Many of the sample badusb scripts from Hak5 use the "Twin Duck" firmware, emulating a USB drive in addition to a keyboard input. This would add a lot more flexibility to the BadUSB feature.

MuddledBox:
This has been requested a bit in the Discord. 

Currently the SD card in the Flipper Zero is not recognized as a mass storage device by Windows. That would probably be a good step one before implementing this feature.

skotopes:
Theoretically possible. But going to be suuuuuuper slow.
We'll keep this thing in backlog, but don't expect it be implemented anytime soon.

Lucaslhm:
1+ to this. This is an important feature on the ducky and bash bunny that allows you to quickly download payloads off the usb onto the target machines, or take data off the target machine and store it on the drive, negating the need for internet.

Hope to see this implemented at some point!

Woodie4:
+1 This would be amazing... I really hope this will get implemented

TotallyNotSethP:
DuckyScript 3.0 transfers data by flashing the CapsLock and NumLock LEDs, meaning the OS wouldn't have to use the card as mass storage at all. Maybe you could implement it that way?

nuschpl:
IMHO The feature of masstorge is important not only from perspective of copying data from/to it. It is also some form of cover channel - one could delay execution of ducky script and pretend it's just a storage, or have an excuse when catched with device connected

PeterHindes:
This is a much needed feature for full badusb functionality. The current level of functionality can be achieved with a $5 arduino.

skotopes:
Storage subsystem went throw a bunch of optimization. Now we can say that there is nothing else we can do to improve it any more.

We still don't think that current performance is enough for normal work of twin ducky, but community can try to implement it. Could work for some cases.

nuschpl:
IMHO we shouldn't directly access Flipper SD card for "Twin Duck" mass storage purposes for Flipperr operator's own safety. IMHO we should emulate mass storage on some buffer with very limited size suitable for moving config files/powershell scripts back and forth. Then the buffer could be synced with flipper SDcard. Or what's about Flipper internal storage - is 100% occupied by firmware those days ?

calcdot:
[...]

PR #1060, opened 2022-03-26
USB Mass Storage emulation (DRAFT)

kevinwallace:
# What's new

I created this hoping that I would be able to put a bunch of bootable ISOs on a microSD card, then emulate and boot off of them using my Flipper Zero.  It turns out that it's WAY too slow for this to be usable (120~180kBps in my testing), but probably still fast enough for some uses of "Twin Duck" mode requested in #1040.  Presenting this as a draft PR to get people's thoughts on integrating it with BadUSB.

Open questions:
* "Mass Storage" app is a bit thrown together, mostly copied from BadUSB as a template.  Should this be completely integrated into BadUSB or should both apps be kept?
* USB stuff is currently implemented in `applications/mass_storage/helpers/` instead of `targets/f7/furi_hal/` like the other USB modes.  Should it be moved there?  SCSI code, too?
* Dual USB Mass Storage + HID mode.  How do we get them to play along?  It might be possible to throw together a combined `usb_device_descriptor` and set of `FuriHalUsbInterface` functions that delegate to existing unmodified HID and Mass Storage functions, or it might be more involved.  Haven't attempted this yet.
* Speed.  Can it be improved?  If the bottleneck is SD SPI, then probably no.  If the bottleneck is USB or FS code, maybe.

If someone is interested in taking this code, polishing it up, and integrating it with BadUSB -- or just using it as a starting point for their own independent implementation -- I would be thrilled.  My personal use case didn't pan out, and I'm not likely to need "Twin Duck" myself, but it would be a shame to throw this out since it could be useful to others, and given the trial, error, blood, sweat, and tears it took to get to this point.

Here I am booting Debian off of my Flipper.  GRUB took several minutes to load.  :)

mass_storage_boot_debian

# Verification 

* Create `mass_storage/` dir on SD card and place disk images in it.
* Open "Mass Storage" app (under plugins) and select your image
* Connect Flipper to computer and access over USB Mass Storage interface

# Checklist (For Reviewer)


DrZlo13:
Unfortunately, this is exactly the case, SD cards are very slow when connected via SPI, and the maximum theoretical speed is around 500 kbps. That is why we have abandoned the idea of mass storage emulation.

skotopes:
There are couple things that can improve speed:

- DMA for SPI 
- Read Cache

zhovner:
Maybe it can be usable for tiny stuff like one file linux https://github.com/zhovner/OneFileLinux

Woodie4:
+1 This would be amazing... I really hope this will get implemented

Foul:
The best would be to use Ventoy (https://www.ventoy.net) on it... but don't absolutely know if it's possible... (i'm just a user..)

kowalski7cc:
> The best would be to use Ventoy (https://www.ventoy.net) on it... but don't absolutely know if it's possible... (i'm just a user..)

For booting a full ISO it's too slow, but maybe could be just enough to boot rEFInd when your bootloader is broken.

Also maybe small payloads 🤔

hramrach:
I would say even if it's slow it can be useful.

Would be also nice if the data flipper collects could be accessed - that's typically small but might be useful to view from the PC.

turmoni:
Just as a potential use case for even very slow mass storage emulation: I stumbled across this PR when looking to see if the Flipper Zero might be able to do one of the things that I want, which is to scan an RFID card and present its ID as a text file over USB.

kevinwallace:
Closing, subsumed by https://github.com/flipperdevices/flipperzero-good-faps/pull/1

PR #1063, opened 2022-03-28
archive: bump MAX_FILES to 800

kevinwallace:
# What's new

- Increase Archive's directory entry limit from 100 to 800.  It seems that the OOM this limit guards against happens somewhere between `MAX_FILES=800` and `MAX_FILES=1000`, so we still can't remove the limit entirely.

# Verification 

- View a directory with many files or subdirectories in Archive, observe that truncation happens at 800 entries now, rather than 100.  <https://github.com/Spexivus/csv2ir>'s `ir/` directory is a good test case with 618 entries, and is what motivated me to make this change.

# Checklist (For Reviewer)

### DIFF
--- a/applications/archive/helpers/archive_files.h
+++ b/applications/archive/helpers/archive_files.h
@@ -1,7 +1,7 @@
 #pragma once
 #include "file_worker.h"
 
-#define MAX_FILES 100 //temp
+#define MAX_FILES 800 //temp
 
 typedef enum {
     ArchiveFileTypeIButton,


skotopes:
Hi @kevinwallace,

This solution is not solving original problem with storage listing, we can not accept this PR.
What is needed is archive refactoring that will change way how it stores view model data.

If you feel like your skill is enough to fix issue, then please continue this work we will happily help you and accept this contribution. Otherwise we will schedule this issue for next sprint (starting this Wednesday).

kevinwallace:
Understood, thanks!  I'm not familiar enough with the view model code to commit to the proper fix, so I'm closing this PR for now.  In the meantime I'll just organize things in directories of <100 entries.

Issue #1019, opened 2022-03-13
Hard crash reading Monzo beta card (NFC) (MasterCard EMV)


hakusaro:
**Describe the bug**
I ran into a hard crash when attempting to read my Monzo beta card (via NFC). I have a video of this, but it's as straightforward as these things go: attempting to read the card as a bank card crashes with `furi_check failed`.

This happens on version 0.50.0 [03-03-2022] / 3f164ef3 [1103], 100% of the time with this card.

I turned on debugging and debug logs set to info level, but I didn't see any logs generated in any obvious places on the SD card or anywhere that I could find accessible in the menus itself. If you give me instructions, I would be more than happy to provide more logs.

Also, out of the 5 or so cards I've tried, this is the only card that hard crashes like this. I haven't had success running the EMV apps from all cards, and it seems like MasterCard is the most unreliable. Visa seems fine.

Video: https://streamable.com/lsc4dk

jman1118:
I have the same issue when reading cards from Apple Pay.

hakusaro:
@jman1118 I don’t actually have this problem reading Apple Pay cards from Visa. I haven’t tried MasterCard though. What’s the issuer?

jman1118:
Any of my physical mastercards didnt read on the first try. Earlier apple pay locked up the flipper but havent been able to reproduce this.

jman1118:
My apple card doesnt read at all but that could be because its metal.

gteague:
apple card i get a popup with nfc-a as the first line and then it refuses to go any further. /guy

hakusaro:
The physical Apple Card doesn't have an NFC chip inside it, iirc.

gteague:
ah! i was trying about a half-dozen cards and only the visa worked as i expected. afterwards i tried reading an air-tag just for s&g's and that was the one which only responded to the /read/ option and said it was a bank card, but wouldn't go any further. and the /bank card/ read option didn't work. sorry but my short term memory isn't what it used to be. and i know zero about this stuff to begin with. 

[later note: confirmed my mistake. using nfc module /read/ with the apple airtag results in a popup starting: 'nfc-a may be: emv bank card ....'.  if you then select /run compatible app/ the light flashes blue and it says 'reading bank card' and it never stops. 

i'm very sorry for injecting confusion into the discussion. carry on.]

/guy

hakusaro:
Fwiw, this still happens on `0.51.0` / f2a84851.

[...]

PR #1211, opened 2022-05-10
Fix for issue #1019, HardFault on certain types of credit/bank cards.

SnowLeopard71:
# What's new

Changed the index increment from 2 to 1 on line 295 in emv.c
The country_code does not always have 2 extra pad bytes in the SFI record and it is before the card number instead of after in some card responses, so incrementing the index by 2 in emv_decode_read_sfi_record() was causing miss the CC number.

Truncated SFI record:      70 81 8A 5F 25 03 20 10 01 **5F 24** 03 24 08 31 **5F 28** 02 01 24 **5A 08** 53 38 66

# Verification 

I tested card that was causing crash previously (BIN 533866).  Also tested all other cards and they still work as before the change. 

# Checklist (For Reviewer)


skotopes:
@gornekich please check

gornekich (approved):

skotopes (approved):

PR #1227, opened 2022-05-13
RFC: NTAG I2C support

GMMan:
This PR is mostly to gather comments about whether the proposed architecture of NTAG I2C storage would be suitable for the project. There are gaps between some pages in NTAG I2C, so to save on space, a tag sector/page address to linear address converter is used so all the pages are packed together. Aside from that, the implementation generally works.

# What's new

- Read and emulation support for NTAG I2C 1K/2K, NTAG I2C Plus 1K/2K
- Support for NTAG I2C-specific commands
- Read/write security for NTAG I2C Plus 1K/2K, includes support for NFC_PROT, 2K_PROT, NFC_DIS_SEC1, page blanking and unauthenticated NAK on read, unauthenticated NAK on write
- SRAM support **not** implemented

# Verification 

- Read of unprotected NTAG I2C (Plus) 1K/2K tags should succeed and report the correct tag type
- Emulation of tag should result in successful read by NXP NFC TagInfo full scan
- Emulated protected tags should:
  - NAK if read starts on page specified by AUTH0
  - Regular read and fast read should have pages starting from AUTH0 blanked out with null bytes
  - PWD and PACK should be null when read out
  - Password auth should NAK if wrong password (except when tag data has zeros for PWD, continuing with current implemented behavior for other NTAGs), return PACK and allow reads/writes if successful.
  - Password protected sectors should return NAK if writing and NFC_PROT is 0
  - Entire sector 1 should NAK on read/write if NFC_DIS_SEC1 set
  - AUTLIM behavior is **not** implemented

# Checklist (For Reviewer)


gornekich:
Hello @GMMan! Thanks for your pull request!
We are waiting for i2c NTAG to be delivered to verify your feature.
I noticed that you didn't add support for saving and loading the NTAG types you introduced. Could you make it, like it's done for other types?

firmware/targets/f7/furi_hal/furi_hal_nfc.c
```diff
@@ -334,6 +335,9 @@ bool furi_hal_nfc_emulate_nfca(
                     break;
                 }
                 if(buff_tx_len) {
+                    if (buff_tx_len == UINT16_MAX)
+                        buff_tx_len = 0;
```
gornekich: Why do you need this?

gornekich (changes requested):
Your changes break NTAG21x read functionality with message:

5824 [D][MfUltralight]: Selecting sector 255
5876 [D][FuriHalNfc]: Timeout during data exchange
5877 [D][MfUltralight]: Failed to issue sector select command
5878 [D][NfcWorker]: Failed reading Mifare Ultralight

Can you fix it?

Also, please, use `make format` to run linter

GMMan:
> I noticed that you didn't add support for saving and loading the NTAG types you introduced. Could you make it, like it's done for other types?

I think it's present? I added the tag types to `nfc_types.c` and currently have some saved NTAG I2Cs that I can load, view, and emulate.

firmware/targets/f7/furi_hal/furi_hal_nfc.c
GMMan: This is to support emulating the `SECTOR_SELECT` command. After the sector number is sent in the command's second packet, the tag ACKs by not replying. It's bizarre, but that's how NXP implemented it. A normal `buff_tx_len = 0` terminates the emulation loop, so I needed a different value to indicate 0.

image

gornekich (approved):

skotopes (approved):

Issue #1186, opened 2022-04-29
Automatic desktop locking with PIN code


digitalcircuit:
**Scenario**
The [automatic desktop locking feature](https://github.com/flipperdevices/flipperzero-firmware/pull/1101 ) works well for avoiding accidental button presses when in one's pocket, etc.  However, it's not very effective at reducing risk if someone else finds the device (lost, etc).

**Describe the solution you'd like**
An additional setting (e.g. `Settings` → `Desktop` → `PIN Setup` → `PIN Lock After`) that locks the Flipper with one's PIN code after a configurable delay.

For example, with `Auto Lock Time: 60s` and `PIN Lock After: 10min`, the Flipper will use the triple-back lock after 60 seconds, then after 10 minutes, it'll lock with the PIN Code (also disabling USB).

**Describe alternatives you've considered**
* Replacing (instead of augmenting) the triple-back lock with PIN code lock
   * In `Settings` → `Desktop` → `PIN Setup` → `Auto Lock w/PIN` as an `ON`/`OFF`, or similar
   * ℹ This might make more sense (especially if auto lock will only ever work while viewing the desktop)
   * ➕ Simple, straightforward
   * ➖ Makes lock time setting a trade-off between guarding against accidental button presses and time it takes to input one's PIN code
* Adding an Auto Lock mode setting (specifying the 3-back lock, PIN code, or DUMB mode)
  * In `Settings` → `Desktop` → `Lock Mode` with `3x Back`, `PIN`, `DUMB`, or similar
  * ➕ Flexible, allows for automatically disguising the Flipper as a normal game device
  * ➖ Potentially more work, DUMB mode is not yet implemented

**Additional context**
To try to reduce confusion, `PIN Lock After` should probably not be allowed to be set to to a shorter delay than `Auto Lock Time` unless `Auto Lock Time` is set to `OFF`.

It might make more sense to move these settings out to a `Security` category if the screen locking feature can be applied system-wide (instead of only happening when on the desktop).  Or, `PIN Lock After` may be placed beneath `Auto Lock Time` so it's possible to see and control both automatic locking options from the same place.

I realize the Flipper's not designed to be completely secure against physical access, but I think this might help with situations where it's been left or lost and someone with good intentions wishes to return the device.

A separate feature request might be optional text to show on the lock screen (e.g. `Found: 123-555-6789`), similar to how Android allows specifying a lock screen message - I can file a new issue for that.

zhovner:
[...]

PR #1244, opened 2022-05-19
Proper Lock With Pin if Pin is set

RogueMaster:
# What's new

- [ Describe changes here ]

# Verification 

# Checklist (For Reviewer)

### DIFF
--- a/applications/desktop/desktop.c
+++ b/applications/desktop/desktop.c
@@ -56,7 +56,12 @@ static bool desktop_custom_event_callback(void* context, uint32_t event) {
         return true;
     case DesktopGlobalAutoLock:
         if(!loader_is_locked(desktop->loader)) {
-            desktop_lock(desktop);
+            if(desktop->settings.pin_code.length > 0) {
+                desktop_pin_lock(&desktop->settings);
+                desktop_lock(desktop);
+            } else {
+                desktop_lock(desktop);
+            }
         }
         return true;
     }


RogueMaster:
#1186 will be resolved with this I think

skotopes:
Breaks qFlipper / Cli+VCP

skotopes:
It's actually quite complex tasks. The reason why we keep current behaviour is because it requires a lot more complex system to manage state transitions and special cases (update).

RogueMaster:
qFlipper / Cli+VCP doesn't have access to the locked device. Makes sense. Wouldn't this have been done intentionally? Is there a status that can be checked to see if there is an active qFlipper or Cli+VCP connection?

skotopes:
> qFlipper / Cli+VCP doesn't have access to the locked device. Makes sense. Wouldn't this have been done intentionally? Is there a status that can be checked to see if there is an active qFlipper or Cli+VCP connection?

nope and that was the main stopper

RogueMaster:
Is there something that can be checked to see if it is plugged into USB?

RogueMaster:
Could disable locking while charging perhaps...  if(power->state == PowerStateCharging)

skotopes:
> Is there something that can be checked to see if it is plugged into USB?

yes, but that's not that simple. We have locking refactoring in backlog. Basically it need to be redone with fixes for all special cases.

If you want to do it, then start from diagramming current locking scheme, then ping me for review, then I can help you we redesigning it and splitting into coding tasks.

skotopes:
There were a bug in locking, fixed in https://github.com/flipperdevices/flipperzero-firmware/pull/1245 please check if it works for you

RogueMaster:
> > Is there something that can be checked to see if it is plugged into USB?
> 
> yes, but that's not that simple. We have locking refactoring in backlog. Basically it need to be redone with fixes for all special cases.
> 
> If you want to do it, then start from diagramming current locking scheme, then ping me for review, then I can help you we redesigning it and splitting into coding tasks.

I'm not sure how it needs to be refactored, or why. Sorry. I think I'll need more understanding of the right calls in Flipper for this. I've been searching around the code for examples on everything I've worked on. I'll try to add a power state check for my auto-lock on pin. Let me know if I should submit those updates

PR #1261, opened 2022-05-21
RFID: add support for Kantech IOProx cards

ktemkin:
# What's new

- Adds supporting for reading, writing, and emulating Kantech IoProx 125kHz RFID cards; as commonly used in e.g. building access control systems.  

# Verification 

If you have an IOProx card:

- Use _Read_ from the _125kHz RFID_ menu, and read an IoProx tag.
- Use `rfid read` from the CLI, and read an IoProx tag.

If you have a T55xx card:

- Use _Add Manually_ from the _125kHz RFID_ menu, and enter in a valid code, such as `3B01AEB5`.
- Select _Emulate_, and validate that the emulated tag is readable from e.g. another flipper.
- Select _Write_, and write to a T55xx card.
- Use _Read_ from the _125kHz RFID_ menu, and read the card.

Repeat with the CLI:

- Use `rfid emulate IoProx 3B01AEB5` and validate that the emulated tag is readable from e.g. another flipper.
- Use `rfid write IoProx 3B01AEB5` to write to a T55xx card.
- Use `rfid read` to read from the card, and verify that it reads as an IoProx card with the given code.

# Checklist (For Reviewer)


applications/lfrfid/helpers/decoder_ioprox.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+constexpr uint32_t jitter_time_us = 20;
+constexpr uint32_t min_time_us = 64;
+constexpr uint32_t max_time_us = 80;
+
+constexpr uint32_t min_time = (min_time_us - jitter_time_us) * clocks_in_us;
+constexpr uint32_t mid_time = ((max_time_us - min_time_us) / 2 + min_time_us) * clocks_in_us;
+constexpr uint32_t max_time = (max_time_us + jitter_time_us) * clocks_in_us;
+
+constexpr uint32_t baud_time = 32000;
```
DrZlo13: Please express the baud rate in something meaningful, magic numbers are not welcome.

applications/lfrfid/helpers/decoder_ioprox.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+    // Always track the time that's gone by.
+    current_period_duration += time;
+    demodulation_sample_duration += time;
+
+    // If a baud time has elapsed, capture a sample.
+    if(demodulation_sample_duration >= baud_time) {
+        store_data(current_demodulated_value);
+        demodulation_sample_duration = 0;
+    }
```
DrZlo13, line 50: Picks up everything and theoretically leads to false positive because we don't check if it's actually FSK modulation like in DecoderHID26 (it's the same modulation as far as I know).

applications/lfrfid/helpers/key_info.h
```diff
@@ -8,6 +8,7 @@ enum class LfrfidKeyType : uint8_t {
     KeyEM4100,
     KeyH10301,
     KeyI40134,
+    KeyIoProx, // TODO: figure out the right codename for this
```
DrZlo13: i suggest KeyIOProxXSF (as in Proxmark)

DrZlo13 (changes requested):

applications/lfrfid/helpers/decoder_ioprox.cpp
ktemkin: Fixed in the most recent patch-set.

applications/lfrfid/helpers/key_info.h
ktemkin: Sure -- done.

applications/lfrfid/helpers/decoder_ioprox.cpp
ktemkin: I can see how this would be an issue in terms of wasting CPU time in the detection loop. (It wouldn't lead to false positives unless the thing that was picked up happened to both have pulse trains with the correct periods _and_ managed a correct packet checksum.)

I've adjusted it so values are only stored if there's valid FSK, so should be fixed.

> (it's the same modulation as far as I know).

This is the same modulation as HID26; but with different sample-timing requirements / a different baud-rate. This *should* hopefully be equivalent, but with a slightly more correct clock-recovery behavior, and the ability to tweak the baud clock by changing a single const. 

(I don't think the slight increase in clock-recovery correctness really actually _matters_, here; but it's a bit nicer given the ioProx tags have way lower transition density than e.g. most HID tags.)

ktemkin:
Review comments should be addressed, now. :)

skotopes:
Also please avoid force push, it's difficult to review changes between iterations

DrZlo13 (approved):

DrZlo13:
Have you tested the emulation? How? It doesn't work for me with flipper > proxmark or flipper > flipper. Reading and writing works OK.

ktemkin:
> Have you tested the emulation? How? It doesn't work for me with flipper > proxmark or flipper > flipper. Reading and writing works OK.

I have – both with a Proxmark3 and with several of the actual ioProx readers. (I don't have a second Flipper to test flipper > flipper, with.)
 
image
image

It's curious that you're not seeing similar results.

DrZlo13:
Rechecked, my proxmark seems to be faulty. F2F emulation works but requires some positioning trickery.

PR #1267, opened 2022-05-23
Fix MiFare DESFire GetKeySettings response parsing

kevinwallace:
Only the bottom nybble of buf[1] is the key ID; the top nybble contains flags.
Notably, the top bit is set for AES mode on EV1+ cards, and ORCA cards in the
Seattle area were just upgraded to include an app that uses AES.

Prior to this, cards with flags set could be read and saved just fine, but
loading them would fail due to missing keys.  After this, tags saved with the
old version will load fine, and when re-saved will have the flags separated
out into a new field.

# What's new

- Fixes loading of saved DESFire tags

# Verification 

- Read a DESFire card that has flags set in GetKeySettings, save tag, load saved tag.

# Checklist (For Reviewer)


gornekich (approved):
Thanks!

skotopes (approved):

PR #1279, opened 2022-05-27
desktop: prompt to power off on back button long press

### LINKED ISSUE #1276: Power off on return long press
**Is your feature request related to a problem? Please describe.**
Sometimes I need to quickly power off the device.

**Describe the solution you'd like**
It would be nice to use long press on return button to power off device. Could you implement behaviour like on old mobile phones or mp3 players? I see long press on return is not used on main screen.

kevinwallace:
Fixes #1276.  Even though idle power usage is impressively low, it's nonzero, and I find myself powering my Flipper Zero off somewhat often.  This would make that easier.  Re-uses existing "power off" scene from power settings app.

# What's new

- desktop: prompt to power off on back button long press

# Verification 

- Hold back button at desktop. Verify it powers off.
- Verify that power settings app continues to work as before when launched normally.

# Checklist (For Reviewer)


azaitsev:
Thank you for prompt PR. Can we put flipper to sleep without confirmation if we do that from desktop? If we run the shutdown from the desktop, then we do not risk losing any data, and the put into sleep itself will not affect work in any way. I think we can do it without confirmation.

kevinwallace:
Given that boot is fairly quick in the case of an accidental power-off, that wouldn't bother me, and it simplifies the overall diff considerably.  I've removed the prompt in a second commit.  I'll leave it up to the maintainers whether it's best to merge with one or both commits, or not at all.

kevinwallace:
After using this without a prompt for the past few days, I've accidentally caused a poweroff more often than I've intentionally caused one.  I suggest that this is merged with a prompt -- reverted b712350ac94cabd15e43058c6db3e1d53a538dad, sorry @azaitsev.  But leaving it in git history for now so that all options are on the table.

azaitsev:
How about to rework power off process to the following one:

1. Long press to initiate power off
2. Screen shows a prompt: "Device will be powered off in 3... 2... 1..." with counter
3. Continue long press
4. Power off

azaitsev:
I also couldn't find the timing for InputTypeLong in the firmware. If it's the same delay as the one used to view a firmware version by holding the key down, then it's garbage. It might be worth adding an InputTypeExtraLong type that would count as a long press for more than 2 or 3 seconds.

azaitsev:
It looks like long press defined here:
[#define INPUT_LONG_PRESS_COUNTS 2](https://github.com/flipperdevices/flipperzero-firmware/blob/7395caa7ce043b68dd761fc8904fa54cdc9b91a8/applications/input/input_i.h#L19)
I thing we can change that value to 4 or 6 for InputTypeExtraLong.

kevinwallace:
Another option would be to wait for a specific number of `InputTypeRepeat` events within desktop, rather than modifying the input system.  In any case, it would be more complex than it is now, and we haven't heard from maintainers yet -- going to wait before making further changes to this PR.

azaitsev:
Hm, yes, counting of InputTypeRepeat is a better solution. Ok, let's wait for maintainer's opinion.

azaitsev:
Perhaps we should think about redefining INPUT_LONG_PRESS_COUNTS at the system level, because right now it is not a long press at all. In the future, this long press could be used to exit the application to the home screen and for other purposes, so it should be long enough to eliminate accidental presses.

skotopes:
We'll merge this PR after 0.60 release

skotopes:
Long press delay is too short for such action. Is it ok if I merge current version and then rewrite delay part?

skotopes:
@azaitsev please fix merge conflict

skotopes:
Also instead of setting_primary_favorite you can use scene data. In second favourite

kevinwallace:
Merge conflict handled in 73d37192b7d7043b5af61c5f81ca4cfb1c9c7998.  I think the comment about scene data is more applicable to #1307, this PR doesn't modify desktop settings.

skotopes:
oh sorry, I saw name in comment and didn't check original PR creator )

skotopes (approved):

skotopes:
i'm going to merge it as is and then slightly modify logic in main view to generate event on long delays (5 second)

PR #1304, opened 2022-06-09
Increased long press interval

azaitsev:
# What's new

I've increased long-press interval to match convension of Android and Arduino:

https://arduinogetstarted.com/tutorials/arduino-button-long-press-short-press
https://stackoverflow.com/questions/1930895/how-long-is-the-event-onlongpress-in-the-android

# Verification 

Before testing this PR, check default interval: open the desktop app and press the down button to enter the archive app. Now press and hold that down button to see firmware version. You will see that short press (archive) and long press (fw version) are about the same.
Now you can test my changes and see that the long press is more separated.

# Checklist (For Reviewer)


skotopes:
Hi. Flipper got its own UX, we are not orienting on android(totally different input method) or arduino(because there is no such standard at all).

azaitsev:
The goal is to make UX better, not to just make as android. Current long press isn't long and almost has no difference to short press. @skotopes

skotopes:
I've brought this topic internally, we've checked timings one more time and most of us think that current option is the best for hardware keyboard.

azaitsev:
You have a difference between short and long press in about 150ms, its negligible. When @kevinwallace have been testing deep sleep pull request he accidentally powered off flipper multiple times.

Your team thinks the current option is the best just because long press is almost not used in flipper yet.

skotopes:
@azaitsev Long press (more precise long press + repeat) is used almost everywhere. Most important part is keyboard and navigation. 

I'm not aware about what exactly @kevinwallace has been testing. Also deep sleep in its current for is highly experimental thing. I can investigate if you provide more details.

kevinwallace:
I think @azaitsev is referring to this comment I made: https://github.com/flipperdevices/flipperzero-firmware/pull/1279#issuecomment-1141516945

The typical situation was that I held down the back button to return to desktop from an app, and failed to let go because I had switched attention away from Flipper -- I'm not sure a longer press interval would have saved me in that case.  I'm happy with the current system-wide long-press timing.

PR #1269, opened 2022-05-24
[FL-2052] New build system based on scons

hedger:
# What's new

- [FL-2052] New build process 

# Verification 

- Install Python 3.8+ as system's default
- Install Python requirements - pip3 install -r scripts/requirements.txt
- Read docs/fbt.md & try building in different modes

# Checklist (For Reviewer)

- [x] PR has description of feature/bug or link to Confluence/Jira task
- [x] Description contains actions to verify feature/bugfix
- [x] I've built this code, uploaded it to the device and verified feature/bugfix

[FL-2052]: https://flipperzero.atlassian.net/browse/FL-2052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1313, opened 2022-06-13
nfc: NTAG21x complete emulation

GMMan:
# What's new

This is a complete implementation of NTAG21x emulation, including quirks and datasheet non-conformance from a genuine NTAG216.

- Authentication mechanism: PWD_AUTH, negative authentication limit, read permission check, write permission check, associated NAKs when checks fail
- Page locking: capability container, static locks, dynamic locks, OTP behavior, NAK on trying to write to locked page, NAK on trying to write locked config pages
- ASCII mirror: UID mirror, counter mirror, UID+counter mirror, associated odd behaviors depending on whether counter is enabled and where the mirror starts
- Read counter: self-incrementing
- Compatibility write
- Read roll-over, including when encountering AUTH0 with PROT set while unauthenticated
- Mifare Ultralight EV1: VCSL command
- Mifare Ultralight + EV1: counter behavior
- Refactoring, including unified feature flags and config pages access

# Verification 

- Emulation should behave in accordance with NTAG21x datasheet. For any deviations, compare behavior with genuine NTAG21x. Timing was not considered for implementation, so that may be off.
- Examples of datasheet deviations:
  - ASCII mirror ranges for UID+counter in datasheet do not apply; instead, UID and counter ranges are evaluated separately, so e.g. if one of UID or counter mirror will fit within the user memory pages, it will be mirrored, while the other won't.
  - ASCII mirror max `MIRROR_BYTE` bits is `10b`, not `01b` as documented in the datasheet
  - Negative verification counter can be reset when `AUTHLIM` is set to 0 and `PWD_AUTH` is successful. This is not documented in the datasheet.
- Notes on edge case experiments [here](https://discord.com/channels/740930220399525928/954422716138676254/984989907292389437)
- Known deviations from actual NTAG21x:
  - `0h` `NAK` response from NTAG21x appears to not be 4 bits of zeroes, but either 0 bits, or something with invalid parity. I don't have a Proxmark so I can't tell which it is. Flipper will respond with 4 bits of zeroes with correct parity.
  - Current emulation framework does not tell handler about field detect status, so there is no way to simulate a power on reset aside from waiting for the 5 second RX timeout. That means if you're trying to increment the read counter, it will only go up if RX timeout occurs.

# Checklist (For Reviewer)


RogueMaster:
Confirmed Amiibo can now register for Super Smash Bros, be trained and be saved back. Tested multiple times. The SHD file can be renamed to NFC to save alternate NFC file at trained state.

gornekich:
Great work!
We will merge it after #1269 , when we have extra free FLASH memory

gornekich (approved):

skotopes (approved):

skotopes:
Good Job!

PR #1319, opened 2022-06-14
nfc: On-device tag generator

GMMan:
# What's new

- Generate Mifare Ultralight tags from the NFC menu
- Supported tag types:
  - Mifare Ultralight (original)
  - Mifare Ultralight EV1
  - NTAG21x
  - NTAG I2C
  - NTAG I2C Plus

# Verification 

- From NFC menu, select "Generate/Add Manually". Select tag type, and a tag of the type with a random UID should be displayed, with option to save or emulate in the "More" menu.
- Tag content should conform to the described content of the corresponding datasheet

# Checklist (For Reviewer)


gornekich:
Hi @GMMan !
Could you explain possible user case for this feature?
Now we think that card generation and modification should be done with mobile app

GMMan:
Usually when people want to create new tags on Flipper, they have to go find a blank tag to scan as a base or download one and trannsfer to Flipper, and after that write with an NFC app. With on-device generation, this would be unnecessary, and the user can go directly to writing to the emulated tag. Having tag generation occur off-device is probably not ideal because the app generating would both have to know Flipper's file format and the correct content of an empty tag, and could become out of sync with the tag types Flipper supports.

applications/nfc/helpers/nfc_generators.c
```diff
@@ -0,0 +1,327 @@
...
+
+static const NfcGenerator mf_ul_generator = {
+    "Mifare Ultralight",
+    nfc_generate_mf_ul_orig,
+    NfcSceneReadMifareUlSuccess};
+
+static const NfcGenerator mf_ul_11_generator = {
+    "Mifare Ultralight EV1 11",
+    nfc_generate_mf_ul_11,
+    NfcSceneReadMifareUlSuccess};
```
skotopes: Success scenes are not intended to be used that way

skotopes: also please specify what exactly you are assigning like `.name="qweqwe"`

applications/nfc/scenes/nfc_scene_start.c
```diff
@@ -27,6 +28,8 @@ void nfc_scene_start_on_enter(void* context) {
         nfc_scene_start_submenu_callback,
         nfc);
     submenu_add_item(submenu, "Saved", SubmenuIndexSaved, nfc_scene_start_submenu_callback, nfc);
+    submenu_add_item(
```
skotopes: Please reuse "Add Manually" menu

applications/nfc/scenes/nfc_scene_generate.c
```diff
@@ -0,0 +1,50 @@
...
+    Nfc* nfc = context;
+    bool consumed = false;
+
+    if(event.type == SceneManagerEventTypeCustom) {
+        nfc_device_clear(nfc->dev);
+        const NfcGenerator* generator = nfc_generators[event.event];
+        generator->generator_func(&nfc->dev->dev_data);
+
+        scene_manager_set_scene_state(nfc->scene_manager, NfcSceneGenerate, event.event);
+        scene_manager_next_scene(nfc->scene_manager, generator->next_scene);
```
skotopes: Please don't reuse success scenes. Their UX is not  for that.

skotopes (changes requested):

GMMan:
Regarding success scenes, what is the suggested route? Success scenes already contain everything one might want after generation, namely an easy way to immediately emulate so the new tag can be written to before saving. The generators are intended to act as a virtual read of a new card and fill out all the data that would otherwise be captured during a full card read. The back navigation from the success scene appears to be working fine. Not having to reload the card to write to it should also prevent an additional .shd file from being created for each generated card, since the main use case is to write new data to the tag after generation.

GMMan:
Pinging @skotopes for further feedback

gornekich:
Hello @GMMan !
We dissed with our UX design team and we come to this "Add manually" flow:
image
You can igrore "NDEF records" and "Write" option for now, because we don't have this functionality yet.

The main issues with read_success scenes are the unnecessary notification and "Retry" option, which is not really clear in this context.

It would be perfect if you rework you PR according to the attached design. Or I can do it, since I am reworking NFC design right now. Let me know your decision.

GMMan:
I think I got it modified according to the flow diagram.

GMMan:
I didn't put tag info in the menu since it's just redirecting to the tag-specific menu, is that OK?

applications/nfc/scenes/nfc_scene_start.c
```diff
@@ -28,7 +29,11 @@ void nfc_scene_start_on_enter(void* context) {
         nfc);
     submenu_add_item(submenu, "Saved", SubmenuIndexSaved, nfc_scene_start_submenu_callback, nfc);
     submenu_add_item(
-        submenu, "Add Manually", SubmenuIndexAddManualy, nfc_scene_start_submenu_callback, nfc);
+        submenu,
+        "Generate/Add Manually",
```
gornekich: I am not sure that Generate is necessary here. From common user's point of view these two are quite similar, I would leave just "Add Manually"

gornekich (commented):
Great work, thanks for the changes!
I think skipping tag info is OK
Let's discuss "Generate" in main menu, before we merge you PR

applications/nfc/scenes/nfc_scene_start.c
GMMan: "Add manually" seems to imply to me I might have to do a bit of work to get a file made, which isn't really the case when generating. How about "Create", or "Create new"?

gornekich: Talked with our UX/UI team, we came to the decision that we should leave just "Add manually"... May be we change this later, but for now let's keep just like it's now in dev
Please, make this change and we will merge you PR

gornekich (approved):
Thanks!

skotopes (approved):

PR #1358, opened 2022-07-03
Mifare Classic INC/DEC/Restore/Transfer commands

Astrrra:
# What's new

- Value block commands for more accurate Mifare Classic emualtion

# Verification 

- Can be verified with a proxmark3 by installing the latest master firmware and running something like `hf mf value --blk 4 --k FFFFFFFFFFFF --inc 1` on a dump that has value blocks (can be created using the MifareClassicTool)

# Checklist (For Reviewer)


xMasterX:
There is some file that maybe came from macOS docker build - "null.d"
I think it should not be here, same as the second one named "usbinstall.flag"

https://github.com/flipperdevices/flipperzero-firmware/pull/1358/files#diff-0db09ba13c32adb1f938a27de5b2145e16c5a31f9e81254da217bf02e557e234

PR #1365, opened 2022-07-04
Mifare Ultralight authentication

ezhevita:
# What's new

- Adds authentication for Mifare Ultralight tags and some algorithms to generate password based on UID of the tag.

Supported algorithms:
- Amiibo
- Xiaomi Purifier
- Default (0xFFFFFFFF)

# Verification 

- Read an ultralight tag with password generated by one of the algorithms (Amiibo is fine) - it should be read successfully

# Checklist (For Reviewer)


lib/nfc_protocols/mifare_ultralight.c
```diff
@@ -492,14 +559,53 @@ bool mf_ultralight_read_pages(
...
+            break;
+        }
+
+        if(tx_rx->rx_bits == 4 && tx_rx->rx_data[0] >> 4 == 0x00) {
+            // This page is locked by a password
+            i -= 4;
+
+            uint32_t key;
+            bool key_found = false;
+            for(size_t j = 0; j < (sizeof(pwd_gens) / sizeof(int32_t)); j++) {
```
gornekich: Please use COUNT_OF from common_defines.h

gornekich (commented):
Good work! What do you think about cards with unsuccessful password verifications limit set?
Here is the part from datasheet:
image
So I guess there is a risk of bricking the card, which will not allow authentication after limitted number of unsuccessful password verification. That's why I don't think that it's right to continuously try different password in a loop.
When I was thinking to add this for writing tags, the best decision I came to is to check if card has set auth limits and if not - try different passwords in the loop. However this is not applicable for reading routine.
Another possible solution it to get user's confirmation each time we want to auth with password. But that seems to be bad UX.
What do you think about it? May be @GMMan has some thoughts to share?

GMMan:
Exceeding negative auth limit counter is definitely a risk. If you want to try authenticating, best start with positively identifying which tag type it is, figure out where its config pages are, and read AUTH0 and AUTHLIM, skipping authentication if AUTH0 is past the last page. Reference NXP NFC TagInfo for behavior: usually if there's no AUTHLIM it'll try factory default password and display it in dump together with the PACK, but will not attempt at all if it can't read the config page or if AUTHLIM is set.

Code-wise, I would suggest moving authentication out of the page read function for separation of concerns, and you also need to check whether authentication is actually supported before attempting to send the command.

ezhevita:
> and read AUTH0 and AUTHLIM

Sadly, pages where they are located may be protected, so we can only try reading them and hope they are accessible.
I would suggest asking for user permission if those are not available.

gornekich:
> > and read AUTH0 and AUTHLIM
> 
> Sadly, pages where they are located may be protected, so we can only try reading them and hope they are accessible. I would suggest asking for user permission if those are not available.

I will discuss with our UX team and come back with our solution later

gornekich:
Hello @ezhevita 
We decided to add "Extra Actions" in main NFC menu to give users option to read locked card with chosen password. I will finish NFC design rework #1364  and then we can merge your PR after small rework

gornekich:
Hello @ezhevita !
Let me share with you our new NFC design: https://miro.com/app/board/uXjVOlznagk=/
First iteration of new design is in dev now however It's not completely equivalent to MIRO reference yet.
I want to discuss new approach for reading locked Mf Ultralight and NTAGs. We want to read as much as possible data from locked cards and notify user that flipper couldn't read all pages.
After that we want to save all data that we managed to read. I think we should add new line in Flipper File Format for Mf Ultralight / NTAG, with how many pages were read.
Then in extra actions we will add new option for reading locked cards with generated or manually entered password. (Not in MIRO yet)

It seems like a lot of work has to be done. However we can't merge your PR and refactor code, because users can brick their cards. Please, let me know if you want to rework your PR to meet mentioned requirements. If you won't, please let me use your code to write everything myself.

ezhevita:
Hello! Sure thing, I'll try to rework this in the nearest future - I have already started moving authentication part to extra actions.

gornekich (approved):

skotopes (approved):

applications/nfc/scenes/nfc_scene_mf_ultralight_menu.c
```diff
@@ -14,7 +15,16 @@ void nfc_scene_mf_ultralight_menu_submenu_callback(void* context, uint32_t index
 void nfc_scene_mf_ultralight_menu_on_enter(void* context) {
     Nfc* nfc = context;
     Submenu* submenu = nfc->submenu;
+    MfUltralightData* data = &nfc->dev->dev_data.mf_ul_data;
 
+    if(data->data_read != data->data_size) {
```
GMMan, line 20: Would recommend checking whether authentication is supported by tag instead of whether the entire tag has been read, since some tags are not read-protected, but are write-protected and emulation may not work with original reader if PACK is incorrect.

applications/nfc/scenes/nfc_scene_mf_ultralight_unlock_menu.c
```diff
@@ -0,0 +1,70 @@
...
+        scene_manager_get_scene_state(nfc->scene_manager, NfcSceneMfUltralightUnlockMenu);
+    submenu_add_item(
+        submenu,
+        "Enter Password Manually",
+        SubmenuIndexMfUlUnlockMenuManual,
+        nfc_scene_mf_ultralight_unlock_menu_submenu_callback,
+        nfc);
+    submenu_add_item(
+        submenu,
+        "Auth As Ameebo",
```
GMMan, line 29: Spelling errors throughout

lib/nfc/protocols/mifare_ultralight.c
```diff
@@ -127,6 +170,37 @@ bool mf_ultralight_read_version(
...
+        }
+
+        // PACK
+        if(tx_rx->rx_bits < 2 * 8) {
+            FURI_LOG_D(TAG, "Authentication failed");
+            break;
+        }
+
+        if(pack != NULL) {
+            *pack = (tx_rx->rx_data[0] << 8) | tx_rx->rx_data[1];
```
GMMan, line 194: Not sure this is right, PACK is stored little endian, this would turn it big endian and subsequently reverse the bytes when sending PACK in emulated auth.

lib/nfc/protocols/mifare_ultralight.c
```diff
@@ -620,6 +697,48 @@ bool mf_ultralight_read_counters(FuriHalNfcTxRxContext* tx_rx, MfUltralightData*
     return counter_read == (is_single_counter ? 1 : 3);
 }
 
+int16_t mf_ultralight_get_authlim(
+    FuriHalNfcTxRxContext* tx_rx,
+    MfUltralightReader* reader,
+    MfUltralightData* data) {
+    mf_ultralight_read_version(tx_rx, reader, data);
```
GMMan, line 704: Separation of concerns: caller should have read version already before calling this function.

lib/nfc/protocols/mifare_ultralight.c
```diff
@@ -620,6 +697,48 @@ bool mf_ultralight_read_counters(FuriHalNfcTxRxContext* tx_rx, MfUltralightData*
...
+    } else if(
+        data->type >= MfUltralightTypeNTAGI2CPlus1K &&
+        data->type <= MfUltralightTypeNTAGI2CPlus1K) {
+        config_pages_index = 0xe3;
+    } else {
+        // No config pages
+        return -2;
+    }
+
+    if(!mf_ultralight_read_pages_direct(tx_rx, config_pages_index, data->data)) {
```
GMMan, line 722: I don't know how this function was intended to be used since nothing calls it, but reading into the beginning of the data instead of where it's supposed to be breaks all sorts of expectations. If you're not actually going to put the data where it's supposed to be, use a separate buffer, not the `MfUltralightData`.

lib/nfc/protocols/mifare_ultralight.h
```diff
@@ -50,6 +56,13 @@ typedef enum {
     MfUltralightTypeNum,
 } MfUltralightType;
 
+typedef enum {
+    MfUltralightAuthLimitUnknown,
+    MfUltralightAuthLimitNotSupported,
+    MfUltralightAuthLimitConfigured,
+    MfUltralightAuthLimitNotConfigured,
+} MfUltralightAuthLimit;
```
GMMan, line 64: Unused?

lib/nfc/protocols/mifare_ultralight.c
```diff
@@ -1,17 +1,60 @@
...
+}
+
 bool mf_ul_check_card_type(uint8_t ATQA0, uint8_t ATQA1, uint8_t SAK) {
     if((ATQA0 == 0x44) && (ATQA1 == 0x00) && (SAK == 0x00)) {
         return true;
     }
     return false;
 }
 
+void mf_ul_reset(MfUltralightData* data) {
```
GMMan, line 44: Personally I would add a comment of the fields retained and why, otherwise it's not clear why a memset is not sufficient.

GMMan (commented):
Sorry for the late review, but I found a few things that should probably be addressed.

As a side note, PWD and PACK are officially 32-bit and 16-bit little-endian integers respectively, not 4- and 2-byte arrays, but I don't really care as long as the endian conversion isn't messed up.

applications/nfc/scenes/nfc_scene_extra_actions.c
```diff
@@ -20,6 +21,12 @@ void nfc_scene_extra_actions_on_enter(void* context) {
         SubmenuIndexMfClassicKeys,
         nfc_scene_extra_actions_submenu_callback,
         nfc);
+    submenu_add_item(
+        submenu,
+        "Unlock NTAG/Ultralight",
```
GMMan, line 26: Possibly unclear wording. Unlock implies removing the protection on the tag, which this doesn't do.

applications/nfc/scenes/nfc_scene_mf_ultralight_menu.c
gornekich: Agree

applications/nfc/scenes/nfc_scene_mf_ultralight_unlock_menu.c
gornekich: That was intentionally, since amiibo is registered trademark

lib/nfc/protocols/mifare_ultralight.c
gornekich: Agree, must be fixed

lib/nfc/protocols/mifare_ultralight.c
gornekich: Actually this function is not called. Missed that after my rework

lib/nfc/protocols/mifare_ultralight.c
gornekich: Unused function should be removed by now, I agree

lib/nfc/protocols/mifare_ultralight.h
gornekich: Same here

lib/nfc/protocols/mifare_ultralight.c
gornekich: I didn't come to better solution to pass MfUltralightAuthMethod field from nfc_scene_mf_ultralight_unlock_menu. I really don't want to add extra scenes and NfcWorkerStates. It seems that we should refactor it

applications/nfc/scenes/nfc_scene_extra_actions.c
gornekich: I guess it might be more clear for users who don't know internal MF Ultralight structure. They just want "unlock" card to read it :)
Feel free to suggest your naming!

GMMan: Maybe just "Read protected NTAG/Ultralight" would work, though that might be a bit long. Can shorten "Ultralight" to "MFUL" in that case.

PR #1367, opened 2022-07-04
Auto detect raw

perspecdev:
# What's new

- Added ability to auto-detect raw signals in Sub-GHz -> Read

# Verification 

- Visit SubGHz -> Read, go to Config, and set Detect Raw to On, then back out of the config screen.  Play a signal from another device that the Flipper doesn't know the protocol for.  The raw signal should be detected and listed (e.g. "Raw 433.92") and you should be able to be save it. 

# Checklist (For Reviewer)

### DIFF
--- a/applications/subghz/scenes/subghz_scene_read_raw.c
+++ b/applications/subghz/scenes/subghz_scene_read_raw.c
@@ -99,6 +99,12 @@ void subghz_scene_read_raw_on_enter(void* context) {
         subghz->txrx->receiver, SUBGHZ_PROTOCOL_RAW_NAME);
     furi_assert(subghz->txrx->decoder_result);
 
+    // make sure we're not in auto-detect mode, which is only meant for the Read app
+    subghz_protocol_decoder_raw_set_auto_mode(
+        subghz->txrx->decoder_result,
+        false
+    );
+
     //set filter RAW feed
     subghz_receiver_set_filter(subghz->txrx->receiver, SubGhzProtocolFlag_RAW);
     view_dispatcher_switch_to_view(subghz->view_dispatcher, SubGhzViewIdReadRAW);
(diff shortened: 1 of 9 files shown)


Astrrra:
It seems to detect RAW data even when receiving regular signals. If that's the intended behavior - the RAW recordings it receives aren't complete.


https://user-images.githubusercontent.com/93453568/177229996-73ecfa7b-bb28-4546-9c27-389d7d783add.mp4

perspecdev:
> It seems to detect RAW data even when receiving regular signals. If that's the intended behavior - the RAW recordings it receives aren't complete.

That is intended behavior.  Interesting that it's not getting a complete recording.  I'm able to replay a keyfob with it, but there must still be something missing.  Thanks for the video!  Going to see if I can work out how to reproduce this without a second Flipper ;)

perspecdev:
It should be much better now.  It does have an rssi threshold (-72) so it's still possible for it to miss a signal if it falls below that threshold.

I only have 1 Flipper so I've been recreating the problem by recording and replaying a ceiling fan remote.  The video above was super helpful.

Skorpionm:
it won't work fine. in the first RAM, we do not always have it, and if you hold the remote control for 5 + seconds, it will no longer fit. write to the flash drive right away, there will not be enough time for the decoders to work. and if you try to accept Keelog, you will not have 100% natural RAW recording, since it includes many decoding algorithms

skotopes:
Hi @perspecdev

As @Skorpionm told there are reasons why we decided to keep read and raw read separately.
Mostly because of performance and memory consumption issues.

This time we are unable to accept this PR.

PS: before implementing such a big things please ping us in issues or on discord, so we'll be able to provide feedback and confirm that PR can be merged.

cloudbreakdaniel:
Thanks.  I opted to store it in memory instead of writing to disk right away because I was concerned about adding a bunch of junk files to the file system as "raw" signals are received.  My intention was for this to be a feature to capture small raw signals like a fan remote, etc.

No worries that it wasn't accepted.  I really just made it for myself.  I shared it on Discord and someone there made the suggestion to submit a PR here.

skotopes:
I'm `@skotopes#4130` on github, feel free to ping me if you have questions.

PR #1405, opened 2022-07-15
Clock

RogueMaster:
# What's new

- adds new clock app with stopwatch. Stopwatch timer has a sound alert at 1:20 that can be changed by pressing right. 

# Verification 

- open clock app, press right to change sound settings, press center to start stopwatch, press back to exit

# Checklist (For Reviewer)


Astrrra:
Please add a proper description and verification steps for your PR.

DoberBit:
looking at previous PRs, I can conclude that this character either does something that he does not understand or does not know how to use the Git system at all

Astrrra:
@DoberBit please abstain from derogatory comments that do not contribute to the discussion.

DoberBit:
> please abstain from derogatory comments that do not contribute to the discussion.

sorry, it's just hard to adequately respond to this when developers save literally every bit, but someone doesn't care at all

RogueMaster:
> Please add a proper description and verification steps for your PR.

Sorry, added the information.

GMMan:
A bit tangential to the PR at hand, but could I request there be proper commit messages in future PRs? It's a bit hard to contextualize what the changes do when they all effectively say "edited on GitHub", making things more difficult to review. Although commits are squashed upon merge, they still show up in changelogs, which are reposted publicly in various places, and these "edited on GitHub" messages would not appear professional there.

RogueMaster:
@CompaqDisc This was your clock code with some of my changes, hope you don't mind the PR

RogueMaster:
Refactored for future merge

DoberBit:
> Refactored for future merge

https://en.wikipedia.org/wiki/Code_refactoring just read this and finally understand that Ctrl+C and Ctrl+V do not fit the definition of "Refactoring"

RogueMaster:
Hi @DoberBit , the functions were changed from older ones to use the newer functions and conventions that are used globally in the firmware...

GMMan:
I think the point is the current code quality is not good and require actual refactoring.

hertzi:
I like the clock and the stopwatch is useful, but please drop the notifications and the 'songs'.
'stop' should just stop the current counter, but not delete it.
Add a 'back/reset' button from the stopwatch screen, to get to 'clock' mode again.

applications/clock_app/clock_app.c
```diff
@@ -0,0 +1,412 @@
...
+    for(bool processing = true; processing;) {
+        FuriStatus event_status = furi_message_queue_get(event_queue, &event, 100);
+        ClockState* plugin_state = (ClockState*)acquire_mutex_block(&state_mutex);
+        if(event_status == FuriStatusOk) {
+            // press events
+            if(event.type == EventTypeKey) {
+                if(event.input.type == InputTypeShort || event.input.type == InputTypeRepeat) {
+                    switch(event.input.key) {
+                    case InputKeyUp:
+                        if(timerStarted) timerSecs = timerSecs + 5;
```
hertzi: Why should you change the running timer?

applications/clock_app/clock_app.c
```diff
@@ -0,0 +1,412 @@
...
+    snprintf(strings[2], 20, "%.2d:%.2d", curMin, curSec);
+    release_mutex((ValueMutex*)ctx, state);
+    canvas_set_font(canvas, FontBigNumbers);
+    if(timerStarted) canvas_draw_str_aligned(canvas, 64, 8, AlignCenter, AlignCenter, strings[1]);
+    if(!timerStarted)
+        canvas_draw_str_aligned(canvas, 64, 26, AlignCenter, AlignCenter, strings[1]);
+    canvas_set_font(canvas, FontSecondary);
+    if(timerStarted) canvas_draw_str_aligned(canvas, 64, 20, AlignCenter, AlignTop, strings[0]);
+    if(!timerStarted) canvas_draw_str_aligned(canvas, 64, 38, AlignCenter, AlignTop, strings[0]);
+    // elements_button_left(canvas, "Alarms");
```
hertzi: please remove commented code

applications/clock_app/clock_app.c
RogueMaster: In case it was started late or early

applications/clock_app/clock_app.c
RogueMaster: will advise when all the changes are complete

applications/clock_app/clock_app.c
hertzi: To me this behavior is not intuitive and its not documented/hinted in the gui either. I would remove it.

hertzi:
> I think the point is the current code quality is not good and require actual refactoring.

Could you please either do a review or give some hints?

applications/clock_app/clock_app.c
RogueMaster: Removed

applications/clock_app/clock_app.c
RogueMaster: Removed

RogueMaster:
> please drop the notifications and the 'songs'.
> 'stop' should just stop the current counter, but not delete it.
> Add a 'back/reset' button from the stopwatch screen, to get to 'clock' mode again.

Changed for the following:
- Notifications removed
- Stop changed to not reset
- Reset changed to left button
- If timer is stopped with a value, it will show the timer value, clicking reset will zero the timer and hide it away.
- Clicking reset while timer runs will reset timer but continue its running and display.

GMMan:
> > I think the point is the current code quality is not good and require actual refactoring.
> 
> Could you please either do a review or give some hints?

Most of the things I would have mentioned have been modified while I was writing this up. Following things remain:

- Suggest calculating timer from datetime instead of a counter to minimize drift
- There is possibility of dereferencing null pointer after the call to `acquire_mutex()` in `clock_render_callback()`
- Probably use `if` instead of `switch` for key handling because you only have two cases
- Maybe comments about what you're drawing to the screen. Not currently immediately obvious

RogueMaster:
> * Probably use `if` instead of `switch` for key handling because you only have two cases

Using switch keeps the code in place for further features in the future

RogueMaster:
> * Suggest calculating timer from datetime instead of a counter to minimize drift

Please provide what should be used in its place, instead of a timerSecs it would be timerStartTime. What type would it get? 

How should I pull what it should store? If it turns up as date, how do I calculate the difference?

GMMan:
> > * Probably use `if` instead of `switch` for key handling because you only have two cases
> 
> Using switch keeps the code in place for further features in the future

I think it's dead code, and you'll notice elsewhere in the firmware `if` is typically used. Because the firmware is built with all warnings, if a new key is added in the future, it will result in this code not building due to missing handling for the new enum value. Only use `switch` if you know all cases need to be handled, and a compile error should be emitted if a new enum value is added and it has not been handled.

RogueMaster:
> * Maybe comments about what you're drawing to the screen. Not currently immediately obvious

Added

RogueMaster:
> * There is possibility of dereferencing null pointer after the call to `acquire_mutex()` in `clock_render_callback()`

I'm not sure how to fix for this

GMMan:
> > * Suggest calculating timer from datetime instead of a counter to minimize drift
> 
> Please provide what should be used in its place, instead of a timerSecs it would be timerStartTime. What type would it get?
> 
> How should I pull what it should store? If it turns up as date, how do I calculate the difference?

You can use `furi_hal_rtc_datetime_to_timestamp()` to convert to a timestamp that is based on seconds.

> > * There is possibility of dereferencing null pointer after the call to `acquire_mutex()` in `clock_render_callback()`
> 
> I'm not sure how to fix for this

Check if returned value is `NULL`, and call `view_port_update()` instead of the other logic if it is. You might have to change some stuff so you have access to the viewport. Or just don't do anything, but the screen will end up blank for a second.

RogueMaster:
> Or just don't do anything, but the screen will end up blank for a second.

I've been using current code for Clock since I started it back in April on a daily basis and never get a blank screen so I guess it's a non-issue at the moment.

GMMan:
> so I guess it's a non-issue at the moment

Sure, things work until they don't. You should be writing robust code, not just code that works now.

RogueMaster:
> You should be writing robust code, not just code that works now.

I'm satisfied with the code working as desired, it not working in some future case means it probably should be re-written by whomever needs to use it in the future. I can't account for all the changes that happen to the flipper firmware on a daily/weekly basis. It's ridiculous to think that any app written for the flipper today will work a year from now on the flipper. The flipper firmware will change as it has in the past. Find me one robust app you built months ago that will work on the flipper today. There probably are none, written by anyone. Too many refactors needed.

With that said, I'm still uncertain on how I can improve on it. I tried to see your tama app to take inspiration, could I replace mutex code with something like this to get the mutex somehow?
`    FuriMutex* mutex = context;`
`    while(furi_mutex_acquire(mutex, FuriWaitForever) != FuriStatusOk) furi_delay_tick(1);`

I also don't completely understand what mutexes are used for, so that doesn't help. But these mutex changes could be added in a future PR and doesn't really need to change this one.

GMMan:
> I'm satisfied with the code working as desired, it not working in some future case means it probably should be re-written by whomever needs to use it in the future. I can't account for all the changes that happen to the flipper firmware on a daily/weekly basis. It's ridiculous to think that any app written for the flipper today will work a year from now on the flipper. The flipper firmware will change as it has in the past. Find me one robust app you built months ago that will work on the flipper today. There probably are none, written by anyone. Too many refactors needed.

It's not about accounting for changes to the firmware in the future. It's about correctness. Not checking for `NULL` pointers when you know it could happen is incorrect. It has nothing to do with how the firmware might change in the future. If someone introduces something that causes other code to break, it's their responsibility to fix wherever it breaks.

> With that said, I'm still uncertain on how I can improve on it. I tried to see your tama app to take inspiration, could I replace mutex code with something like this to get the mutex somehow? ` FuriMutex* mutex = context;` ` while(furi_mutex_acquire(mutex, FuriWaitForever) != FuriStatusOk) furi_delay_tick(1);`
> 
> I also don't completely understand what mutexes are used for, so that doesn't help. But these mutex changes could be added in a future PR and doesn't really need to change this one.

Read about mutexes on [Wikipedia](https://en.wikipedia.org/wiki/Lock_(computer_science)). In a nutshell, it prevents multiple threads from modifying the same information, or reading the information while it's being modified. Try to understand why it might be necessary, then ask about how you could use it.

In your case both the app thread and the drawing thread is accessing the date at the same time. If the date is updated midway through the drawing thread reading it, you will get incorrect data, which would likely be pretty minor for this app, but potentially catastrophic in others. As a matter of fact, your timer variables are not guarded and should be put into `ClockState` instead.

RogueMaster:
> * Suggest calculating timer from datetime instead of a counter to minimize drift

Now this is done 😄 Thank you!

RogueMaster:
> As a matter of fact, your timer variables are not guarded and should be put into `ClockState` instead.

Agreed, thank you, updated.

GMMan:
Here's a thought: why do you need current datetime in the state at all? Just get current whenever you need it.

For `timerSecs` not sure you actually need two copies. I would just calculate diff between start time and stop time and store it into the variable directly, no need to add to it. You can probably also use the `%` operator to calculate seconds instead of doing multiplying and subtracting.

RogueMaster:
If the person stops the timer, I need to still calculate the seconds that were timed before the Stop, they can Start again without a Reset, so timerSecs is needed.

RogueMaster:
> Here's a thought: why do you need current datetime in the state at all? Just get current whenever you need it.

True, done.

Think you could help me find what could be causing a hard fault?

GMMan:
> If the person stops the timer, I need to still calculate the seconds that were timed before the Stop, they can Start again without a Reset, so timerSecs is needed.

I didn't say get rid of `timerSecs`, I said you don't need two of them (i.e. `timerTempSecs`)

RogueMaster:
> > If the person stops the timer, I need to still calculate the seconds that were timed before the Stop, they can Start again without a Reset, so timerSecs is needed.
> 
> I didn't say get rid of `timerSecs`, I said you don't need two of them (i.e. `timerTempSecs`)

I think I need both, you can PR my branch for this if you find a different solution, however, one more variable isn't killing anything.

The main issue here is that having the variables not be global seems to crash every attempt I make at this. It is completely unstable the way it is now, and 100% stable with the global vars. I understand the logic behind putting it the current way, just not sure how to fix what is broken aside from reverting to those few global variables.
`uint32_t timerStartTime;`
`uint32_t timerLastRunTime;`
`bool timerStarted;`
`int timerSecs;`
`int timerTempSecs;`

If you can gleam a solution for this, please PR.

RogueMaster:
Actually, it crashes less often with global vars, but it can still crash. I think changing it to do all the calculations, vs a dumb timerSecs+1, is making it crash now. Could multiple calls to `furi_hal_rtc_datetime_to_timestamp` and/or `furi_hal_rtc_get_datetime` cause flipper to crash? Should I try doing it through:
`            uint32_t time = LL_RTC_TIME_Get(RTC); // 0x00HHMMSS`
`            uint32_t date = LL_RTC_DATE_Get(RTC); // 0xWWDDMMYY`

RogueMaster:
> Could multiple calls to `furi_hal_rtc_datetime_to_timestamp` and/or `furi_hal_rtc_get_datetime` cause flipper to crash? Should I try doing it through:
> ` uint32_t time = LL_RTC_TIME_Get(RTC); // 0x00HHMMSS`
> ` uint32_t date = LL_RTC_DATE_Get(RTC); // 0xWWDDMMYY`

Don't think this could solve it since that's what `furi_hal_rtc_get_datetime` is doing anyways... `furi_hal_rtc_datetime_to_timestamp` doesn't seem like it could crash anything either... 

Maybe the reviewers have some insight?

RogueMaster:
I tested 1-4 scenarios below. Only 1 was stable without crashes so reverting the app to work that way. Method 1 is how I've had it working for months without crashes. Optimizations are in place to improve performance, however, the more accurate timer calculations make flipper crash and so do the non-global vars.

1) Time Calculated Based on timerSecs + 1 and Global Vars
2) Time Calculated Based on timerSecs + 1 and Non Global Vars
3) Time Calculated Based on TS and Global Vars
4) Time Calculated Based on TS and Non Global Vars

RogueMaster:
[ca59952](https://github.com/flipperdevices/flipperzero-firmware/pull/1405/commits/ca599528feb8c803e9e6c1f46a028e17a778fab9) is stable without Hard Faults like the versions before it as described above

GMMan:
Refactored: https://github.com/GMMan/flipperzero-firmware/tree/clock

RogueMaster:
Since the moderators of this community don't enforce GPL restrictions on other members and moderators changing the licensing of an open source submission, I am withdrawing my Pull Request. GPL is made the way it is to avoid these problems. The moderators of the community are actively hindering the evolution of the project by going against GPL. Don't pretend to support open source and then go against it. Buyers beware!

PR #1419, opened 2022-07-20
Enable building without Git

adisbladis:
# What's new
I want to build my Flipper firmware inside the [Nix](https://nixos.org/) build sandbox.

In this case the .git directory has been discarded by Nix for reproducibility reasons which currently causes the build to fail since
it cannot extract version information.

This can also be useful when building from source dumps such as automatically created Git archives.

The fallback when the Git repository is not available is to allow setting this information via environment variables.

# Verification 

- Verification will be contained in upcoming PR with Nix expressions.

# Checklist (For Reviewer)

### DIFF
--- a/scripts/version.py
+++ b/scripts/version.py
@@ -6,6 +6,24 @@ import subprocess
 import os
 import json
 from datetime import date
+import os.path
+
+
+_version_info_keys = [
+    "GIT_COMMIT",
+    "GIT_BRANCH",
+    "GIT_BRANCH_NUM",
+    "VERSION",
+    "BUILD_DIRTY"
+]
+
+
+class EnvVersion:
+    def get_version_info(self):
+        return {
+            k: os.environ.get(k, "" if k != "BUILD_DIRTY" else 0)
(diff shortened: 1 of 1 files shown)


hedger:
Thanks for initiating an effort for bringing reproducible builds to Flipper's firmware. That's a neat feature.

However, the changes you are proposing are not sufficient for everything to work without Git - we are also invoking it in `fbt_assets.py` for fetching current tag on protobuf submodule, which is necessary for our RPC. And there's a `get_fast_git_version_id` in `site_scons/fbt/version.py` we're using internally for build system's tasks.

We're going to add full source tree exports, containing Git versioning info within them, and fix build pipeline to fall back to these versions when Git is not available. But that would require more changes to `fbt` for it to create such exports and to be able to extract version info from them.

hedger:
If you're willing to contribute to reworking the build system in the way described, you're welcome.

skotopes:
Please talk with @hedger if you'd like to continue on nix port. He'll explain limitations on our side.

PR #1420, opened 2022-07-20
Add Nix expressions to build firmware with Nix

adisbladis:
# What's new
[Nix](https://nixos.org/) is distro & language agnostic package manager that can also be seen
as a meta build system.
This can be used to build on any Linux distro as long as the Nix
package manager is installed, and possibly also on MacOS though I
haven't tested this myself.

This is intended to be used either purely for building the firmware.
Building like this runs the build inside a sandbox.
- `$ nix-build ./.`

Or interactively using nix-shell:
- `$ nix-shell`
- `$ ./fbt firmware_flash`

# Verification 

- Run above steps.

# Checklist (For Reviewer)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,6 @@ build/
 
 # openocd output file
 openocd.log
+
+# Nix result symlinks
+result*
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,82 @@
+{ pkgs ? import <nixpkgs> { }
+, lib ? pkgs.lib
+, src ? lib.cleanSource ./.
+, version ? "dev"
+, rev ? let
+    gitRepo = "${toString src.origSrc or ""}/.git";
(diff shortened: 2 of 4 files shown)


adisbladis:
Marking as draft since it's blocked by https://github.com/flipperdevices/flipperzero-firmware/pull/1419

skotopes:
Hi, 

Thanks for contribution, but I'm not sure that we ready to include it:
- We have some specific toolchain requirements and we don't support building with 3rd party toolchains, so we are not ready to support one more build system that doesn't cover all our target platforms and requires another set of dependencies 
- As for version from environment it looks good though we don't pack source packages, that part we'll merge

cab404:
@skotopes haha nix community go brrr

chrippa:
Just want to confirm that this still works on Apple M1 Macbooks, and lets me build this firmware without the need to install anything with homebrew. Thanks @adisbladis!

PR #1421, opened 2022-07-20
Respect SOURCE_DATE_EPOCH when setting build date

adisbladis:
# What's new
When caring about reproducible builds we cannot embed dates into build
outputs, fix this by respecting the value of `SOURCE_DATE_EPOCH` (if set).

The behaviour of `SOURCE_DATE_EPOCH` is documented at https://reproducible-builds.org/docs/source-date-epoch/.

# Verification 

- Build with environment `SOURCE_DATE_EPOCH` set to some valid UNIX timestamp.

# Checklist (For Reviewer)


hedger:
Good stuff, but more changes are needed to bring fully-functional reproducible builds. See #1419.
Merging this.

hedger (approved):

PR #1452, opened 2022-07-25
Revert "Fixed building for users with space in windows username"

RogueMaster:
Reverts flipperdevices/flipperzero-firmware#1437

This push seems to break compile with assets.


RogueMaster:
I guess disregard since this is an issue in my build with #1437

PR #1471, opened 2022-07-27
nfc: Capture and display PWD_AUTH for NTAGs

GMMan:
# What's new

- Reader authentication attempts on emulated NTAGs will now display the password in a log in a similar way to UID emulation logs.

Implementation notes:
- PWD_AUTH event needs to pass additional information to event handler, and to avoid massive refactoring, I just added it as an additional field with a getter to `nfc_worker` that's only valid during the event callback. IMO it's pretty ugly, if you'd like me to refactor the callback to add an additional parameter, I can do that. Alternatively I could change the event type enum parameter to a struct pointer.
- Scene state only stores a single integer and is limiting, and I don't want to add a stuct into `nfc` just for MFUL emulate scene, so there's a bit of a crazy flags system trying to stuff multiple state variables into the int.

# Verification 

- Add manually or load any NTAG that supports password auth, emulate it, and read using NXP NFC TagInfo on Android on the full scan tab. A few seconds after reading, a log button should show up, and pressing the OK button should show a log that displays the password attempted (typically `FF FF FF FF`).

# Checklist (For Reviewer)


gornekich:
Hello @GMMan !
It wasn't clear from Discord conversation. Do you want to continue work with auto reader auth capture in this PR?
Please, let me know if you continue here. You can mark this PR as Draft and switch to Ready for review after you finish

GMMan:
I'm leaving this PR as-is, new flow stuff will be in a different PR.

gornekich:
@GMMan I am still thinking about this PR. I am working on a module which will parse tag-reader communication in different thread and NFC application can display it on screen. I think I will add this functionality differently.

PR #1484, opened 2022-07-28
nfc: collect nonces during mf classic emulation

indutny:
# What's new

- When emulating a mifare classic key it is now possible to collect and save nonces for a failed authentication and use them to recover the key through [mfkey32v2](https://github.com/equipter/mfkey32v2) tool

# Verification 

- Emulate mifare classic key and hold flipper next to a wrong reader. The "log" and "save" buttons should become visible

# Checklist (For Reviewer)

### DIFF
--- a/applications/nfc/nfc_i.h
+++ b/applications/nfc/nfc_i.h
@@ -53,6 +53,7 @@ struct Nfc {
     SceneManager* scene_manager;
     NfcDevice* dev;
     FuriHalNfcDevData dev_edit_data;
+    bool card_data_changed;
 
     char text_store[NFC_TEXT_STORE_SIZE + 1];
     string_t text_box_store;
--- a/applications/nfc/scenes/nfc_scene_mf_classic_emulate.c
+++ b/applications/nfc/scenes/nfc_scene_mf_classic_emulate.c
@@ -1,16 +1,50 @@
 #include "../nfc_i.h"
 #include <dolphin/dolphin.h>
 
-#define NFC_MF_CLASSIC_DATA_NOT_CHANGED (0UL)
-#define NFC_MF_CLASSIC_DATA_CHANGED (1UL)
+enum {
+    NfcSceneMfClassicEmulateStateWidget,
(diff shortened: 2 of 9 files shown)


gornekich:
Hi @indutny . Great work!
Mfkey32 attack is a feature that we want to implement. Since it can't bu run on Flipper, we want to offload resource demanding part to companions (mobile apps and web). It's good idea to save data necessary for mfkey32 on SD card. We need to come to the save format so that companions can easily parse data and run the attack.
Personally, I think that it's better to add this functionality to Detect Reader. We can emulate hardcoded MFC card and collect 2 pairs of nr + ar and write them to SD. Seems like partially saved data will cause many questions from users. It is not clear what you have to do with  data on the screen and raw string in log file for common user. If you collect only one MfClassicEmulatorNonce structure, you can't run mfkey32.

Also, now flipper crashes when I try to save logs.

Let me know your thoughts on data logging in Detect Reader and more clear log format.

indutny:
I agree, this belongs to other UI affordance. In my limited experiences a pair of nr + ar wasn't sufficient to deconstruct the key. In some cases it leads only to a recovered keystream, but no to the actual key. I need at least 3 pairs to get it working and I suspect others might need more than two as well.

Sorry it crashes! I'll take a look into UI placement and crashes later this week.

gornekich:
Did you use mfkey32 or mfkey32v2? The difference here is in first you use constant NT and 2 pairs of AR+NR. In mfkey32v2 you can use different NT.
In my experience your logs perfectly woks with mfkrey32v2.

gornekich:
Also, if you plan further work on this PR, could you please set it in Draft state

indutny:
@gornekich I used a v2 and it didn't really work. Converted PR to draft. Thanks!

Solilord:
Hey,
Sorry for asking: is there any Progress?
Iam waiting exactly for this 😅
I have no possibility to use WebCLI while using the reader.

gornekich:
I added nonces collection in Detect Reader here #1643

indutny:
Awesome. Thanks!

Issue #1441, opened 2022-07-24
Use custom NFC dictionaries based on their UID for known Mifare cards


der-pepe:
The nfc_worker_mifare_classic_dict_attack function takes quite some time to iterate along all the key on the mf_classic_dict.nfc file. For new cards this is just fine, since we don't actually know its keys, but for well known cards, this is quite tiresome.

A quicker method could be, before loading the generic dictionary file, to check for a custom one (maybe something in the format of {{CARD_UID}}.nfc) so it would load only the specific keys for that card.

This would help keeping the original dictionary clean, and speed up greatly the process for known cards, and in the future, even some kind of configuration mapping between multiple UIDs and a shared dictionary file could be implemented.

Astrrra:
A similar optimization is already present in the draft NFC-redesign branch, so this will be solved once that branch gets finished and merged into dev

tkerby:
So I'm trying some of the new NFC branch on the RogueMaster firmware. My comments

1 - The cache shouldn't be per card. It would be great to be able to read my own facility card for instance then instantly read others. That either means an option should be there to view the cached keys and add them to the user dictionary or the cache files should be read quickly on each card to check for matches

2 - One of the big slowdown issues in reading cards is that if I add keys to the user dictionary, then for each sector, the flipper tries through the list from top to bottom. So if I have 1000 keys in the dictionary and mine is last for sectors 1 and 5, then for sector 1 the flipper will try 1000 keys and also for sector 5 it will try 1000 keys. We should cache keys while reading a card and make them the first keys tried against any additional sector

simkard69:
This has be done in RogueMaster firmware if I'm correct 👍

tkerby:
> This has be done in RogueMaster firmware if I'm correct 👍 

Yes - at the moment it does as they've pulled the development branch of FZ which includes this

Astrrra:
Is this still relevant? Most of the improvements described above have already been implemented.

PR #1488, opened 2022-07-29
nfc: dict attack: check for key reuse

Semptum:
# What's new

- During the dict attack on MfClassic, for each new sector the device first checks keys which were already found in case of key reuse.

# Verification 

- Use the read app on a Mifare Classic badge with reused keys. The search should be faster for the sectors whose keys have already been found.

# Checklist (For Reviewer)


tkerby:
This fixes https://github.com/flipperdevices/flipperzero-firmware/issues/1441#issuecomment-1196790608 Would be great to see it included

gornekich:
Described feature was added in #1822 PR.

PR #1503, opened 2022-07-31
Add new card parsers

Astrrra:
# What's new

- Added parsers for Plantain (SPB), Two Cities (SPB+MSK), All-In-One (MSK), and added support for completely reading 4K Troika cards

# Verification 

- Read a Plantain card, verify the number and the balance
- Read a Two Cities card, verify Plantain and Troika data inside
- Read an All-In-One card, verify the number and balance
- Read a Troika 4K card, verify that it recognised it as such and has dumped the whole 4K of memory
- Read a Troika 1K card, verify that it didn't recognise it as a 4K card
- Read another Mifare Classic card, verify that it doesn't trigger any parsers
- Read another Mifare Ultralight card, also verify that it doesn't trigger any parsers

- Save the resulting dumps of all of those cards and compare them to known-good ones (for example from proxmark), verify that they match

# Checklist (For Reviewer)


Astrrra:
Currently the last digit of the plantain card is not shown as it's the checksum, and calculating it requires math with numbers that are more than 32 bits, which can't be printed properly yet. Should be fixed in zlo/printf-under-control.

gornekich (commented):
Hello @Astrrra 
Found a few problems
1) Troika card after saving parsed as Plaintain in Info scene
2) Ultralight card is not saved correctly

Astrrra:
@gornekich This should be fixed now

gornekich (dismissed):

skotopes (approved):

PR #1504, opened 2022-07-31
Add a Mifare Classic info screen to parser output

Astrrra:
# What's new

- If the Mifare Classic card has been parsed, a new "Info" option will appear in the "More" menu to allow to quickly look at the ISO data of the card (Type, size, UID, Keys found etc.)

# Verification 

- Read a Troika card. The "Info" option should be present in the "More" menu and choosing it should lead to an info screen that would be shown if the card data hasn't been parsed.
- Read a non-Troika Mifare Classic card. The "Info" option should not be present, and the ISO info should be shown on the main screen.

# Checklist (For Reviewer)


applications/nfc/scenes/nfc_scene_mf_classic_menu.c
```diff
@@ -19,6 +20,11 @@ void nfc_scene_mf_classic_menu_on_enter(void* context) {
         submenu, "Save", SubmenuIndexSave, nfc_scene_mf_classic_menu_submenu_callback, nfc);
     submenu_add_item(
         submenu, "Emulate", SubmenuIndexEmulate, nfc_scene_mf_classic_menu_submenu_callback, nfc);
+    if(string_size(nfc->dev->dev_data.parsed_data)) {
```
gornekich: You can't rely on data in this string here

gornekich (commented):
Hi @Astrrra 
What will be if you save card and leave application? Or if you read another supported card (which will be added later) and enter again? I guess there will be incorrect info
I suggest to add another method in supported cards to verify card by data, without re-read. And in Info scene call this method for all supported cards and fill parsed_data string again

gornekich (approved):

skotopes:
@Astrrra not in sync with dev

skotopes (approved):

PR #1513, opened 2022-08-01
NFC: Edit UID feature

fork-bombed:
# What's new

- Added an "Edit UID" option for saved NFC cards where only the UID can be emulated

# Verification 

- Check a saved NFC file where only the UID can be emulated, and a new option named "Edit UID" will appear
- The edit UID option will take you to the SetUid scene where you can change the hex value
- The UID will be automatically updated if you hit save, or the changes will be discared if you hit back

# Checklist (For Reviewer)


gornekich:
Hello @fork-bombed 
Why do you need to edit UID? Why can't you generate manually a new card?

fork-bombed:
> Hello @fork-bombed Why do you need to edit UID? Why can't you generate manually a new card?

Hi @gornekich. I actually came across the need for this today when testing a printer at work. The printer only checked the UID and I kept needing to create a new card manually, remember the previous bytes, and delete the old card to test different combinations. Given that only the UID can be emulated at this point, it makes sense that there should be a quick and easy way to edit the data.

applications/nfc/scenes/nfc_scene_saved_menu.c
```diff
@@ -27,6 +28,8 @@ void nfc_scene_saved_menu_on_enter(void* context) {
             SubmenuIndexEmulate,
             nfc_scene_saved_menu_submenu_callback,
             nfc);
+        submenu_add_item(
```
gornekich: Add check here
if (nfc->dev->dev_data.protocol == NfcDeviceProtocolUnknown)

gornekich (changes requested):
I see. The main concern here is that you can't change UID for Mf Classic and Ultralight/NTAG tags, since there will be inconsistent UID on anticollision level and data inside these cards. I suggest to add Edit UID option only for cards, which data we can't read, except anticollision Info

applications/nfc/scenes/nfc_scene_saved_menu.c
fork-bombed: Added.

gornekich (approved):
Great work, thanks!
Will be merged this week after 0.63 release

skotopes (approved):

PR #1545, opened 2022-08-05
Fix serial no and btn display for princeton encoding on scripts

bkerler:
# What's new

Running any subghz script with princeton encoding, the display doesn't show the right serial and btn. (Unlike when using saved recordings). Serial and btn are by default 0. This fix corrects this behaviour and displays the serial and button.

# Verification 

Create a script : 
`Filetype: Flipper SubGhz Key File
Version: 1
# 
Frequency: 433920000
Preset: FuriHalSubGhzPresetOok650Async
Protocol: Princeton
Bit: 24
Key: 00 00 00 00 00 75 ED 64
TE: 338`

If the script is being run, it should display serial as 0x75ED6 and button (btn) as 4.

# Checklist (For Reviewer)

- [X] PR has description of feature/bug or link to Confluence/Jira task
- [X] Description contains actions to verify feature/bugfix
- [X] I've built this code, uploaded it to the device and verified feature/bugfix

### DIFF
--- a/lib/subghz/protocols/princeton.c
+++ b/lib/subghz/protocols/princeton.c
@@ -158,6 +158,7 @@ bool subghz_protocol_encoder_princeton_deserialize(void* context, FlipperFormat*
             FURI_LOG_E(TAG, "Missing TE");
             break;
         }
+
         if(instance->generic.data_count_bit !=
            subghz_protocol_princeton_const.min_count_bit_for_found) {
             FURI_LOG_E(TAG, "Wrong number of bits in key");
@@ -355,6 +356,9 @@ void subghz_protocol_decoder_princeton_get_string(void* context, string_t output
 
     uint32_t code_found_reverse_lo = code_found_reverse & 0x00000000ffffffff;
 
+    int32_t serial = ((instance->generic.data >> 4) & 0xFFFFF);
+    int8_t  btn = (uint8_t)(instance->generic.data & 0x00000F);
+
     string_cat_printf(
         output,
         "%s %dbit\r\n"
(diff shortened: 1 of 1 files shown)


skotopes:
Hi. There is a fix in https://github.com/flipperdevices/flipperzero-firmware/pull/1516 please check it, meanwhile I'll close your PR

bkerler:
#1516 doesn't fix the issue actually. #1516 only seems to restructure code in regards to princeton.

skotopes:
@bkerler can you give us way to reproduce it?

bkerler:
As in the pr written. Just use my sub and load it. If it does display 000000 as serial, then it doesn't work.
If instead the serial no is displayed as 0x75ED6 and button (btn) as 4, it's all ok. The issue is primary with scripts being loaded as the serial field is never loaded from the data field.

bkerler:
In #1516 and prior, the issue is that the .serial and .btn field is never read from the generic.data field but instead points to the struct being initialized with zeros when a sub is being loaded and executed.

bkerler:
Furthermore the btn field is only of size int8(4bit) and serial of int32 (20bit), and not uint32 (which may lead to buffer overflow bugs if used in the future).

Skorpionm:
have you looked at the PR https://github.com/flipperdevices/flipperzero-firmware/pull/1516
he solves the Princeton problem. and int8 is 8 bits of data Int32 is 32 bits of data, but unsigned

bkerler:
@Skorpionm yes I tried and it doesn't work for me. Using the patches the serial and btn are still zeroed out.

bkerler:
Just tried again .. it seems to display a serial now but it's wrong. shows 0xC075ED64 instead of 0x75ED64

bkerler:
Replacing 
`instance->serial = instance->data >> 4;` with 
`instance->serial = (int32)(instance->data >> 4); `
seems to resolve the issue with #1516.

Skorpionm:
Screenshot_20220806-233631
???

Skorpionm:
if you are already completely paranoid, then you need to do instance-> serial = (instance-> data >> 4) & 0xfffff

Issue #1546, opened 2022-08-05
Lowercase keyboard displays instead of uppercase.


DrEverr:
### Describe the bug.

Lowercase keyboard displays after going to `name the file` screen, even though first letter we input is uppercase letter.

### Reproduction

(example one of many)
1. Go to Sub-GHz
2. Read RAW
3. REC (anything, doesn't matter if you have something to record, we just need to enter `Save` mode)
4. STOP
5. Go to Save
6. As you can see. Lowercase keyboard is displayed
7. Type a letter
8. It is uppercase letter

### Anything else?

Fix: keyboard should display uppercase letters on default

skotopes:
fixed in dev

PR #1548, opened 2022-08-05
Change default keyboard to Uppercase

DrEverr:
# What's new

According to Issue #1546 . Changed default keyboard to be displayed Uppercase. Because keyboard clears the screen and writes uppercase letter, even though displaying lowercase keyboard.

**[It just checks if "clear text" is set and than displays uppercase keyboard.](#)**

Before:
Screenshot-20220817-163637
Screenshot-20220817-163847

After:
Zrzut ekranu 2022-08-5 o 20 37 19
Zrzut ekranu 2022-08-17 o 16 40 03

Letter we are inputting is uppercase letter so keyboard also displays uppercase letters.

# Verification 

Even if text is present in text_box, we have displayed uppercase keyboard. Because we will input uppercase letter first.
Build project and install it on your Flipper and see if you like it. I think it's more intuitive this way. It displays what it inputs.

# Checklist (For Reviewer)


DrEverr:
Merged with newest version of code.

DrEverr:
I'm asking for feedback :D
It's not big change, but I think it upgrades comfort of using keyboard. If we gonna input first letter uppercase its good to show that on keyboard @hedger what ya think? ;D

skotopes:
Looks like there are some merge artifacts in this PR: littlefs module should not be modified.

DrEverr:
but how to fix it now 😖
my change is not that big, but I found it useful in UX meaning

skotopes (approved):

PR #1595, opened 2022-08-13
Increment U2F counter before authentication

wdoekes:
**_THIS CHANGESET BUILDS UPON #1592_**

# What's new

Start first U2F counter at 1 instead of 0. This corresponds with this flow:

image

(from https://developers.yubico.com/U2F/Protocol_details/Overview.html )

There the counter is incremented before signing/using, so the first value a relying-party will see for an authentication attempt is 1 (not 0).

This fixes so "Cloned Device" errors are not triggered on first use, as seen here: https://github.com/ossobv/django-otp-u2f/blob/33f38dbc5dfb01a4c56e82707fe7168f75f1b3fa/otp_u2f/models.py#L111-L127

# Verification 

- Delete `sdcard/u2f/cnt.u2f` and `sdcard/u2f/key.u2f` if they exist.
- Register U2F somewhere where you can see the counters.
- Log in using U2F, and confirm that the last-seen-counter is 1 instead of 0.

# Design choices

We increment twice now:
- once when also doing the bswap32/htonl
- once at the end, when storing

Simply moving the `U2F->counter++` upwards would cause the increments to increase by steps larger than 1. (Because it keeps it in memory, even when not storing the value.)

# Checklist (For Reviewer)


skotopes (approved):

skotopes:
@wdoekes thanks for PR, good job.

PR #1616, opened 2022-08-17
UPC-A barcode generator

Kavitate:
# What's new

- The readme file needs updated as the current image shows the old naming of "UPC Generator" where it should show the new "Barcode Generator" name similar to the image I have attached.

# Verification 

- Verified the name change consistent with the new RogueMaster version 0.64.5-0817-RM.

# Checklist (For Reviewer)

Menu

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,6 @@ get-pip.py
 site_scons/site_tools/crosscc.py
 site_scons/site_tools/crosscc.py
 site_scons/site_tools/crosscc.py
+.vs/flipperzero-firmware/v16/ipch/AutoPCH/a8e8884ee1b69cb1/INFRARED_SCENE.ipch
+.vs/flipperzero-firmware/v16/ipch/AutoPCH/70a3bfe70e3b6cc8/BARCODE_SCENE.ipch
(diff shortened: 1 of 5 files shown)


RogueMaster:
Did you post this here in error? Your application.fam doesn't have your desired naming according to PR notes.

RogueMaster:
Also, @McAzzaMan would need to permit for this to be pushed to main

Kavitate:
I only wanted to update the readme file to reflect the updated image. I am unsure why it has pulled everything else under it. The changes described above are the only thing I am wanting to update. The "9 commits" it has tacked in with it are irrelevant to what I am trying to accomplish. If that makes sense?

RogueMaster:
It as a PR on the main github instead of to your own still doesn't make sense.

Kavitate:
I think it defaulted to that and I must not have seen it. All I am trying to do is update the first image on the below link to the image I have uploaded above in order to make the documentation more accurate.

https://github.com/McAzzaMan/flipperzero-firmware/blob/UPC-A_Barcode_Generator/applications/barcode_generator/readme.md

Kavitate:
I believe I have found the correct way to do this so I will close this request. Thank you for your time.

PR #1629, opened 2022-08-20
OpenOCD: get logs over SWO

itizir:
# What's new

After finally getting an ST-LINK, I was a bit annoyed of not getting flipper logs back through it. Being novice with embedded devices I didn't find it straightforward to figure out how to do it, but did manage in the end.

If the SWD interface has a SWO pin, once can simply wire it to flipper GPIO pin 13 (aka USART TX / PB6), and then correctly set up openOCD, as done in this PR...

[On discord](https://discord.com/channels/740930220399525928/986635575664726026/1010121926083092480) @zhovner requested a writeup, but I figured a PR was maybe more straightforward and useful.

Checked it works on osx, and windows (to validate writing to `CON` works). The default behaviour may not be entirely portable (when `/dev/stdout` isn't a thing), but that's also why it's configurable.

Only adding this as an opt-in config option to make sure it doesn't suddenly break things for people.

# Verification 

1. Have an ST-LINK, plugged into computer
2. Connect SWD pins (VDD, CLK, GND, DIO, SWO) to flipper pins 9-13.
3. Run `./fbt OPENOCD_LOGS_SWO=True openocd`
4. Produce some logs with flipper
5. Observe they are being printed to terminal through openocd

`OPENOCD_LOGS_OUTPUT` can be set to change where logs are being sent. Can be a file, a local TCP port (preceded by `:`), or `-` to have them forwarded to TCL.

# Checklist (For Reviewer)

### DIFF
--- a/debug/swo.cfg
+++ b/debug/swo.cfg
@@ -1,0 +1,19 @@
+# script for configuring the TPIU in order to
+# pipe flipper logs through ST-LINK's SWO pin
+
+set _HCLK 64000000
+set _CONSOLE_BAUDRATE 230400
+
+set _OPENOCD_LOG_OUTPUT "/dev/stdout"
+if { $tcl_platform(platform) eq "windows" } {
+    set _OPENOCD_LOG_OUTPUT "CON"
+}
+if { [info exists ::env(OPENOCD_LOGS_OUTPUT)]
+        && $::env(OPENOCD_LOGS_OUTPUT) ne "" } {
+    set _OPENOCD_LOG_OUTPUT $::env(OPENOCD_LOGS_OUTPUT)
+}
+
+set _TPIUNAME $_TARGETNAME.tpiu
+tpiu create $_TPIUNAME -dap $_CHIPNAME.dap -ap-num 0
+$_TPIUNAME configure -protocol uart -traceclk $_HCLK -pin-freq $_CONSOLE_BAUDRATE -output $_OPENOCD_LOG_OUTPUT
+$_TPIUNAME enable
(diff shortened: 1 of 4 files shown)


Astrrra:
I think what you're looking for is just reading logs via serial?

You can do this by connecting to the ST-Link using a serial terminal like minicom or putty, for example, on mac you can use `minicom -D /dev/cu.usbmodem<number> -b 230400` and read logs that way

Or is there a reason why that doesn't work for you?

itizir:
Hey Astra! Thanks for the reply. Hmm, maybe I have an issue with drivers, or haven't configured things properly, but my ST-LINK doesn't appear as serial device in this way.

Astrrra:
@itizir what OS are you testing this on? This should work regardless of the driver...

itizir:
Right, I think I figured out what's going on: looks like only more recent V2-x and V3 have a VCOM interface ([docs](https://www.st.com/resource/en/technical_note/tn1235-overview-of-stlink-derivatives-stmicroelectronics.pdf)). I happen to only have a basic V2 ST-LINK... So no, that doesn't work for me.

skotopes:
Please talk with @hedger about how this can be integrated in compatible way and feel free to reopen this PR.

itizir:
Sure, I'm happy to do more work on it to make it fit into the codebase the way you want, if there is interest... Just let me know what's required.

I also understand it is a rather niche tooling tweak, and thankfully it wouldn't be hard for me to maintain some version of this hack locally for my own needs. Just thought it would be nice to share with the community in case other people are in the same situation.

@hedger do you want to reply here? You can also reach out on Discord if you prefer, or through whatever other means of communication.

PR #1639, opened 2022-08-23
RFID Debug: rename from `LF RFID Debug`, add LF & HF field toggles

non-bin:
# What's new

- Rename from `LF RFID Debug`
- Add options to toggle HF and LF fields

# Verification 

PXL_20220822_131740799 MP

# Checklist (For Reviewer)


Astrrra:
Isn't that partly a duplicate of an already existing feature? You can already enable the HF field in the NFC app's debug menu

skotopes:
Hi.

This PR is not going to be accepted because:
- It's prohibited to enable both fields simultaneously, it may damage radio
- We do not accept change to C++ apps, they will be dropped till v1.0

PR #1662, opened 2022-08-25
nfc: Change furi_assert to furi_crash for default switch cases

GMMan:
# What's new

- Changed select instances of `furi_assert` to `furi_crash` so code will build when `DEBUG=0`
- Any additions to the tag type enum (i.e. new type supported) **must** be handled to provide the correct values or it will crash at runtime

# Verification 

- NFC still works the same as before, no crashing

# Checklist (For Reviewer)

### DIFF
--- a/applications/nfc/helpers/nfc_generators.c
+++ b/applications/nfc/helpers/nfc_generators.c
@@ -254,7 +254,7 @@ static void
         session_register_page = 234;
         break;
     default:
-        furi_assert(false);
+        furi_crash("new MFUL type values missing\n");
         break;
     }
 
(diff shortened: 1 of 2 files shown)


applications/nfc/helpers/nfc_generators.c
```diff
@@ -254,7 +254,7 @@ static void
         session_register_page = 234;
         break;
     default:
-        furi_assert(false);
+        furi_check(false);
```
skotopes: please use furi_crash

skotopes (changes requested):

skotopes (approved):

Issue #1638, opened 2022-08-22
Option to Detect Driver/Protocol on Saved files


ppseprus:
### Description of the feature you're suggesting.

Save [this](https://github.com/flipperdevices/flipperzero-firmware/pull/1625#issuecomment-1221525241) comment yesterday, by @tkerby and I wanted to follow up on it. I did not see any Feature Request like that, so I decided to open this.

The list of known drivers and protocols change by each firmware version, and it is also subject to customisation.

An option to "Detect Driver/Protocol" for raw files would be really practical.

tkerby:
Interestingly you can process a raw file using the CLI but not the UI. Would be great to integrate post processing in the UI

ppseprus:
Wow! I did not realise that! Do you have a docs on that?

qistoph:
@ppseprus, There are brief docs in the `subghz` command on the command line.

To decode a raw recording use `subghz decode_raw <FILE>`

ppseprus:
Awesome! Thanks!

Still, I think a UI feature would be practical.

Skorpionm:
https://github.com/flipperdevices/flipperzero-firmware/pull/1667

PR #1667, opened 2022-08-26
Decode RAW recordings

qistoph:
# What's new

Flipper_SubGhz_RAW_decoder-2022-08-26_21 36 43

- Add `decode` to the More menu of saved RAW recordings
- Reuses the SubGhz Receiver View
- Tweaked this view to support:
  - Live versus File view
  - File progress (instead of Config button) in left bottom
- Reuses the SubGhz Receiver Info scene
- Add `subghz_file_encoder_worker_get_text_progress`

The icon could use a little improvement. Unfortunately I'm not a fluent designer.
assets/icons/SubGhz/Decoding_123x52.png

# Verification 

- Make a RAW recording of signals supported by available decoders
- Save the RAW recording
- From a saved RAW recording, open the `More` menu
- Select the `Decode` option
- See how Flipper decodes the recorded signals

# Checklist (For Reviewer)


skotopes:
Hi. We already have it as a cli command and intentionally decided not to make GUI version of it.

Is there any real life use cases for it?

azaitsev:
@skotopes as for me it's cool to have it in gui

qistoph:
While new protocol support is still being developed, it's a great way to test support for previously recorded codes.

I also feel, like @azaitsev, it should be part of the interface. It will also help beginner users to more easily understand and try-out the relationship between raw recordings and their decoded counter parts.

white-gecko:
> Hi. We already have it as a cli command and intentionally decided not to make GUI version of it.
> 
> Is there any real life use cases for it?
> @skotopes

Could you please point to the respective CLI?

qistoph:
> Could you please point to the respective CLI?

Checkout my comment in the related issue: https://github.com/flipperdevices/flipperzero-firmware/issues/1638#issuecomment-1223560058

RogueMaster:
I definitely like it better in the GUI

HenryDorsettC4se:
I second this. Some of us cannot use the cli often and, some of us, cannot use the phone app at all (I do not have enough money to buy a new phone and my old one is not supported :/)
The GUI/Onboard possibility to decode raw encodings is awesome imo.

jagotu:
I also very much like this feature. Unlike the CLI command, it creates an output in a format you can save. With the CLI, you'd have to take the text output and manually convert it into the .sub file format.

For me this feature is useful where you want to limit the time "in the field", for example when recording doorbells, as you can only trigger the bell twice (record raw and then try if the recording works), and then use this UI to decode the recording or even decode it later once the protocol gets added. If you were to start with non-raw recordings and the protocol was unsupported, you increase the number of recording attempts you have to make.

All-in-all for me this is a very nice additition to the FW.

skotopes:
Hi. Sorry for long delay. 

We started to split SubGhz application into specialized apps and can accept this PR as external application. Would you like to try to make this feature separate app?

skotopes (changes requested):
See comments in PR thread

skotopes:
I'll move this PR to draft, please un-draft when you complete.

CodyTolene:
Any updates on this?

skotopes:
@CodyTolene at this point the best option to implement whole thing as external application.

PR #1673, opened 2022-08-28
Lossless optimization of PNG files

Need4Video:
# What's new

- I reduced the size of PNG images (mostly by removing useless metadata)

# Verification 

- Images have to be showned exactly as before

# Checklist (For Reviewer)

- [X] PR has description of feature/bug or link to Confluence/Jira task
- [X] Description contains actions to verify feature/bugfix
- [X] I've built this code, uploaded it to the device and verified feature/bugfix


skotopes:
Not like we were having any issues with them or deploy them in raw form.
Also that's not the way how it should be done, at least it should be part of CI/CD check and be fully automated.

PR #1722, opened 2022-09-08
[FL-2674] Show error popup when NFC chip is not init/disconnected

Astrrra:
# What's new

- If rfalNfc state is not init (because the ST25R3916 chip is disconnected/broken), the NFC app will show an error popup and exit.

# Verification 

- Verify that the NFC app works as usual
- Disconnect the NFC board ribbon cable and verify that the NFC app shows an error screen

# Checklist (For Reviewer)


skotopes:
@Astrrra  we actually have dialog service to show standard errors

Astrrra:
@skotopes Right, didn't know about those, that's much better!

applications/nfc/nfc.c
```diff
@@ -224,6 +224,21 @@ void nfc_show_loading_popup(void* context, bool show) {
...
+        DialogMessage* message = dialog_message_alloc();
+        dialog_message_set_text(
+            message,
+            "Error!\nNFC chip failed to start\n\n\nSend a photo of this to:\nsupport@flipperzero.one",
+            0,
+            0,
+            AlignLeft,
+            AlignTop);
+        dialog_message_show(dialogs, message);
+        return 0;
```
gornekich: Memory leak!
After dialog_message_show() you need
1) Free message: dialog_message_free(message);
2) Close record: furi_record_close(RECORD_DIALOGS);

gornekich: Also, I would prefer to take this to separate function
static bool nfc_is_hal_ready() {...}

and in nfc_app(){
if(!nfc_is_hal_ready()) return 0

gornekich (changes requested):

applications/nfc/nfc.c
Astrrra: Fixed!

gornekich (dismissed):

skotopes (approved):

PR #1745, opened 2022-09-18
UX: Quiet subgig Read and Read RAW

Himura2la:
# What's new

- Disable flashing LED while scanning on subgig.
  - Consistent with Frequency Analyzer
  - Less suspicious when using in public places

# Checklist (For Reviewer)


skotopes:
@Himura2la san, you basically can turn off led in Settings-> Notification. There is no need to disable it for everyone.
Frequency analyzer is a little bit different thing and its notification patterns is not applicable here.

Issue #1172, opened 2022-04-27
Stopwatch and Clock


RogueMaster:
Would be nice to have a stopwatch and a clock. It could even be an accessible feature in the unimplemented dumb mode.

skotopes:
Possible, low priority. Anyone who will want to implement are welcomed to do that.

zhovner:
There is no guarantees that Flipper's internal clock is synced and  accurate. Thats why we not implement clock in it. 

About stopwatch -- doesn't your phone can do this better?

Suggested to close.

RogueMaster:
I have a clock and stop watch application now. Seems to work well. Not sure why you would suggest to close such a valuable feature that is so easy to add. Clock is now my main favorite/first app on my flipper and my most used.

RogueMaster:
I will confirm with CompaqDisc on his plans to submit his clock app. I modified his clock app to have the stopwatch.

RogueMaster:
Here is the code for clock and stopwatch. https://gist.github.com/RogueMaster/ed2e16647a0ade3372dcbeb078507517 
I can remove sound alerts if it is preferred

skotopes:
furi_hal_speaker is meant to be used throw notification service.

RogueMaster:
How do you recommend it be changed? If that is what you are suggesting.

I am using it to throw a notification yes. Service was not needed as it is operating as desired without bug/compile errors.

skotopes:
Something like: open notification api record, send notification sequence

RogueMaster:
Updated gist to remove the sound notifications... https://gist.github.com/RogueMaster/ed2e16647a0ade3372dcbeb078507517

RogueMaster:
I have updated the sound notifications so it uses the the proper process. Please see code here: https://github.com/RogueMaster/flipperzero-firmware-wPlugins/blob/dev/applications/clock_app/clock_app.c

jimilinuxguy:
@RogueMaster, what value does a stop watch bring to Flipper?

Not trying to put down @RogueMaster’s work, as I think you are attempting to make a contribution to the community and a product you support and that’s awesome!

Mirroring @zhovner’s sentiment, I already have a more accurate stop-watch on my wrist and I’m my pocket on my phone.

The flipper is supposed to be a toy/multi-tool for signal investigation and USB hid emulation.

[...]

PR #1750, opened 2022-09-19
Clock v1

kowalski7cc:
# What's new

- See #1172
- Built-in clock and settings
- Single view, only clock mode
- Userful for dummy mode?

# Verification 

- Menu > Clock
- Menu > Settings > Clock

# Checklist (For Reviewer)


kowalski7cc:
Rebased on dev

digitalcircuit:
> * Userful for dummy mode?

Beyond adding to dummy mode, this would be useful for utilizing the Flipper as an oversized smart watch by adding a wrist-strap.  (Yes, I am seriously considering this.)

A future enhancement after this is merged might be integrating the hours/minutes in the status bar of the desktop view itself.

applications/main/clock/application.fam
```diff
@@ -0,0 +1,25 @@
...
+    appid="clock",
+    name="Clock",
+    apptype=FlipperAppType.APP,
+    entry_point="clock_app",
+    cdefines=["APP_CLOCK"],
+    requires=["gui"],
+    provides=["clock_settings"],
+    icon="A_Clock_14",
+    stack_size=2 * 1024,
+    order=9,
```
skotopes: Please place application between `U2F` and `Applications`

applications/main/clock/clock_app.c
```diff
@@ -0,0 +1,197 @@
...
+    FURI_LOG_T(TAG, "clock_render_callback - start");
+
+    canvas_clear(canvas);
+    canvas_set_color(canvas, ColorBlack);
+
+    ClockState* state = ctx;
+    if(furi_mutex_acquire(state->mutex, 200) != FuriStatusOk) {
+        FURI_LOG_D(TAG, "Can't obtain mutex, requeue render");
+        PluginEvent event = {.type = EventTypeTick};
+        furi_message_queue_put(state->event_queue, &event, 0);
```
skotopes: Nope. Bad pattern, Draw call is only for drawing.
The only thing you are allowed to do is to take model mutex, draw and then release mutex.
Active logic must be handled in application main loop.

applications/main/clock/clock_app.c
```diff
@@ -0,0 +1,197 @@
...
+        snprintf(
+            date_string,
+            DATE_LEN,
+            CLOCK_RFC_DATE_FORMAT,
+            state->datetime.day,
+            state->datetime.month,
+            state->datetime.year);
+    }
+
+    furi_mutex_release(state->mutex);
```
skotopes: You are releasing mutex too early. Release it only after last model dereference.

applications/main/clock/clock_app.c
```diff
@@ -0,0 +1,197 @@
...
+
+    canvas_set_font(canvas, FontBigNumbers);
+    canvas_draw_str_aligned(canvas, 64, 28, AlignCenter, AlignCenter, time_string);
+    canvas_set_font(canvas, FontSecondary);
+    canvas_draw_str_aligned(canvas, 64, 42, AlignCenter, AlignTop, date_string);
+
+    if(state->settings.time_format == H12)
```
skotopes: Here last use.

applications/main/clock/clock_app.c
```diff
@@ -0,0 +1,197 @@
...
+    if(!load_success) {
+        state->settings = DEFAULT_SETTINGS;
+    }
+    FURI_LOG_I(TAG, "Time format: %s", state->settings.time_format == H12 ? "12h" : "24h");
+    FURI_LOG_I(
+        TAG, "Date format: %s", state->settings.date_format == Iso ? "ISO 8601" : "RFC 5322");
+    furi_hal_rtc_get_datetime(&state->datetime);
+}
+
+// Runs every 1000ms by default
```
skotopes: Systick 1000ms are not perfectly aligned with 1000ms RTC.
Also systick may be delayed in some cases.
So seconds counter will be slightly "shaky".
To make it "smoother" other time source or oversampling should to be used.

Overall it's not an issue, we'll accept PR with current implementation.

applications/main/clock/clock_app.c
```diff
@@ -0,0 +1,197 @@
...
+    plugin_state->event_queue = furi_message_queue_alloc(8, sizeof(PluginEvent));
+    if(plugin_state->event_queue == NULL) {
+        FURI_LOG_E(TAG, "Cannot create event queue");
+        free(plugin_state);
+        return 255;
+    }
+    FURI_LOG_D(TAG, "Event queue created");
+
+    plugin_state->mutex = furi_mutex_alloc(FuriMutexTypeNormal);
+    if(plugin_state->mutex == NULL) {
```
skotopes: Never going to happen. 
All alloc functions are always return valid instance or crash system.

applications/main/clock/clock_app.c
```diff
@@ -0,0 +1,197 @@
...
+
+    // Set system callbacks
+    ViewPort* view_port = view_port_alloc();
+    view_port_draw_callback_set(view_port, clock_render_callback, plugin_state);
+    view_port_input_callback_set(view_port, clock_input_callback, plugin_state->event_queue);
+
+    FuriTimer* timer =
+        furi_timer_alloc(clock_tick, FuriTimerTypePeriodic, plugin_state->event_queue);
+
+    if(timer == NULL) {
```
skotopes: Same.

applications/main/clock/clock_app.c
```diff
@@ -0,0 +1,197 @@
...
+    Gui* gui = furi_record_open(RECORD_GUI);
+    gui_add_view_port(gui, view_port, GuiLayerFullscreen);
+
+    furi_timer_start(timer, furi_kernel_get_tick_frequency());
+    FURI_LOG_D(TAG, "Timer started");
+
+    // Main loop
+    PluginEvent event;
+    for(bool processing = true; processing;) {
+        FuriStatus event_status = furi_message_queue_get(plugin_state->event_queue, &event, 100);
```
skotopes: Using 100ms timeout make sense only if you do some work on expire, otherwise just put forever.

applications/main/clock/clock_app.c
```diff
@@ -0,0 +1,197 @@
...
+
+        if(event_status != FuriStatusOk) continue;
```
skotopes: Set wait forever and the remove this line

applications/main/clock/clock_app.c
```diff
@@ -0,0 +1,197 @@
...
+
+        if(furi_mutex_acquire(plugin_state->mutex, FuriWaitForever) != FuriStatusOk) continue;
```
skotopes: FuriWaitForever will always return Ok. You probably want to add furi_check here. `if` is not needed

applications/main/clock/clock_app.c
```diff
@@ -0,0 +1,197 @@
...
+                    // Exit the plugin
+                    processing = false;
+                    break;
+                }
+            }
+        } else if(event.type == EventTypeTick) {
+            furi_hal_rtc_get_datetime(&plugin_state->datetime);
+        }
+
+        view_port_update(view_port);
```
skotopes: update after mutex release is better option

applications/main/clock/clock_app.h
```diff
@@ -0,0 +1,47 @@
...
+// typedef enum {
+//     Clock,
+//     Timer,
+//     Counter,
+// } Mode;
+
+typedef struct {
+    ClockSettings settings;
+    FuriHalRtcDateTime datetime;
+    FuriMutex* mutex;
```
skotopes: Better to split model data into separate entity.
So it will be easier to understand what exactly you are guarding with mutex.

applications/main/clock/clock_app.c
```diff
@@ -0,0 +1,197 @@
...
+    if(state->settings.time_format == H12)
+        canvas_draw_str_aligned(canvas, 65, 12, AlignCenter, AlignCenter, meridian_string);
+
+    FURI_LOG_T(TAG, "clock_render_callback - end");
+}
+
+static void clock_state_init(ClockState* const state) {
+    bool load_success = LOAD_CLOCK_SETTINGS(&state->settings);
+    if(!load_success) {
+        state->settings = DEFAULT_SETTINGS;
```
skotopes: See comment on settings structure.

applications/main/clock/clock_settings.h
```diff
@@ -0,0 +1,53 @@
...
+#define LOAD_CLOCK_SETTINGS(x) \
+    saved_struct_load(           \
+        CLOCK_SETTINGS_PATH,   \
+        (x),                     \
+        sizeof(ClockSettings), \
+        CLOCK_SETTINGS_MAGIC,  \
+        CLOCK_SETTINGS_VER)
+
+typedef enum {
+    Sound = 1,
```
skotopes: We usually use 0 as default. In that case you don't have to fill default values when setting not loaded.

applications/main/clock/clock_settings.h
```diff
@@ -0,0 +1,53 @@
...
+    Vibro = 2,
+    Silent = 3,
+} AlarmMode;
```
skotopes: Also please don't add things if they are not implemented.

applications/main/clock/clock_settings_app.c
```diff
@@ -0,0 +1,102 @@
...
+}
+
+static void date_format_changed(VariableItem* item) {
+    ClockAppSettings* app = variable_item_get_context(item);
+    uint8_t index = variable_item_get_current_value_index(item);
+    variable_item_set_current_value_text(item, date_format_text[index]);
+    app->clock_settings.date_format = date_format_value[index];
+}
+
+static ClockAppSettings* alloc_settings() {
```
skotopes: Always prefix methods with application name in applications compiled into firmware.

skotopes: Yes, static keyword makes it local, but debugging application still becomes hell

applications/main/clock/clock_settings_app.c
```diff
@@ -0,0 +1,102 @@
...
+    ClockAppSettings* app = malloc(sizeof(ClockAppSettings));
+    LOAD_CLOCK_SETTINGS(&app->clock_settings);
```
skotopes: If you open settings first you'll get invalid setting file. See my previous comments on how to implement it safe.

skotopes (changes requested):

applications/main/clock/clock_app.c
GMMan: IIRC I wrote that last time this went around. Reason this was here was because drawing only occurs every second, so if mutex was locked by update logic, you'd miss drawing and the display would end up blank for a whole second. Maybe update loop could tick more frequently and use a flag for whether a redraw is required.

skotopes: There are 2 ways how you can solve it:

- Lock-free data model: when all fields are atomic then you don't have to lock it. For example if used types not exceed machine word.
- Representation Model filled in one go and mutex is only held for a short time. (View is doing it).

Both can be applied here.

applications/main/clock/clock_app.h
kowalski7cc: The issue is that I'm using settings data to choose if draw 12/24h and RFC or ISO date

kowalski7cc:
@Eng1n33r @GMMan Would you be ok if this gets upstreamed? I don't know the original author of the clock icon.

GMMan:
> @Eng1n33r @GMMan Would you be ok if this gets upstreamed? I don't know the original author of the clock icon.

I'm fine with it

skotopes:
We need to move this app to SD ;-(

skotopes:
Sorry for delay. We are currently working on locale settings that will cover time format and then I'll merge this one.

skotopes (dismissed):

skotopes (approved):

PR #1769, opened 2022-09-22
NFC Fix Mifare Classic

alsacchi:
Wrong cast lead to unexpected behavior converting key from str to int.

# Verification 

- Compile firmware before this commit and the mifare classing reading shouldn' t work because the key conversion lead to unexpected key.
- Compile firmware with this patch and the mifare classic reading should work again

# Checklist (For Reviewer)


skotopes:
@alsacchi can you also drop some unit tests?

Astrrra (dismissed):

skotopes (approved):

skotopes:
@alsacchi thanks for nice catch.

PR #1802, opened 2022-09-29
Add support for building/flashing on NixOS

twitchyliquid64:
NixOS users can run `nix-shell` to land in a development environment where ./fbt commands will work.

# What's new

- Compiling & flashing now works on NixOS.

# Verification 

- Tested with ./fbt, ./fbt flash_usb, and ./fbt plugin_dist

# Checklist (For Reviewer)

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -78,7 +78,7 @@ brew bundle --verbose
 
 ## Linux Prerequisites
 
-The FBT tool handles everything, only `git` is required.
+The FBT tool handles everything, only `git` is required. NixOS users can run `nix-shell` to drop into a development shell.
 
 ### Optional dependencies
 
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,20 @@
+{ pkgs ? import <nixpkgs> {} }:
+(pkgs.buildFHSUserEnv {
+  name = "flipperzero";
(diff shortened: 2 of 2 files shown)


hedger:
Hello!
fbt comes with scrips that download & use a specifically crafted & tested toolchain with explicit library dependecies and everything. That toolchain is only used by fbt internally and does not contaminate your system's PATH.

With your approach, we lose control over toolchain part, and implementing changes in the build system would also require verifying & maintaining  nix-shell configuration. We cannot afford that at the moment.
cc: @skotopes @drunkbatya

twitchyliquid64:
I wasn't able to get the toolchain that was shipped working unfortunately. My best guess is that it assumes a FHS system which won't work with a paradigm like NixOS.

What are the customizations youve made to the toolchain? The nix-shell in this PR just creates a FHS environment with the python stuff ready to go, and scons + zlib + gcc10 installed (leaving the rest to fbt).

I'm happy to maintain this and/or add a warning that this is not officially supported. If none of this is palatable though and I best close the PR I understand.

skotopes:
@twitchyliquid64 as @hedger said: toolchain that we ship contains carefully selected dependencies that is very difficult to recreate if os package manager is used. NixOS is not our main target for development and we are not ready to support it at this point.

PR #1822, opened 2022-10-03
[FL-2803] Mifare Classic read improvements Part 1

Astrrra:
# What's new

- The last line of Mifare Classic dictionaries is not skipped now
- Check for duplicate keys on the card for each found key

# Verification 

## Skipping of the last line
- Modify the dictionary to contain only 1 key key with no newline at the end, create a card with all sectors protected by that key
- Check that the dictionary works as expected, the card gets fully read
- Add a newline to the end of the dict, check that it still works
- Create a card with 5 different keys, add them to the dict and re-run steps 2 and 3

## Re-used keys
- Create a card that has the same key for all sectors
- Edit the dictionary to contain > 10 random keys and the correct one at the end
- Try reading the card, the sectors after the first one should get checked much quicker than others (but not yet read)
- Verify that the dump is correct and no data is lost

# Checklist (For Reviewer)


lib/nfc/helpers/mf_classic_dict.c
```diff
@@ -69,9 +69,15 @@ MfClassicDict* mf_classic_dict_alloc(MfClassicDictType dict_type) {
...
+            if(!stream_read_line(dict->stream, next_line)) {
+                FURI_LOG_T(TAG, "No keys left in dict");
+                break;
+            }
+            string_strim(next_line);
+            FURI_LOG_T(
+                TAG, "Read line: %s, len: %d", string_get_cstr(next_line), string_size(next_line));
             if(string_get_char(next_line, 0) == '#') continue;
-            if(string_size(next_line) != NFC_MF_CLASSIC_KEY_LEN) continue;
+            if(string_size(next_line) != NFC_MF_CLASSIC_KEY_LEN - 1) continue;
```
gornekich: I suggest changing NFC_MF_CLASSIC_KEY_LEN value since you add string_strim() call

lib/nfc/nfc_worker.c
```diff
@@ -453,6 +453,36 @@ void nfc_worker_emulate_mf_ultralight(NfcWorker* nfc_worker) {
...
 }
 
+void nfc_worker_mf_classic_key_attack(NfcWorker* nfc_worker, uint64_t key, FuriHalNfcTxRxContext* tx_rx) {
+    furi_assert(nfc_worker);
+
+    MfClassicData* data = &nfc_worker->dev_data->mf_classic_data;
+    uint32_t total_sectors = mf_classic_get_total_sectors_num(data->type);
+
+    // Check every sector's A and B keys with the given key
+    for(size_t i = 0; i < total_sectors; i++) {
```
gornekich: Let's pass start sector to nfc_worker_mf_classic_key_attack() arguments. We don't need to try found key starting from 0 sector each time

lib/nfc/nfc_worker.c
```diff
@@ -453,6 +453,36 @@ void nfc_worker_emulate_mf_ultralight(NfcWorker* nfc_worker) {
     }
 }
 
+void nfc_worker_mf_classic_key_attack(NfcWorker* nfc_worker, uint64_t key, FuriHalNfcTxRxContext* tx_rx) {
```
gornekich: Please make this function static so that we understand that this is local function

lib/nfc/helpers/mf_classic_dict.c
Astrrra: I tried doing that at first, but it breaks everything else as most functions expect this to be 13

lib/nfc/nfc_worker.c
Astrrra: Yep, sounds good!

gornekich (approved):

skotopes (approved):

Issue #1842, opened 2022-10-07
nfc: Using "Unlock With Password" option for MFUL and cancelling results in blank data


GMMan:
### Describe the bug.

When you get to the read auth step in the "Unlock With Password" flow for NTAG and go back out to the read success screen, you get blank UID and pages read 0/0.

### Reproduction

1. Go to NFC app and select "Read"
2. Scan a password-protected NTAG
3. Select "More"
4. Select "Unlock With Password"
5. Use any of the unlock options
6. Accept the warning
7. Go back all the way to the read success screen when Flipper is tries to scan
8. Observe invalid data on read success screen

### Target

f7

### Anything else?

Note that the unlock flow also doesn't check whether you're running it on the original tag that was read. Of course, it can't check because it cleared all the data before starting read. Perhaps just not clearing the data is OK?

PR #1843, opened 2022-10-07
nfc: NTAG password auto capture (and other password-related changes)

GMMan:
# What's new

- Add flow to automatically capture NTAG PWD from reader
- Miscellaneous MFUL bug fixes
- Default MFUL manual password entry value is 0xFFFFFFFF (as shipped on tags)
  - Note: a simple read of the tag never captures the PWD and PACK because they are blanked out by the tag during read, so the unlock options will always be available for a newly read tag.
- Show PWD and PACK on tag info page if available
  - Useful if you need to write to an emulated tag with something and need the password to unlock it before writing, since password authentication is faithfully emulated
- Add options to unlock MFUL from loaded file
  - You could certainly just read the card from scratch again, and currently it still asks you to save to a different file, so some discussion might be needed whether this is useful or whether it should auto save back to the original file after unlock

Supercedes #1471

The bug in #1842 is present, this PR **does not** attempt to address the issue

# Verification 

- Automatic NTAG PWD capture flow
  1. Read password-protected NTAG as usual
  2. Select "Unlock", then "Unlock With Reader"
  3. Flipper emulates read card so far; tap Flipper to original reader the card was supposed to be used with
  4. If the reader sent a PWD_AUTH command, Flipper will beep and continue after it has been removed from reader
  5. Flow resumes to "Unlock With Password" flow after entering password; following through should result in the matching card being completely read
 - Default MFUL manual password entry value is 0xFFFFFFFF
   1. Read password-protected NTAG as usual
   2. Select "Unlock With Password"
   3. Select "Enter Password Manually"
   4. The password should show `FF FF FF FF` by default
- Show PWD and PACK on tag info page if available
  1. Load an NTAG NFC file that has been unlocked with password
  2. Select "Info"
  3. Scroll down; Password and PACK should be displayed
- Add options to unlock MFUL from loaded file
  1. Load an NTAG NFC file that is password protected but has not been unlocked
  2. "Unlock" option should appear in the menu and function as if the card was newly read

# Checklist (For Reviewer)


GMMan:
I just remembered the extra actions menu also lets you access unlock options, and auto unlock is not valid without a partial read since the data needs to be emulated. If we're going to merge the auto unlock into the unlock menu, it'll need a check so it won't show up if you get to the unlock menu from extra actions.

lib/nfc/nfc_worker.c
```diff
@@ -49,6 +50,10 @@ NfcWorkerState nfc_worker_get_state(NfcWorker* nfc_worker) {
     return nfc_worker->state;
 }
 
+void* nfc_worker_get_event_data(NfcWorker* nfc_worker) {
```
gornekich: I see you faced the problem passing data from nfc worker to nfc scene. Adding event_data in nfc_worker doesn't look like the best solution. 

I see 2 different approaches here. 
1) To change NfcWorkerCallback signature. We can extend NfcWorkerEvent type, so that it will have NfcWorkerEventType and NfcWorkerEventData. This approach is used frequently used in flipper firmware. 

2) To add MfUltralightAuth structure to NfcDeviceData. The NfcDeviceData structure can be used to share data between worker and application. We have already a union with reader_data and mf_classic_dict_attack_data, which were created for similar reasons. 

I would prefer 2nd solution for now. It takes less refactoring and data filled in NfcDeviceData has potentially longer lifetime and could be used in different scenes

gornekich (changes requested):
Hello @GMMan !
Thanks for the PR, great work! I tested it and it works good.
I am waiting for the feedback from our UX/UI team to discuss some naming changes with you.
From technical point of view everything is almost done. I left a comment with something I would want to be changed. Please tell me if it is not clear

gornekich:
I notice some problems while testing:
1) When Unlock with reader is running, after Password was captured, the view is frozen for some time. That's what you already mentioned and I want to debug it.
2) When I go through all process and save fully unlocked card, it doesn't seem that PASSWORD and PACK are saved. When I load saved file, lasy pages are still zero and flipper suggest to unlock card again.
3) If I want to stop unlocking at some point and press back, the first read screen looks broken

Also the unlocking algorithm is quite complex and may become confusing at some point. I am waiting for the suggestions from UX/UI team to make GUI more clear and interactive.

Btw, how do you test this feature?

GMMan:
> 2\. When I go through all process and save fully unlocked card, it doesn't seem that PASSWORD and PACK are saved. When I load saved file, lasy pages are still zero and flipper suggest to unlock card again.

This one I'm not sure about, but since the flow is currently passing off to manual entry unlock after capture, it could possibly be preexisting. Which particular tag type?

> 3\. If I want to stop unlocking at some point and press back, the first read screen looks broken

This one is #1842 as mentioned and preexisting, happens if you use any other unlock as well.

> Btw, how do you test this feature?

I demo it [here](https://twitter.com/GMMan_BZFlag/status/1579209759399776256) using a Vital Bracelet toy and its accompanying app.

gornekich:
> This one I'm not sure about, but since the flow is currently passing off to manual entry unlock after capture, it could possibly be preexisting. Which particular tag type?

I am testing with NTAG215.

I am still waiting for the design for this feature.

There is another thing I want to discuss. Since most NTAGs has default AUTH0 = FF, PROT = 0, AUTHLIM = 0, PWD = FFFFFFFF and PACK = 0000, I suggest to add this logic.
1) Read all pages, including configuration pages. (already implemented)
2) If configuration pages have default values (no protection, no auth attempl limits) try default PWD.
3) If AUTH success - read PACK and now we have fully read tag.

The motivation behind this is that in my experience most NTAGs has default configuration pages. It's a little bit confusing for users to see unlock suggestions, even if they already read all necessary data.

Also I am still working on reducing delay after PWD was captured from reader.

GMMan:
> There is another thing I want to discuss. Since most NTAGs has default AUTH0 = FF, PROT = 0, AUTHLIM = 0, PWD = FFFFFFFF and PACK = 0000, I suggest to add this logic.
> 
> 1. Read all pages, including configuration pages. (already implemented)
> 2. If configuration pages have default values (no protection, no auth attempl limits) try default PWD.
> 3. If AUTH success - read PACK and now we have fully read tag.
> 
> The motivation behind this is that in my experience most NTAGs has default configuration pages. It's a little bit confusing for users to see unlock suggestions, even if they already read all necessary data.

Sounds reasonable. I've pushed up the change, although it only checks whether the whole tag has been read and `AUTHLIM == 0`. I believe this aligns with the behavior seen on NXP NFC TagInfo. This does add a tiny bit of delay from the tag's anti-bruteforcing if the PWD happens to be incorrect, so if we want to try avoiding that I can add in the other checks.

gornekich:
So I figured out the problem with a delay between password is captured and "Risky function" scene appears. Currently if AUTH command is sent by reader and flipper doesn't store password, the Amiibo PACK is sent. After that in furi_hal_nfc_emulate_nfca() we wait for 5 seconds to receive next command and only after that we check if emulator.auth_attempt in nfc_worker_emulate_mf_ultralight().

Currently I am working on nfc "low level" refactoring. I want to significantly change nfc API in furi_hal_nfc layer and after that refactor all upper layers. This will take some time, but after it's done It will be much easier to write code to add new features and protocols.

I would want to see "password capture" feature the way it's done in Detect Reader function for capturing nonces for mfkey32 attack. The reader analyzer can process communication between reader and tag in a different thread asynchronously and call callback on different event.
For now I think I will add Reader Analyzer function to mifare ultralight emulation and push it to your branch or send you patch

GMMan:
> > This one I'm not sure about, but since the flow is currently passing off to manual entry unlock after capture, it could possibly be preexisting. Which particular tag type?
> 
> I am testing with NTAG215.

I am able to reproduce this, will look into what's happening.

GMMan:
> I am able to reproduce this, will look into what's happening.

This was actually caused by something from my latest commit, so not related to original issue. Other than that my sample NTAG215 seems to be working fine.

I suggest turning on debug logging and check what commands are being sent to the tag while doing the authenticated read.

applications/main/nfc/scenes/nfc_scene_mf_ultralight_read_auth_result.c
```diff
@@ -20,16 +20,19 @@ void nfc_scene_mf_ultralight_read_auth_result_on_enter(void* context) {
...
     FuriString* temp_str;
     temp_str = furi_string_alloc();
 
     if((mf_ul_data->data_read == mf_ul_data->data_size) && (mf_ul_data->data_read > 0)) {
-        widget_add_string_element(
-            widget, 64, 0, AlignCenter, AlignTop, FontPrimary, "All pages are unlocked!");
+        if(mf_ul_data->auth_success)
+            title = "All pages are unlocked!";
+        else
+            title = "All unlocked but failed auth!";
```
GMMan, line 30: Will need to make this shorter somehow because the unlocked message is already barely fitting on the screen.

gornekich: Don't worry about design, we will send you suggestions soon

gornekich:
Hi @GMMan !
There is no good way to capture password with reader analyzer. I will rework it myself after nfc low level API is done and MFU emulation is reworked. I think I will finish it next week.
I attach a patch that can be used as a temporary solution to decrease time between password capturing a
nd switching to next scene. You can run `git apply faster_password_capture` .
[faster_password_capture.txt](https://github.com/flipperdevices/flipperzero-firmware/files/9794905/faster_password_capture.txt)

Also the suggested "auto unlock" feature design is now in Miro. https://miro.com/app/board/uXjVOlznagk=/
Let's discuss it here. And let me know if you want to implement it yourself or you need some help

gornekich:
Hello @GMMan !
What do you think about suggested design? Do you want to implement it yourself? Or can we continue work based on your PR?

GMMan:
I had a look through the diagram and I think I can implement. Will be doing that this weekend.

skotopes:
Any progress?

GMMan:
Currently working on it

GMMan:
Got busy again, I'll resubmit this after I get the changes in.

applications/main/nfc/scenes/nfc_scene_mf_ultralight_read_auth_result.c
GMMan: New designs doesn't seem to cover this case, so still need wording that will fit.

gornekich: It's totally OK. Could you please add braces {} in if else statements

applications/main/nfc/scenes/nfc_scene_mf_ultralight_read_auth_result.c
```diff
@@ -81,8 +85,20 @@ bool nfc_scene_mf_ultralight_read_auth_result_on_event(void* context, SceneManag
...
-            nfc->scene_manager, NfcSceneMfUltralightUnlockMenu);
+        MfUltralightData* mf_ul_data = &nfc->dev->dev_data.mf_ul_data;
+        if(mf_ul_data->auth_method == MfUltralightAuthMethodManual ||
+           mf_ul_data->auth_method == MfUltralightAuthMethodAuto) {
+            consumed = scene_manager_previous_scene(nfc->scene_manager);
+        } else {
+            NfcScene next_scene;
+            if((mf_ul_data->data_read == mf_ul_data->data_size) && (mf_ul_data->data_read > 0))
+                next_scene = NfcSceneMfUltralightMenu;
+            else
```
gornekich: Also add {} here too please

gornekich (dismissed):
Thanks for your changes! Tested it and it works great!

gornekich (approved):

skotopes (approved):

skotopes:
@GMMan That is some serious work that you've done. お疲れ様でした。

PR #1864, opened 2022-10-11
Allow pins 0 and 1 as RTS/DTR for USB UART Bridge

antimatter15:
# What's new

I've been working on a mini ESP32 wifi module (https://forum.flipperzero.one/c/3rd-party-modules/15). It's a "half-size module" which only uses the right half of the output ports. I'd like to be able to flash the ESP32 using only the Flipper Zero as a UART bridge and the RTS/DTR pins. However, the only current options for RTS pins are 2,3 and 6,7 which are on the left side of the board. If we could use the PC0 and PC1 GPIO ports as RTS/DTR then it could be much easier to flash an ESP32 with the mini wifi module.

# Verification 

- Compile and upload
- Use USART bridge from GPIO application, ensure all variants behave correctly

# Checklist (For Reviewer)


applications/main/gpio/usb_uart_bridge.c
```diff
@@ -11,9 +11,10 @@
 #define USB_CDC_BIT_DTR (1 << 0)
 #define USB_CDC_BIT_RTS (1 << 1)
 
-static const GpioPin* flow_pins[][2] = {
+static const GpioPin* flow_pins[][3] = {
     {&gpio_ext_pa7, &gpio_ext_pa6}, // 2, 3
     {&gpio_ext_pb2, &gpio_ext_pc3}, // 6, 7
+    {&gpio_ext_pc0, &gpio_ext_pc1}, // 0, 1
```
skotopes: Incorrect pin numbers, also not going to work because of missing logic in gpio_scene_usb_uart_cfg

skotopes (changes requested):

applications/main/gpio/usb_uart_bridge.c
antimatter15: does this fix it?

skotopes: Nope: 16,15.

And one more blocker: pins 15,16 are also responsible for LPUART, so you can not use them if LPUART selected.

skotopes (changes requested):

skotopes:
Also please do `./fbt format` before submitting PR.

applications/main/gpio/scenes/gpio_scene_usb_uart_config.c
```diff
@@ -13,7 +13,7 @@ static UsbUartConfig* cfg_set;
 
 static const char* vcp_ch[] = {"0 (CLI)", "1"};
 static const char* uart_ch[] = {"13,14", "15,16"};
-static const char* flow_pins[] = {"None", "2,3", "6,7"};
+static const char* flow_pins[] = {"None", "2,3", "6,7", "0,1"};
```
skotopes: 16,15

applications/main/gpio/scenes/gpio_scene_usb_uart_config.c
```diff
@@ -116,7 +116,12 @@ void gpio_scene_usb_uart_cfg_on_enter(void* context) {
     variable_item_set_current_value_index(item, cfg_set->uart_ch);
     variable_item_set_current_value_text(item, uart_ch[cfg_set->uart_ch]);
 
-    item = variable_item_list_add(var_item_list, "RTS/DTR Pins", 3, line_flow_cb, app);
+    // GPIO pins PC0, PC1 (16,15) are unavailable for RTS/DTR when LPUART is
+    // selected.
+    int available_flow_pins = cfg_set->uart_ch == FuriHalUartIdLPUART1 ? 3 : 4;
```
skotopes: Nope. That's not enough.
See line_flow_cb and line_port_cb

skotopes (changes requested):

applications/main/gpio/scenes/gpio_scene_usb_uart_config.c
skotopes: Ensure that when you go to configuration it is impossible to set invalid combination.

skotopes:
Just in case: I still can not merge this PR because it allows to select invalid combination.

antimatter15:
I made some changes so it'll enforce non-overlap, let me know if there is a better way to do it.

skotopes (approved):

skotopes:
Nice work.

Issue #1691, opened 2022-09-01
Add UI option to clear NFC cache


no-usernames-left:
As a feature request, please consider adding a UI option to clear the NFC cache so that expanded dictionaries can be tried against cards that were previously read.

gornekich:
Hello @ThePowerOfDreams 
Why do you need to delete nfc cache for that? Saved keys caches allows to read already known sectors, and if card is still not fully read, dictionary attack starts after that

no-usernames-left:
>Saved keys caches allows to read already known sectors, and if card is still not fully read, dictionary attack starts after that

That's not the behaviour I'm seeing with a Mifare card with only some keys found; after one second or so, Flipper shows me 26/32 keys and 16/16 sectors, but will not actually retry it even when I tap left for Retry. (This is on 0.65.2.)

gornekich:
Got it, it's a bug. I don't see why we need to delete key cache, but we definitely have to start dictionary attack unless not all keys found. Will be fixes soon

Astrrra:
The core issue was fixed in #1868. Please re-open the issue if that was not enough.

Issue #1666, opened 2022-08-26
Interupted NFC read doesn't contiune


legomannetje:
### Describe the bug.

If you are just out of reach of an NFC card while reading a card (in my case MF classic) and while it's running the dict, it wont contiune after providing the card

### Reproduction

1. Start reading a MF classic card
2. While running the dict attack, remove the card
3. Place the card again. 
4. It won't contiune

### Target

NFC

### Logs

```Text
Don't have logs of it, can try to extract some, but don't know how atm
```

### Anything else?

Running the 0.65.2-RC firmware

skotopes:
Please check latest release

Astrrra:
Closing as fixed by #1868, please re-open if the issue is still there.

PR #1868, opened 2022-10-11
[FL-2876] MFC Improvements Part 2/2

Astrrra:
# What's new

- The dictionary attack starts if the flipper wasn't able to read the whole card while using key cache (solves #1691)
- The keys are no longer skipped after removing the card and placing it back (solves #1666)
- Fixes for part 1 improvements, now the "card removed" notification is shown even when checking for re-used keys

# Verification 
## Key cache
- Read any card
- Change some keys on it (by writing to it using a proxmark or an android phone with Mifare Classic Tool)
- Read the card again, it should read the keys that haven't been changed and proceed to bruteforce the other ones

## Key skipping and part 1 fixes
- Start reading a card with all default keys (FFFFFFFFFFFF)
- Remove the card mid-read and place it back
- The read should continue as if nothing happened

# Checklist (For Reviewer)


lib/nfc/nfc_worker.c
```diff
@@ -522,6 +542,18 @@ void nfc_worker_mf_classic_dict_attack(NfcWorker* nfc_worker) {
...
+        if(mf_classic_is_key_found(data, sector, MfClassicKeyA) &&
+           !mf_classic_is_sector_read(data, sector)) {
+            mf_classic_set_key_not_found(data, sector, MfClassicKeyA);
+        }
+        if(mf_classic_is_key_found(data, sector, MfClassicKeyB) &&
+           !mf_classic_is_sector_read(data, sector)) {
+            mf_classic_set_key_not_found(data, sector, MfClassicKeyB);
+        }
+    }
+
```
gornekich: I am not sure that it's correct. Imagine sector that can be read by key B, which is not in dictionary. Then will remove correctly found key A

gornekich (changes requested):
A few changes upcoming

lib/nfc/nfc_worker.c
Astrrra: Solved

gornekich (approved):

skotopes (approved):

PR #1888, opened 2022-10-18
Picopass: Read Elite

pcunning:
# What's new

- Finishes #1690 (cc @bettse)
- Adds CSN to the success display

# Verification 

- Load FAP
- Scan iClass Elite keyed card with key in `picopass/assets/iclass_elite_dict.txt` or `picopass/assets/iclass_elite_dict_user.txt`

Screenshot-20221017-190025

# Checklist (For Reviewer)

### DIFF
--- a/applications/plugins/picopass/helpers/iclass_elite_dict.c
+++ b/applications/plugins/picopass/helpers/iclass_elite_dict.c
@@ -1,0 +1,151 @@
+#include "iclass_elite_dict.h"
+
+#include <lib/toolbox/args.h>
+#include <lib/flipper_format/flipper_format.h>
+
+#define ICLASS_ELITE_DICT_FLIPPER_PATH EXT_PATH("picopass/assets/iclass_elite_dict.txt")
+#define ICLASS_ELITE_DICT_USER_PATH EXT_PATH("picopass/assets/iclass_elite_dict_user.txt")
+
+#define TAG "IclassEliteDict"
+
+#define ICLASS_ELITE_KEY_LINE_LEN (17)
+#define ICLASS_ELITE_KEY_LEN (8)
+
+struct IclassEliteDict {
+    Stream* stream;
+    uint32_t total_keys;
+};
+
+bool iclass_elite_dict_check_presence(IclassEliteDictType dict_type) {
+    Storage* storage = furi_record_open(RECORD_STORAGE);
+
+    bool dict_present = false;
+    if(dict_type == IclassEliteDictTypeFlipper) {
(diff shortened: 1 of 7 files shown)


applications/plugins/picopass/helpers/iclass_elite_dict.c
```diff
@@ -0,0 +1,151 @@
+#include "iclass_elite_dict.h"
+
+#include <lib/toolbox/args.h>
+#include <lib/flipper_format/flipper_format.h>
+
+#define ICLASS_ELITE_DICT_FLIPPER_PATH EXT_PATH("picopass/assets/iclass_elite_dict.txt")
```
bettse, line 6: I wonder if we want to make this ANY_PATH and include an empty file, or maybe the `5b7c62c491c11b39` value from proxmark?

applications/plugins/picopass/picopass_worker.c
```diff
@@ -197,11 +199,69 @@ ReturnCode picopass_read_card(PicopassBlock* AA1) {
...
+            return err;
+        }
+        memcpy(ccnr, rcRes.CCNR, sizeof(rcRes.CCNR)); // last 4 bytes left 0
+
+        loclass_iclass_calc_div_key(AA1[PICOPASS_CSN_BLOCK_INDEX].data, key, div_key, true);
+        loclass_opt_doReaderMAC(ccnr, div_key, mac);
+
+        err = rfalPicoPassPollerCheck(mac, &chkRes);
+        if(err == ERR_NONE) {
+            memcpy(pacs->key, key, PICOPASS_BLOCK_LEN);
```
bettse, line 250: _nice_

bettse: Future idea: save into the .picopass file

bettse (approved):

applications/plugins/picopass/helpers/iclass_elite_dict.c
pcunning: I think there is some refactoring that could happen and add other keys from pm3. It would be nice to detect unpersonalized/default key cards as well.

applications/plugins/picopass/picopass_worker.c
pcunning: It’s not updated when the primary key works and saving in the dump is kind of an incorrect representation of the data in that block.

bettse: I was thinking it would be like the PACS data in the .picopass file: In the header, not part of the dump of the card contents.

skotopes (approved):

PR #1920, opened 2022-10-24
[FL-2887] actions unit tests runner

doomwastaken:
# What's new

- Added basic unit tests runner with mock for device manager
- Currently hard-coded device ID for specific runner
- Currently arbitrary number for memory leak (until units are fixed, value - 3000)
- Units running only for PR

# Verification 

Setup:
- Need to setup new runner
- Need to connect physical device (F7 flipper) to an that runner
- Specify device name in scripts (or setup additional runners) 

In order to verify, just go to actions page and see that actions are triggered to correct actions (Pull Requests)

# Checklist (For Reviewer)


.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: 'Get flipper from device manager (mock)'
+        id: device
+        run: |
+          echo "flipper=/dev/ttyACM0" >> $GITHUB_OUTPUT
+
+      - name: 'Compile unit tests firmware'
+        id: compile
+        continue-on-error: true
+        run: |
+          FBT_TOOLCHAIN_PATH=/opt ./fbt FIRMWARE_APP_SET=unit_tests flash_usb FORCE=1
```
DrZlo13: flash_usb will not flash with flipper if it hangs in case of critical error.
You need to use a separate programmer for the flashing process.

doomwastaken: ran the tests after toolchain fixes, now it (temporarily, until we need to scale it) uses hard-coded serial to flash

skotopes (approved):

PR #1937, opened 2022-10-27
Mifare Classic nonce log

head47:
# What's new

Some readers use different keys, depending on the tag's UUID. In these cases, Detect Reader's output is useless for key calculation with tools like [mfkey32v2](https://github.com/equipter/mfkey32v2), and it is necessary to use debug logs for this purpose. However, these logs require Flipper to be connected to another device during emulation. It is inconvenient in some scenarios.

This PR introduces a `MFC nonce log` toggle found in a new submenu NFC -> Settings. When enabled, nt/nr/ar received during Mifare Classic emulation will be written to `/int/.mfkey32_emul.log`. The toggle's state is stored in `/int/.nfc.settings`.

# Verification 

- Navigate to NFC -> Settings, set `MFC nonce log` to ON
- Emulate a Mifare Classic tag
- Read `/int/.mfkey32_emul.log`, received nonces should appear there
- Navigate to NFC -> Settings again, set `MFC nonce log` to OFF
- Emulate a Mifare Classic tag
- Read `/int/.mfkey32_emul.log`, nothing new should appear there

# Checklist (For Reviewer)


skotopes:
@head47 thanks for PR, can you resolve merge conflict?

@gornekich review, would you kindly?)

max-baz:
This PR is probably obsolete now, since https://github.com/flipperdevices/flipperzero-firmware/pull/1941 we can use a saved card with not all keys found to launch "detect reader" with the card's UID.

gornekich:
Hello @head47 !
Thanks for suggested feature. As @maximbaz mentioned, we added the possibility to launch mfkey nonces collection from saved card menu. I guess it solves your issue completely. Do you agree?

By the way, writing nonces to internal storage is not a good idea. The file size can be large and fill flash memory very quickly.

head47:
@gornekich yes, I agree with that. I guess that's a lesson for me to take a look at the dev builds before writing anything :)

PR #1965, opened 2022-11-01
infrared: add Kaseikyo IR protocol

### LINKED ISSUE #1199: Kaseikyo IR protocol
**Is your feature request related to a problem? Please describe.**

Currently any IR remote using the Kaseikyo protocol (e.g. Denon) is not supported. It works with raw IR support, but having a specific codec would make it possibly (easier) to construct the .ir using the manufacturers documentation without needing to record every button press or if the remote is not available.

**Describe the solution you'd like**

Encoder and decoder for the Kaseikyo IR protocol as described in the following links:

- https://github.com/Arduino-IRremote/Arduino-IRremote/blob/master/src/ir_Kaseikyo.hpp
- http://www.hifi-remote.com/johnsfine/DecodeIR.html#Kaseikyo
- https://www.denon.com/-/media/files/documentmaster/denonna/avr-x3700h_avc-x3700h_ir_code_v01_04062020.doc

**Describe alternatives you've considered**

Generating RAW ir files is possible but tedious and error prone.

**Additional context**

I've implemented encoders and decoders and can open a PR if what I've implemented looks acceptable and there isn't already a codec being worked on.

https://github.com/samuel/flipperzero-firmware/tree/kaseikyo-ir

samuel:
# What's new

Add Kaseikyo IR protocol support. This protocol is also called the Panasonic protocol and is used by a number of manufacturers including Denon.

The protocol includes a vendor field and a number of fields that are vendor specific. To support the format of address+command used by flipper the vendor+genre1+genre2+id fields are encoded into the address while the data is used for the command.

There are older versions of the protocol that used a reverse bit order that are not supported.

Protocol information:
- https://github.com/Arduino-IRremote/Arduino-IRremote/blob/master/src/ir_Kaseikyo.hpp
- http://www.hifi-remote.com/johnsfine/DecodeIR.html#Kaseikyo
- https://www.denon.com/-/media/files/documentmaster/denonna/avr-x3700h_avc-x3700h_ir_code_v01_04062020.doc

Resolves https://github.com/flipperdevices/flipperzero-firmware/issues/1199

# Verification 

- Using a Kaseikyo remote (e.g. Denon receiver) receive transmissions and see the code.
- Retransmit the received codes to the device to verify proper encoding.

# Checklist (For Reviewer)


gsurkov (changes requested):
Thanks!
Could you please add unit test for the Kaseikyo protocol?

**How to run unit tests:**
1. Copy the `assets/unit_tests` folder to your Flipper's SD card.
2. Compile the firmware with unit tests enabled: `./fbt FIRMWARE_APP_SET=unit_tests` and flash it using your preferred method.
3. Open the CLI session and run `unit_tests infrared`.

**How to add unit tests:**
1. Create a file named `assets/unit_tests/infrared/test_kaseikyo.irtest`.
2. Fill it with the test data (more on it below).
3. Write the test code in `applications/debug/unit_tests/infrared/infrared_test.c`.
4. Update the `assets/unit_tests` folder with the new .irtest file and run the tests. They should pass obviously.

**Test data format**
Infrared `.irtest` files are basically regular `.ir` files with a few additions.
Decoder input (raw) data has signal names `decoder_input_N` where N is a number of the sequence. Naturally, the expected data goes under the name `decoder_expected_N`. When testing the encoder these two are switched.
Decoded data is represented in arrays (since a single raw signal may decode to several messages). If there is only one signal, then it has to be an array of size 1.
Please use the existing files as an example.

After making any changes to the code, please do not forget to run `./fbt format` to ensure that it stays properly formatted. Add and commit any changes that were made by this command.

samuel:
Thanks for the details on adding unit tests.

I've added the unit tests and formatted the files.

gsurkov (approved):

hedger (approved):

PR #2037, opened 2022-11-21
Added log to file to Subghz Frequency Analyzer

cberetta:
# What's new
Added ability to save the captured frequency and rssi to a log file.
This can be enabled or disabled by renaming a file in the subghz folder: if a file named `fqalogon.txt` is found the logging to file is enabled. If file is renamed to `fqalogoff.txt` logging is disabled. By using a filename as a flag you can enable and disable logging directly on the flipper with the file browser.
At the first startup of the frequency analyzer it would create _(if necessary)_ a default file named `fqalogoff.txt` _(by default logging is disabled)_. The logging is done in files with the capture date used in the name, so every day a new file will be created. The file is also constructed so that it can be easily imported into your preferred spreadsheet for subsequent analysis. The log file is basically a CSV but with a TXT extension.

# Verification 
- Run Subghz Frequency Analyzer
- On first run, the file `fqalogoff.txt` will be created in `subghz` folder
- Use browser _(down button)_ to find and rename file to `fqalogon.txt`
- Run againg Subghz Frequency Analyzer and all captured frequency will be saved to files into `subghz` folder

# Checklist (For Reviewer)

### DIFF
--- a/applications/main/subghz/views/subghz_frequency_analyzer.c
+++ b/applications/main/subghz/views/subghz_frequency_analyzer.c
@@ -35,6 +35,7 @@ struct SubGhzFrequencyAnalyzer {
     void* context;
     bool locked;
     uint32_t last_frequency;
+    bool log_to_file;
 };
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


skotopes:
And use cases are?

cberetta:
> And use cases are?

I think that it could be useful mainly for statistical purposes, knowing which are the most used frequency in an area could be useful. Or you can use it to track the distance that a fob can reach without the needs of a friend that check the Flipper while you are pressing the fob button while stepping away.

skotopes:
What you've described is not going to work: frequency analyzer intentionally disconnects transceiver from antenna, so only close proximity signals can be received(1 inch close). I'll close this PR, there is no meaningful use for this feature.

LowSkillDeveloper:
@cberetta I think that this is not a bad idea and maybe it can be implemented, but in unofficial firmware, for example in [unleashed-firmware](https://github.com/DarkFlippers/unleashed-firmware).
Try submit your idea there if you want.

PR #2055, opened 2022-11-26
POCSAG protocol decoder for subghz

Shmuma:
# What's new

- POCSAG protocol used in pager networks (https://hampager.de/#/). All four types of messages (functions) are supported: BCD, alpha-numeric, alert1 and alert2. CRC of codewords are not validated (yet).

Technically, this is just a first step towards "flipper as a pager" activity and proper FAP is planned for this. But having POCSAG decoder in codebase might still be useful:

- for procotols experimentation
- as an example of fixed-rate protocol decoder. As far as I know, this is the only protocol of this sort in flipper, all others are using pulse-width.

POCSAG requires 9-15kHz bandwidth, which is not compatible with default FM modulations. File `setting_user.pocsag` contains example of custom CC1101 preset to receive POCSAG messages.

# Verification 

- You can use provided raw recordings (with all 4 kinds of messages). To test POCSAG "in the wild", MMDVM RPi hat could be used to send the messages over the radio. 

# Checklist (For Reviewer)


hedger:
- This protocol requires a non-standard modulation profile that's not a part of main firmware
- SubGHz GUI is suboptimal for displaying text-based messages.

I think this PR should be converted to a separate paging app, similar to Weather Station.

Shmuma:
> I think this PR should be converted to a separate paging app, similar to Weather Station.

Fully agree. I believe it will happen eventually. But this is more convenient for testing to have this protocol in mainstream rather than is some separate fork.

In the end, weather protocols were also implemented in subghz first and then migrated into separate app.

skotopes:
We'd rather to see it as separate app from the beginning. WS was initially implemented in subghz app only because there were no appropriate APIs available for faps.

PR #2070, opened 2022-12-01
added DigitalSequence and PulseReader

g3gg0:
added DigitalSequence to chain multiple DigitalSignals
added PulseReader for hardware assisted digital signal sampling

# What's new

- adds DigitalSequence to queue multiple DigitalSignal in a scripted manner to reducre precalculation effort
- adds PulseReader to sample GPIOs whenever their state change and capture TIM2 value along with it

# Verification 
Maybe with such a test code? Where to put smth like that?
```
    /* read pulses from PB2 */
    PulseReader *reader_signal = pulse_reader_alloc(&gpio_ext_pb2, 512);
    pulse_reader_set_timebase(reader_signal, PulseReaderUnitNanosecond);
    pulse_reader_start(reader_signal);
    while(true) {
        uint32_t length= pulse_reader_receive(reader_signal, timeout_ms * 1000);
        FURI_LOG_D(TAG, "Pulse: %lu", length);
    }
    pulse_reader_stop(reader_signal);

```

# Checklist (For Reviewer)


lib/digital_signal/digital_signal.c
```diff
@@ -72,102 +98,377 @@ uint32_t digital_signal_get_edges_cnt(DigitalSignal* signal) {
...
+
+    if(sequence->sequence_used >= sequence->sequence_size) {
+        sequence->sequence_size += 256;
+        sequence->sequence = realloc(sequence->sequence, sequence->sequence_size);
+    }
+
+    sequence->sequence[sequence->sequence_used++] = signal_index;
+}
+
+void digital_signal_update_dma(DigitalSignal* signal) {
```
g3gg0: I know this looks a bit low level.
We also can use a C-only version but then minimum length of a the *last* pulse in a signal has to be a bit longer.

lib/digital_signal/digital_signal.c
```diff
@@ -72,102 +98,377 @@ uint32_t digital_signal_get_edges_cnt(DigitalSignal* signal) {
...
+        /* R0 */ (uint32_t)&(DMA1_Channel2->CCR),
+        /* R1 */ DMA1_Channel2->CCR & ~DMA_CCR_EN,
+        /* R2 */ (uint32_t)signal->reload_reg_entries,
+        /* R3 */ (uint32_t)&(TIM2->ARR),
+        /* R4 */ (uint32_t)signal->reload_reg_buff,
+        /* R5 */ DMA1_Channel2->CCR | DMA_CCR_EN };
+
+    
+    /* hurry when setting up next transfer */
+    asm volatile("\t"
```
g3gg0: Also thought about using DMAs to automate things, but that would probably occupy all available DMA channels of both DMA.
And even then I am not sure if it would work out.

g3gg0:
ouch. i guess i know what to do...

gornekich:
Hi @g3gg0 
Could you please also commit firmware/targets/f7/api_symbols.csv file. You can replace '?' field with '-' for now

g3gg0:
Ah cool, it seems you are doing PIL tests with a flipper attached?

Currently I am trying to get the unit test setup running on my setup.
I think I will have it ready in a few minutes, but is there some documentation / hints about this topic?

EDIT: dumb question: how to get the debug log in unit_test mode?

g3gg0:
Ok, was due to signal->gpio being uninitialized, where the prepare code used it to configure the GPIO mask.
That should be updated in the unit_test file to also check signal->gpio_buff to be properly initalized.

Then we are left with a difference in the timings, which i have to check. Maybe that's due to the trailing level (?)
grafik

As I said, I have to look closer.

gornekich:
Hello @g3gg0 
Thanks for this PR! I can also see that nfc_digital_signal_test unit test fails. Could you have a look on it?
Some hints regarding unit tests: you can build firmware with unit tests included via ./fbt FIRMWARE_APP_SET=unit_tests More details you can find in documentation/fbt.md
The default way to read logs is to connect Flipper GPIO 13 (TX) to uart-usb converter. The UART baudrate is 230400

g3gg0:
okay, took me some days to get back to work on it.
yes, it is due to the trailing level.

Let's get more into detail
NfcA code is "asking" to generate pulses of a specific length by concatenating them using digital_signal_append().
when normalizing the µs values to the smallest (NfcA) pulse length for readability, we get those patterns:
grafik
"one" is the one-bit, "zero" the zero-bit. both consist of 1 long period and 8 short pulses, changing modulation strength.
Due to polarities, for the "one", one short and the long period merge:
grafik


In our discussion nov 21 on telegram, we talked about two interpretations of the code:

grafik
vs
grafik

My first proposal was A, giving just the guarantee that the pulse will be *at minimum* duration d.
You said the latter (B) is the one you expect, which I agree, because it feels more natural if you use the code to generate pulses.
Just when appending signals it might confuse users a bit more than variant A as you might want to explain: "digital_signal_append() will *extend* the last pulse when the signal to append starts with the same level"

However the test data seems to show a whole different interpretation, as the last zero bit has only *7* pulses.
So it neither does a trailing polarity change as in B, nor it would guarantee a minimum pulse length like in A.
It would just check for generating a polarity change without respecting the edge duration.

grafik

So I have some questions:
a) is there some specification on which the test is based on, or was it based on the first (reference) implementation
b) shall we change the test to expect "...641 36 37 37 37 37 36 37 37" to match variant B?
c) or am I missing something? :)

skotopes:
@gornekich @g3gg0 un-draft when ready. ;-)

g3gg0:
@gornekich 

shall i change the test files to match either A or (preferrably) B ?

g3gg0:
okay i changed the status back to "open" for a review.

- I have changed the test case to match the expected behavior
- not sure about the API version i should have picked

gornekich:
Hi @g3gg0 
Thanks for changes. Very sorry for late responses, I will test PR today

g3gg0:
no worries.
is there anything i can help with - demo apps or so?

gornekich:
Great work!
Digital Signal and Sequence work just fine.
While testing PulseReader I noticed strange behavior. If I run the code you suggested for testing and generate 1 Hz square wave with 50% duty cycle, I receive these logs:

25207 [D][PulseReader] Pulse: 131
26203 [D][PulseReader] Pulse: 999854
26205 [D][PulseReader] Pulse: 132
27202 [D][PulseReader] Pulse: 999854
27204 [D][PulseReader] Pulse: 134
28201 [D][PulseReader] Pulse: 999852
28203 [D][PulseReader] Pulse: 153
29200 [D][PulseReader] Pulse: 999833
29201 [D][PulseReader] Pulse: 124
30199 [D][PulseReader] Pulse: 999862
30200 [D][PulseReader] Pulse: 129
The timebase is set to PulseReaderUnitMicrosecond

I expected to see roughly the same values. Am I wrong?

gornekich:
Also thanks for changing unit tests data. I agree with your solution.

Could you also modify build scripts so that PulseReader is build and installed as static library. You can apply the patch attached
[static_lib_patch.txt](https://github.com/flipperdevices/flipperzero-firmware/files/10269504/static_lib_patch.txt)

g3gg0:
> If I run the code you suggested for testing and generate 1 Hz square wave with 50% duty cycle, I receive these logs:

uh thats weird. looks like 
    LL_DMAMUX_SetRequestGenPolarity(NULL, LL_DMAMUX_REQ_GEN_0, LL_DMAMUX_REQ_GEN_POL_RISING);
would have been set to LL_DMAMUX_REQ_GEN_POL_RISING_FALLING
the polarity change creates an EXTI, which is a pulse with approximately the duration you have seen IIRC

how did you create the signal? using an external device or with the device creating the signal itself?
guess you are sure the pulses were 50% duty?

will try a similar setup. maybe there is something odd due to timer overruns or such...

g3gg0:
also: the code checks the polarity everytime you get an pulses's duration
grafik

if the polarity didn't change (because the edge was shorter than the EXTI pulse itself), you get PULSE_READER_LOST_EDGE.
so if you are sure you had 50% duty, then i would have said its a fluke in the code, but that check should have triggered and telling you.

gornekich:
I just took another flipper and generated signal with Signal generator application. I will check with logic analyzer the signal just to be sure.

g3gg0:
Finding: the pulse_reader's first returned signal was a missing edge. fixed that.

Adding a note that interrupts must not get disabled when using pulse_reader, as it uses EXTI, 
which is an interrupt line which would get blocked...

Also added a unit test routine, which I am cleaning up right now.

g3gg0:
okay, added a unit test, creating different signals using TIM1 that will get read by digital_reader. 

for this PA6 and PA7 need to get bridged.
we might have to remove the test anyway, as it doesn't seem to be too stable.
i guess the scheduling kicks in and causes a delay in the code that counts the "sent" pulses and stops the TIM1.
this delay makes the tests of 1-2ms cycles too unstable. 10ms and above seem to work fine.

can you test it if it works for you?

theses the digital pulses that are generated:
grafik

g3gg0:
> I just took another flipper and generated signal with Signal generator application. I will check with logic analyzer the signal just to be sure.

@gornekich  Can you double-check the signal generation? for me it looks like it is working as intended.
Just the very first signal was a missed edge signal, which is fixed now.

lib/pulse_reader/pulse_reader.c
```diff
@@ -0,0 +1,202 @@
...
+    uint32_t dma_pos = signal->size - (uint32_t)LL_DMA_GetDataLength(DMA1, signal->dma_channel);
+
+    return ((signal->pos + signal->size) - dma_pos) % signal->size;
+}
+
+void pulse_reader_stop(PulseReader* signal) {
+    LL_DMA_DisableChannel(DMA1, signal->dma_channel);
+    LL_DMA_DisableChannel(DMA1, signal->dma_channel + 1);
+    LL_DMAMUX_DisableRequestGen(NULL, LL_DMAMUX_REQ_GEN_0);
+    LL_TIM_DisableCounter(TIM2);
```
gornekich, line 137: Please, reset gpio mode here. 
furi_hal_gpio_init_simple(signal->gpio, GpioModeAnalog);
is enough

gornekich (commented):
@g3gg0 the issue with signal generation was on my side. It works amazing!
I managed to capture signal up to 410 kHZ. I think we can do more if we add more asynchronous API. I was thinking about pulse reader mode when you set the buffer size, timeout and callback. After buffer is filled or timeout reached, the callback is called and user application can copy data from ring buffer. That will help to avoid LL_DMA_GetDataLength() calls in pulse_reader_receive(). I am not completely sure, but may be that calls slow down internal DMA work.
Also during signal capturing the CPU should avoid working with memory. Since DMA performs memory to peripheral transfers I guess there are some "hardware mutexes" which synchronize memory related operation. That may help us to capture faster signals.

While there are some interesting things to research, your code is working and I would love to see it in our firmware.

g3gg0:
i am happy to hear that :)

One limitation is the pulse length of the EXTI interrupt line.
Also if a asynchronously running code is disabling interrupts, we get in trouble.
This unfortunately limits the usability to some degree, unless we can find a way out of how the pin change gets routed.

I think it's a great feature, showing off it's usability in the ISO15693 code that will come after this has been merged.

gornekich:
@g3gg0 Could you please format python sources :)
run `./fbt format_py`

g3gg0:
@gornekich done, seemed a missing newline issue.

g3gg0:
yeah signal generation seems to be too unstable due to jitter. shall we drop the unit test or change it so the number of pulses isn't checked?

gornekich:
@g3gg0 yes, let's remove this test for now. The unit tests are run automatically on the Flipper, there is no way to connect required pins.

The last thing I am wondering is that your PR added ~2Kb of used flash memory. That's not what I expected, since you made a few changes in digital signal lib functions that are actually used in firmware. Maybe linker doesn't remove unused functions on firmware build. Need some time to figure this out, and we are ready to merge

gornekich:
@g3gg0 Since we added pulse_reader to API fbt links pulse reader lib in firmware. It weights ~1Kb. Now we try to use flash memory as efficiently as possible and extra 1Kb of code which is not used so far is not good.

We can build pulse_reader static library, add header to SDK but exclude symbols from API for now. This will allow users to write applications and link with pulse_reader library. I attach the patch which makes these changes: 
[remove_pulse_reader_from_api.txt](https://github.com/flipperdevices/flipperzero-firmware/files/10280612/remove_pulse_reader_from_api.txt)

g3gg0:
@gornekich done
As I am quite new to the build system and how the .fap stuff works, I can just blindly follow what you say ;)

gornekich (dismissed):
Thank you!

g3gg0:
great, thank you too :)

g3gg0:
@gornekich 
is it okay to add another API until it is fully reviewed?

grafik

If that is included, ISO15693 will most likely also allow AM mode instead of OOK

gornekich:
@g3gg0 yes, I think it's OK
We will merge you PR after we publish release

g3gg0:
@gornekich great, do you think ISO15693 support would be also welcome for this release?

lib/digital_signal/digital_signal.c
skotopes: I don't think that C version will be slower here. This code is quite straight forward, no special instructions used. Please provide C-version.

lib/pulse_reader/pulse_reader.h
```diff
@@ -0,0 +1,141 @@
...
+ * unit of the edge durations to return
+ */
+typedef enum {
+    PulseReaderUnit64MHz,
+    PulseReaderUnitPicosecond,
+    PulseReaderUnitNanosecond,
+    PulseReaderUnitMicrosecond,
+} PulseReaderUnit;
+
+typedef struct {
```
skotopes: Do you need to expose this structure here? Can it be anonymous?

lib/digital_signal/digital_signal.h
```diff
@@ -5,23 +5,53 @@
...
 extern "C" {
 #endif
 
+/* helper for easier signal generation */
+#define DIGITAL_SIGNAL_MS(x) (x * 100000000UL)
+#define DIGITAL_SIGNAL_US(x) (x * 100000UL)
+#define DIGITAL_SIGNAL_NS(x) (x * 100UL)
+#define DIGITAL_SIGNAL_PS(x) (x / 10UL)
+
 typedef struct {
```
skotopes: Does it has to be public?

skotopes (changes requested):
Couple things that is bothering me:

- STM LL HAL leaking into headers
- No proper encapsulation, structs exposing their content
- ASM where C works fine

skotopes:
@g3gg0 great work!

There are couple thing that need to be improved, feel free to ping me on discord if you have questions.
And un-draft when ready.

lib/digital_signal/digital_signal.h
g3gg0: these were meant as helper macros to provide users an easy way to call code like:
`sig->edge_timings[0] = DIGITAL_SIGNAL_US(100);`
it just helps with unit conversion, not requiring calculation, but it's not required and i can remove it if you want.

lib/pulse_reader/pulse_reader.h
g3gg0: it is required here:
grafik

lib/digital_signal/digital_signal.h
skotopes: I meant PulseReaderUnit itself. Basically this structure exposes its internals which makes difficult future porting.

Couple examples how we usually do that:

Public API provided in header file contains anonymous structure and provides getters/setters/methods to work with it: https://github.com/flipperdevices/flipperzero-firmware/blob/dev/applications/services/gui/view.h#L34
Everything else is in implementation or in private API.

lib/pulse_reader/pulse_reader.h
g3gg0: > digital_signal_update_dma

In fact the asm version is faster than any C version.
It prepares all data to be written into registers, which get written into memory using a single-fetch, multiple write STM operation right after the "transaction finished" bit flipped.

Speaking in pipelining, there is just a single FETCH and DECODE pipeline step, but only multiple cycles in EXECUTE step.
This was required to get even NfcA emulation working fully in software.
Without this optimization, the short pulses in NfcA modulation were too short to set up the next DMA transaction.

> structure

of course this stucture can be made anonymous. i just made it similar to the digital_signal paradigm.
but will make it anoynmous.

lib/digital_signal/digital_signal.h
g3gg0: absolutely. will change it.

lib/pulse_reader/pulse_reader.h
skotopes: It's not only about performance but also about portability. Please provide C version alongside so we'll be able to compare them ;-)

g3gg0: Sure, will provide it.
But keep in mind that the key is the latency between "DMA finished" and "next DMA started" is the key.
grafik

If the calculations I did are correct, we have 37 clock cycles from a finished DMA to DMA being started again.
Minus some triggering delays. This caused emulating NfcA this way not to work if I used some low level C code.

g3gg0:
@skotopes 
okay now i made the structures as privately as possible :)
also added the C implementation for reference.

will test it now using ISO15693 emulation and apply fixes if required

g3gg0:
@skotopes 
okay now i made the structures as privately as possible :)
also added the C implementation for reference.

also tested the ISO15693 emulation and works as expected

lib/digital_signal/digital_signal.c
```diff
@@ -80,94 +160,385 @@ uint32_t digital_signal_get_edge(DigitalSignal* signal, uint32_t edge_num) {
...
+        sequence->sequence_size += 256;
+        sequence->sequence = realloc(sequence->sequence, sequence->sequence_size);
+    }
+
+    sequence->sequence[sequence->sequence_used++] = signal_index;
+}
+
+#if defined(DIGITAL_SIGNAL_PORTABLE_CODE)
+
+static void digital_signal_update_dma(DigitalSignal* signal) {
```
skotopes: C version logic doesn't match to ASM version, there is a missing section:
```
    LL_DMA_ClearFlag_TC1(DMA1);
    LL_DMA_ClearFlag_TC2(DMA1);
```

Also I've checked instructions generated in both cases. You ASM is actually twice bigger and much slower because it uses RAM for dmaX_data.

Also this function is a tiny bit of digital_sequence_send/digital_sequence_send_signal and it make no sense to optimize only this specific part.

skotopes (changes requested):

lib/digital_signal/digital_signal.c
g3gg0: i am so sorry, you are absolutely right. 
unfortunately i pushed an incomplete C code. didnt notice because i flashed an old version...

here the disassembly of the two variants next to each other.
grafik

it clearly shows the larger footprint you noticed, but also the lower re-enable latency if the DMA finished its transfer.

g3gg0: okay optimized a bit more
grafik

instruction count difference is 13 instructions, however i am not happy with the stack usage. could be optimized.

g3gg0: okay further optimized the code.
grafik
i am skipping DMA channel 1 initialization. will have to double-check that there are no side effects.
now the C implementation (right) might be in the usable range of delay. did not check with NfcA emulation to prove this is working out.

g3gg0:
@skotopes 
I think I addressed all issues you mentioned. Is there anything else I can improve?

g3gg0:
I am now reworking it to use a single DMA transfer and using a ringbuffer.
The C variant of that former asm-routine is nearly everytime quick enough.
Nearly - but not guaranteed, which is a problem for reliability.

Reworking it to use a ringbuffer and a circular DMA, so that the average CPU speed matters, not the worst case latency.

skotopes:
Please un-draft when ready. And feel free to ping me on discord if you need any help.

nvx:
From my testing the latest commit (https://github.com/flipperdevices/flipperzero-firmware/pull/2070/commits/ae1e93e519d6289f67d83ae0f114ca100ecc347f) fixes the intermittent failures I was seeing previously and I haven't noticed any other issues.

Is there anything still planned for this @g3gg0 or is it worth un-marking as a draft to get the ball rolling again on merging this?

g3gg0:
there is nothing to fix I am aware of, so in theory we could go on.
will look over it and un-draft

xMasterX:
> there is nothing to fix I am aware of, so in theory we could go on. will look over it and un-draft

Hello,
Any updates planned on this? Also same for NFC V support #2316

maybe-hello-world:
> there is nothing to fix I am aware of, so in theory we could go on. will look over it and un-draft

@g3gg0  Good day, are there any roadblocks for this PR? 
Many people would kindly appreciate this PR (and #2316) being merged. :smile: 
Thank you.

nvx:
Rebased onto latest dev and retested with the ISO15693 branch successfully as well. Should be ready for merge now.

g3gg0:
thank you @nvx for your support!

@skotopes @gornekich 
hoping for a review and eventually a merge :)

g3gg0:
@nvx can you fix this?

API versions aren't matching for available targets. Please update!
==> firmware/targets/f18/api_symbols.csv <==
entry,status,name,type,params
Version,+,23.0,,

nvx:
> @nvx can you fix this?
> 
> API versions aren't matching for available targets. Please update! ==> firmware/targets/f18/api_symbols.csv <== entry,status,name,type,params Version,+,23.0,,

I noticed this, what is f18 though? I assume there's a script or something that updates this like the normal f7 target rather than editing the file by hand.

@skotopes any guidance or docs on how to handle the f18 stuff properly?

skotopes:
> > @nvx can you fix this?
> > API versions aren't matching for available targets. Please update! ==> firmware/targets/f18/api_symbols.csv <== entry,status,name,type,params Version,+,23.0,,
> 
> I noticed this, what is f18 though? I assume there's a script or something that updates this like the normal f7 target rather than editing the file by hand.
> 
> @skotopes any guidance or docs on how to handle the f18 stuff properly?

`./fbt TARGET_HW=18` will verify symbols and then ensure that api version number is same for F7 and F18

nvx:
Updated f18 api_symbols.csv and rebased on latest dev again. Hopefully the workflows like that better. Will catch up the ISO15693 branch after this is merged prior to un-drafting that PR.

lib/digital_signal/digital_signal.h
```diff
@@ -10,18 +10,35 @@
 extern "C" {
 #endif
 
-typedef struct {
+/* helper for easier signal generation */
+#define DIGITAL_SIGNAL_MS(x) (x * 100000000UL)
```
skotopes: image

skotopes (changes requested):

skotopes:
Un-draft when ready.

lib/digital_signal/digital_signal.h
g3gg0: knowing it better and still having written this code, bugs me a bit :)
also thought i already fixed it somewhere.

g3gg0: not sure why it thinks the expression in digital_signal.c:482 is always false.

the code caches dma_buffer->read_pos and checks for a change after calling digital_sequence_update_pos() which updates dma_buffer->read_pos.
am i missing something?

grafik

lib/digital_signal/digital_signal.c
```diff
@@ -80,94 +179,480 @@ uint32_t digital_signal_get_edge(DigitalSignal* signal, uint32_t edge_num) {
...
+DigitalSequence* digital_sequence_alloc(uint32_t size, const GpioPin* gpio) {
+    furi_assert(gpio);
+
+    DigitalSequence* sequence = malloc(sizeof(DigitalSequence));
+
+    sequence->gpio = gpio;
+    sequence->bake = false;
+
+    sequence->dma_buffer = malloc(sizeof(struct ReloadBuffer));
+    sequence->dma_buffer->size = 32;
```
gornekich: Please avoid magic numbers. Define macros with explanation

lib/digital_signal/digital_signal.c
```diff
@@ -80,94 +179,480 @@ uint32_t digital_signal_get_edge(DigitalSignal* signal, uint32_t edge_num) {
...
+    sequence->dma_config_timer.MemoryOrM2MDstIncMode = LL_DMA_MEMORY_INCREMENT;
+    sequence->dma_config_timer.PeriphOrM2MSrcDataSize = LL_DMA_PDATAALIGN_WORD;
+    sequence->dma_config_timer.MemoryOrM2MDstDataSize = LL_DMA_MDATAALIGN_WORD;
+    sequence->dma_config_timer.PeriphOrM2MSrcAddress = (uint32_t) & (TIM2->ARR);
+    sequence->dma_config_timer.MemoryOrM2MDstAddress = (uint32_t)sequence->dma_buffer->buffer;
+    sequence->dma_config_timer.NbData = sequence->dma_buffer->size;
+    sequence->dma_config_timer.PeriphRequest = LL_DMAMUX_REQ_TIM2_UP;
+    sequence->dma_config_timer.Priority = LL_DMA_PRIORITY_HIGH;
+
+    digital_sequence_alloc_signals(sequence, 32);
```
gornekich: same here

lib/digital_signal/digital_signal.c
```diff
@@ -80,94 +179,480 @@ uint32_t digital_signal_get_edge(DigitalSignal* signal, uint32_t edge_num) {
...
+
+    return sequence;
+}
+
+void digital_sequence_free(DigitalSequence* sequence) {
+    furi_assert(sequence);
+
+    if(!sequence) {
+        return;
+    }
```
gornekich: I guess if() is not necessary here

lib/digital_signal/digital_signal.c
```diff
@@ -80,94 +179,480 @@ uint32_t digital_signal_get_edge(DigitalSignal* signal, uint32_t edge_num) {
...
+    sequence->send_time = send_time;
+    sequence->send_time_active = true;
+}
+
+void digital_sequence_add(DigitalSequence* sequence, uint8_t signal_index) {
+    furi_assert(sequence);
+    furi_assert(signal_index < sequence->signals_size);
+
+    if(sequence->sequence_used >= sequence->sequence_size) {
+        sequence->sequence_size += 256;
```
gornekich: Please add another macros here.

lib/pulse_reader/pulse_reader.c
```diff
@@ -0,0 +1,237 @@
...
+void pulse_reader_set_pull(PulseReader* signal, GpioPull pull) {
+    signal->pull = pull;
+}
+
+void pulse_reader_free(PulseReader* signal) {
+    furi_assert(signal);
+
+    if(!signal) {
+        return;
+    }
```
gornekich: consider to remove one of these checks

gornekich (changes requested):
Thank you!
Please consider my suggestions.
Also I attach a patch that suppresses static code analyzer warnings
[fixes.txt](https://github.com/flipperdevices/flipperzero-firmware/files/11395598/fixes.txt)

gornekich:
Also I noticed that mf classic emulation works worse than dev brunch. I will try to figure it out

g3gg0:
> Also I noticed that mf classic emulation works worse than dev brunch. I will try to figure it out

wait. you mean that these changes make mf classic emulation worse?

gornekich:
Yes, but it's partially my fault. I have log level set to debug and  message "[prepare] pulse_duration out of range" makes timings worse. When I set debug level to info, the emulation is exactly like in dev brunch.

Now I want to figure out why we fall in `if(pulse_duration < 10 || pulse_duration > 10000000)` check during mf classic emulation. It looks like we prepare signal->edge_timings not correctly. Sometimes the last value in edge_timings buffer is 0x0a, which cause problems in filling reload_reg_buff

g3gg0:
at some point in time in my dev branches i moved the NfcA code over to the DigitalSequence stuff.
didn't notice it there - or - accidentally fixed it along with rewriting :D

g3gg0:
https://github.com/g3gg0/flipperzero-firmware/commit/0cb784aadd344865bbf71a5320abbeb466fb0775
there i have undone my own rewriting
and there is the source:
https://github.com/g3gg0/flipperzero-firmware/blob/c1b5394b0f94d52d2b1867aadacde85db47dc66b/lib/nfc/protocols/nfca.c

lib/digital_signal/digital_signal.c
```diff
@@ -80,94 +179,480 @@ uint32_t digital_signal_get_edge(DigitalSignal* signal, uint32_t edge_num) {
...
+        return;
+    }
+
+    /* Configure gpio as output */
+    signal->internals->gpio = gpio;
+    furi_hal_gpio_init(
+        signal->internals->gpio, GpioModeOutputPushPull, GpioPullNo, GpioSpeedVeryHigh);
+
+    /* single signal, add a temporary, terminating edge at the end */
+    signal->edge_timings[signal->edge_cnt++] = 10;
```
gornekich: @g3gg0 I thing the problem is here. For Mf Classic emulation this extra edge leads to incorrect pulse_duration set. Without this line emulation works much better.
Why do you add this?

gornekich:
@nvx Sorry for incorrect patch. Please format sources
./fbt format

lib/digital_signal/digital_signal.c
g3gg0: IIRC this was due to the tests assumed there is a trailing edge.
see the discussion there:
https://github.com/flipperdevices/flipperzero-firmware/pull/2070#issuecomment-1344861543
so i added a small edge, causing the polarity to switch

g3gg0: does this introduce a 10-tick delay or does it cause a unwanted polarity switch?

g3gg0: thinking about it. 
as it is only used from existing code, then maybe changing forth and back stuff made me think this end edge is required.
probably it is not required anymore and just wasn't noticed that this edge is not needed anymore.

@nvx 
if you remove that line, none of your code is affected, right?

nvx: I wouldn't imagine so if the rest of the ISO15693 code still works (my picopass code doesn't go that low level). Happy to remove it for now and we can re-test

g3gg0: i also found only one location where this function is used, which is the [furi_hal_nfc.c](https://github.com/flipperdevices/flipperzero-firmware/blob/0a3ff7f85ae653629957dfdd970a6fd3310ad14e/firmware/targets/f7/furi_hal/furi_hal_nfc.c) code.
so i **agree** in removing this trailing edge.

i hope the 10 ticks are not troublesome due to the timing get messed up (delaying the next signal)?

gornekich:
@nvx @g3gg0 thanks for the PR and fixes. Please undraft if no changes will be done

gornekich (dismissed):

skotopes (approved):

PR #2113, opened 2022-12-09
Add Mouse Jiggler to HID Remote

MrLemur:
# What's new

- Adds a new remote "Mouse Jigggler" for USB and Bluetooth. Moves the mouse back and forth when started (inspired from this [repo](https://github.com/jplesner/flipperzero-firmware-mouse-jiggler))
- Tries to use same functions as in existing mouse  app

# Verification 

- Either on USB or Bluetooth remote, run the remote and press Start.

# Checklist (For Reviewer)


applications/plugins/hid_app/views/hid_mouse_jiggler.c
```diff
@@ -0,0 +1,112 @@
...
+        elements_multiline_text_aligned(canvas, 91, 36, AlignLeft, AlignBottom, "Start");
+    }
+    canvas_set_color(canvas, ColorBlack);
+
+    // Back
+    canvas_draw_icon(canvas, 74, 49, &I_Pin_back_arrow_10x8);
+    elements_multiline_text_aligned(canvas, 91, 57, AlignLeft, AlignBottom, "Quit");
+
+    if(running) {
+        hid_hal_mouse_move(parentJiggler->hid, MOUSE_MOVE_SHORT, 0);
```
skotopes: View/ViewPort draw call is meant only for drawing, no API calls allowed in its context.
Please rewrite this code in a way when hid_hall only called from application main thread.

skotopes (changes requested):

skotopes:
Please un-draft when you finish.

applications/plugins/hid_app/views/hid_mouse_jiggler.c
MrLemur: Thanks for the comment! As I am trying to use the existing HID app, I have not seen any examples where a function can run continously while not locking the main thread and allowing input callbacks.

Is there something like a event loop that can be used in addition to the main thread, or would it be better to move this to its own app and logic?

skotopes: You can use furi_timer and do the work in timer callback

MrLemur: Ah fantastic! Thank you!

MrLemur:
> Please un-draft when you finish.

Changed to use furi_timer

skotopes (approved):

MrLemur:
@skotopes Thank you for the refactor! This is my first time using C coming from Golang, so great to see how to effectively use the Furi API.

PR #2120, opened 2022-12-11
Weather Station: Fix display of temps lower than -9

xMasterX:
# What's new

- Added check to move element a bit right if temp is lower than -9 (in result we have 3 characters instead of normal two)

### Before:
Screenshot 2022-12-11 at 03 12 54

### After:
Screenshot 2022-12-11 at 03 10 20

# Verification 

- Compile plugin, transmit signal of sensor that has temp lower than -9, verify that (-) symbol fits now

# Checklist (For Reviewer)


applications/plugins/weather_station/views/weather_station_receiver_info.c
```diff
@@ -82,8 +82,14 @@ void ws_view_receiver_info_draw(Canvas* canvas, WSReceiverInfoModel* model) {
     if(model->generic->temp != WS_NO_TEMPERATURE) {
         canvas_draw_icon(canvas, 6, 43, &I_Therm_7x16);
         snprintf(buffer, sizeof(buffer), "%3.1f C", (double)model->generic->temp);
-        canvas_draw_str_aligned(canvas, 47, 47, AlignRight, AlignTop, buffer);
-        canvas_draw_circle(canvas, 38, 46, 1);
+        uint8_t temp_x1 = 47;
+        uint8_t temp_x2 = 38;
+        if(model->generic->temp < -9.0) {
```
skotopes, line 87: Looks like AlignCenter will be better solution

xMasterX: I not tried it here, i think it will work ok for text but `canvas_draw_circle` has no ability to use align
Currently I don’t found better solution for `canvas_draw_circle`
I can use it for text but what to do with circle ?

skotopes: True, just realized that small ° is drawn not by text symbol.

skotopes (approved):

PR #2126, opened 2022-12-13
Frequency analyzer: allow setting rssi threshold

iRonJ:
# What's new

On the frequency analyzer, you can set the RSSI threshold that the UI will show a detected frequency.
image

# Verification 

Launch the Sub-Ghz frequency analyzer, and press the up/down buttons to change the threshold. Lowering the threshold will allow signals that are weaker to be shown as detected, and logged if they meet the appropriate criteria for logging (unchanged).

# Checklist (For Reviewer)

### DIFF
--- a/applications/main/subghz/helpers/subghz_frequency_analyzer_worker.c
+++ b/applications/main/subghz/helpers/subghz_frequency_analyzer_worker.c
@@ -22,6 +22,7 @@ struct SubGhzFrequencyAnalyzerWorker {
 
     volatile bool worker_running;
     uint8_t sample_hold_counter;
+    float rssi_min_thresh;
     FrequencyRSSI frequency_rssi_buf;
     SubGhzSetting* setting;
 
@@ -154,7 +155,7 @@ static int32_t subghz_frequency_analyzer_worker_thread(void* context) {
             (double)rssi_min);
 
         // Second stage: fine scan
-        if(frequency_rssi.rssi_coarse > SUBGHZ_FREQUENCY_ANALYZER_THRESHOLD) {
+        if(frequency_rssi.rssi_coarse > instance->rssi_min_thresh) {
             furi_hal_subghz_idle();
             subghz_frequency_analyzer_worker_load_registers(subghz_preset_ook_58khz);
             //for example -0.3 ... 433.92 ... +0.3 step 20KHz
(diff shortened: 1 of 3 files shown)


Skorpionm:
sorry, but what is it for?

iRonJ:
> sorry, but what is it for?

If you want to only see signals that are very strong, or if you want to see what weaker signals are around, you can adjust the level the frequency analyzer triggers at.

Skorpionm:
no need to unscrew rssi below the set threshold in the official firmware.  and everything will work as it should

skotopes:
@iRonJ I think what @Skorpionm is trying to say is "is it working? Is it working in meaningful way?"

One important things is: antenna is being disconnected when you use Frequency Analyzer, so there is not so much of the signal. Lowering trigger RSSI doesn't make much sense because it comes too close to the noise floor.

If you want to catch weak signal you need to connect antenna first. But if you'll do that you'll start to receive a lot of different signals from around you and distinguishing your signal from the others will be quite complex.

Implementing it right will require some work:
- Use antenna in FA
- Dynamically calculate and subtract environment spectrum
- Tune chip RF params 
- And a lot of experiments to fine tune it

iRonJ:
> @iRonJ I think what @Skorpionm is trying to say is "is it working? Is it working in meaningful way?"
> 
> One important things is: antenna is being disconnected when you use Frequency Analyzer, so there is not so much of the signal. Lowering trigger RSSI doesn't make much sense because it comes too close to the noise floor.
> 
> If you want to catch weak signal you need to connect antenna first. But if you'll do that you'll start to receive a lot of different signals from around you and distinguishing your signal from the others will be quite complex.
> 
> Implementing it right will require some work:
> 
> * Use antenna in FA
> * Dynamically calculate and subtract environment spectrum
> * Tune chip RF params
> * And a lot of experiments to fine tune it

That is good information I didn’t know. My use case was trying to determine if a zigbee radio was operating subghz. I wasn’t getting any triggering with stock fw. Lowering the threshold let me confirm the zigbee radio was likely operating at the higher carrier and that it wasn’t some other error in the detection.

iRonJ:
> > @iRonJ I think what @Skorpionm is trying to say is "is it working? Is it working in meaningful way?"
> > One important things is: antenna is being disconnected when you use Frequency Analyzer, so there is not so much of the signal. Lowering trigger RSSI doesn't make much sense because it comes too close to the noise floor.
> > If you want to catch weak signal you need to connect antenna first. But if you'll do that you'll start to receive a lot of different signals from around you and distinguishing your signal from the others will be quite complex.
> > Implementing it right will require some work:
> > 
> > * Use antenna in FA
> > * Dynamically calculate and subtract environment spectrum
> > * Tune chip RF params
> > * And a lot of experiments to fine tune it
> 
> That is good information I didn’t know. My use case was trying to determine if a zigbee radio was operating subghz. I wasn’t getting any triggering with stock fw. Lowering the threshold let me confirm the zigbee radio was likely operating at the higher carrier and that it wasn’t some other error in the detection.

What if the UI wasn’t visible by default, and then only if the user pressed up or down the value showed up, as an additional diagnostic?

PR #2142, opened 2022-12-16
[FL-2811] Fix PVS-Studio warnings

gsurkov:
# What's new

- Fix warnings generated by PVS-Studio by fixing the problems and suppressing the false positives either globally via the `.pvsconfig` file or via special comments in the code.
- Refactor some code (without changing its functionality).
- Add `float_is_equal` function within the `float_tools` library for safe equality test for float values.

# Verification 

1. Prepare the firmware:
```
./fbt COMPACT=1 version_json proto_ver icons firmware_cdb dolphin_internal dolphin_blocking _fap_icons api_syms
```
2. Run pvs-studio: `./run-pvs.sh` (see attached file).
3. Open `pvs-reports/index.html`. No warnings should be present.
4. Run unit tests?
5. Use the firmware and hope that it doesn't break.

# PVS-Studio launch script
```sh
#!/bin/bash

set -e

SRC_DIR='flipperzero-firmware'
REPORTS_DIR='pvs-reports'

if [ -d "$REPORTS_DIR" ]; then rm -rf "$REPORTS_DIR"; fi

cd "$SRC_DIR"

. scripts/toolchain/fbtenv.sh
pvs-studio-analyzer analyze @.pvsoptions -C gccarm -j $(nproc) -f build/f7-firmware-DC/compile_commands.json  --ignore-ccache
plog-converter -a GA:1,2,3 -t fullhtml PVS-Studio.log -o "../$REPORTS_DIR"

cd ..
```

# Checklist (For Reviewer)


lib/subghz/protocols/bett.c
```diff
@@ -242,7 +242,6 @@ void subghz_protocol_decoder_bett_feed(void* context, bool level, uint32_t durat
         if(!level) {
             if(DURATION_DIFF(duration, subghz_protocol_bett_const.te_short * 44) <
                (subghz_protocol_bett_const.te_delta * 15)) {
-                instance->decoder.parser_step = BETTDecoderStepSaveDuration;
```
Skorpionm, line 245: bring it back. this is magic, you need the decoder to continue decoding the stream without looking for a new guard time

lib/subghz/protocols/secplus_v2.c
```diff
@@ -151,7 +151,7 @@ static bool subghz_protocol_secplus_v2_mix_order_decode(uint8_t order, uint16_t
     case 0x06: // 0b0110  2, 1, 0],
     case 0x09: // 0b1001  2, 1, 0],
         p[2] = a;
-        p[1] = b;
+        // p[1]: no change
```
Skorpionm, line 154: here you can just comment out the lines so that PVS does not swear, otherwise the readability of the code suffers

lib/subghz/protocols/bett.c
gsurkov: :magic_wand:

lib/subghz/protocols/secplus_v2.c
gsurkov: I believe that my version is more clear because a valid C code that was commented out is confusing (why was it commented out? did someone forget to uncomment it? is this code a work in progress? etc)

scripts/fbt_tools/fbt_debugopts.py
```diff
@@ -47,6 +47,8 @@ def generate(env, **kw):
             "target extended-remote ${GDBREMOTE}",
             "-ex",
             "set confirm off",
```
hedger: move to gdbinit?

hedger: also pagination a few lines before

hedger (dismissed):

furi/core/event_flag.c
```diff
@@ -9,7 +9,11 @@
 
 FuriEventFlag* furi_event_flag_alloc() {
     furi_assert(!FURI_IS_IRQ_MODE());
-    return ((FuriEventFlag*)xEventGroupCreate());
+
+    EventGroupHandle_t handle = xEventGroupCreate();
+    furi_check(handle);
```
DrZlo13, line 14: furi_check doesn't make sense here, xEventGroupCreate will only return 0 in case of OOM, which is not possible with our allocator

lib/nfc/nfc_worker.c
```diff
@@ -454,10 +452,7 @@ void nfc_worker_read_type(NfcWorker* nfc_worker) {
                     break;
                 }
             } else {
-                if(!card_not_detected_notified) {
-                    nfc_worker->callback(NfcWorkerEventNoCardDetected, nfc_worker->context);
-                    card_not_detected_notified = true;
-                }
+                nfc_worker->callback(NfcWorkerEventNoCardDetected, nfc_worker->context);
             }
```
gornekich, line 461: @gsurkov why do we need this change?
The idea was to send WorkerEventNoCardDetected event only once after card was lost. In this code we will always send WorkerEventNoCardDetected event if no card is detected

lib/nfc/parsers/all_in_one.c
```diff
@@ -73,24 +73,14 @@ bool all_in_one_parser_parse(NfcDeviceData* dev_data) {
...
-        // I hate this with a burning passion.
-
-        // The number starts at the second half of the third byte on page 4, and is 32 bits long
-        // So we get the second half of the third byte, then bytes 4-6, and then the first half of the 7th byte
-        // B8 17 A2 A4 BD becomes 81 7A 2A 4B
-        serial = (dev_data->mf_ul_data.data[4 * 4 + 2] & 0x0F) << 28 |
-                 dev_data->mf_ul_data.data[4 * 4 + 3] << 20 |
-                 dev_data->mf_ul_data.data[4 * 4 + 4] << 12 |
-                 dev_data->mf_ul_data.data[4 * 4 + 5] << 4 |
-                 (dev_data->mf_ul_data.data[4 * 4 + 6] >> 4);
```
gornekich, line 93: @Astrrra could you confirm that this code is unused

gornekich: Looks like we forget to display serial number

lib/nfc/protocols/mifare_classic.c
```diff
@@ -924,7 +914,7 @@ bool mf_classic_emulator(MfClassicEmulator* emulator, FuriHalNfcTxRxContext* tx_
                     memcpy(block_data, plain_data, MF_CLASSIC_BLOCK_SIZE);
                 }
             }
-            if(memcmp(block_data, emulator->data.block[block].value, MF_CLASSIC_BLOCK_SIZE)) {
+            if(memcmp(block_data, emulator->data.block[block].value, MF_CLASSIC_BLOCK_SIZE) == 0) {
```
gornekich: Guess there must be != 0

gornekich (changes requested):

lib/nfc/parsers/all_in_one.c
Astrrra: Yep, it's a duplicate of what happens on line 99 and onwards, not sure why it's there, but it's definitely unnecessary

DrZlo13 (dismissed):

lib/nfc/protocols/mifare_classic.c
gsurkov: Yep, absolutely. My bad.

furi/core/event_flag.c
DrZlo13: minor issue

lib/nfc/nfc_worker.c
gsurkov: The `card_not_detected_notified` variable is reset on every iteration and therefore does nothing. Perhaps there initially was a logical error.

gornekich: Corrected the logic, thanks

gornekich (dismissed):

furi/core/event_flag.c
skotopes: that is incorrect statement, xEventGroupCreate  uses pvPortMalloc which can return null

skotopes: I think we can leave it here for compatibility with stock allocator.

skotopes (approved):

Issue #1901, opened 2022-10-21
Scrolling text when longer than screen


krisss117:
### Description of the feature you're suggesting.

Please add text scrolling when text is longer than screen. I think only selected position will scroll text. Then it allow to read longer text for example NFC, SUBGHZ.

It will be very usefull.

skotopes:
We'll come back to this task after v1.0

ahumeniy:
Something like this? 
https://drive.google.com/file/d/12yFBldnQ5us1lq8ejsAJJNFWMUM8a1n0/view?usp=sharing

skotopes:
@ahumeniy yep, looks like you've nailed it.

ahumeniy:
Should I submit a PR?

skotopes:
@ahumeniy feel free to do that ;-)

skotopes:
fixed

PR #2087, opened 2022-12-04
WS: Show received signal age

xMasterX:
Original feature by @LY2NEO with some fixes from me - [(read details)](https://github.com/DarkFlippers/unleashed-firmware/pull/197#issuecomment-1336348811)

Now rewritten to apply recommended UI changes, also contains extra change similar to #2111

# What's new

- When data is received from weather station it saves datastamp and calculates difference between actual time and data received time.
- Shows time counter on signal screen, when we reach above 1min it blinks with text [OLD], when we reach 59+ mins, it just shows [OLD]
- Now its possible to understand when signal was received if data on screen isn’t changed (for example temp stays the same)

# Verification 

- Launch Weather Station plugin, receive any supported temp sensor signal
- Select received signal and open it, check counter
- Send same signal again and see counter reset to 0

[Example_signal.zip](https://github.com/flipperdevices/flipperzero-firmware/files/10148576/Example_signal.zip)

# Checklist (For Reviewer)

- [x] PR has description of feature/bug or link to Confluence/Jira task
- [x] Description contains actions to verify feature/bugfix
- [x] I've built this code, uploaded it to the device and verified feature/bugfix

PR #2144, opened 2022-12-17
File browser scroll large names

ahumeniy:
# What's new
- Implements scrolling text on File Browser when the file name is longer than allowed on-screen as requested in #1901 

# Verification 

- Open any app which uses the file browser.
- Search for a file which has a name longer than allowed on screen. Look for the three dots (...) at the right.
- The file name should start scrolling after 1 second.
- Only the selected file should be scrolling.

# Checklist (For Reviewer)

### DIFF
--- a/applications/services/gui/modules/file_browser.c
+++ b/applications/services/gui/modules/file_browser.c
@@ -476,21 +476,29 @@ static void browser_draw_list(Canvas* canvas, FileBrowserModel* model) {
         bool is_long_filename = canvas_string_width(canvas, furi_string_get_cstr(filename)) >
                                 frame_width;
 
-        if(is_long_filename && model->item_idx == idx &&
-           current_frame_tick - model->animation_begin_frame > 1000) {
+        if(is_long_filename && model->item_idx == idx) {
+            int elapsed_time = (current_frame_tick - model->animation_begin_frame);
+
             // Begin animating the file name 1 second after selection...
-            size_t name_length = furi_string_size(filename);
-            uint8_t offset = ((int)furi_get_tick() / 250) % (int)name_length;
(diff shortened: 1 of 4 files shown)


applications/services/gui/modules/file_browser.c
```diff
@@ -534,6 +566,7 @@ static bool file_browser_view_input_callback(InputEvent* event, void* context) {
                     if(event->key == InputKeyUp) {
                         model->item_idx =
                             ((model->item_idx - 1) + model->item_cnt) % model->item_cnt;
+                        model->animation_begin_frame = furi_get_tick();
```
skotopes: couple notes:

- View/ViewPort doesn't refresh itself. You need to send redraw request
- This code only works because of the other processes requesting redraw
- It make sense to add FuriTimer that will increment counter in the model

skotopes: Also there will be cases when file browser starts with long filename, looks like this code doesn't handles it.

skotopes (changes requested):

skotopes:
Some examples with FuriTimer and view model:
- https://github.com/flipperdevices/flipperzero-firmware/pull/2113
- https://github.com/flipperdevices/flipperzero-firmware/pull/2087 

Please un-draft when ready.

ahumeniy:
Made the requested changes. Because it makes things more complex, I dropped the 1 second delay to start scrolling thing.

skotopes:
Code doesn't properly handle navigation and default positioning on file with long name.
I'm going to close this PR and implement this feature in a different way.

skotopes:
Implemented in #2159

PR #2146, opened 2022-12-19
fbt: add `fap_deploy` target to build and copy all .fap apps to flipper

smkent:
# What's new

This adds a new `./fbt` target called `fap_deploy`. This builds and uploads all discovered FAPs to the connected Flipper via USB. Similar to how `./fbt launch_app` builds and uploads a single FAP over USB, `fap_deploy` builds and copies all FAPs (no apps are launched).

I wrote this for easier (re)building and uploading of all FAPs when the Flipper firmware is updated and the ABI version changes.

# Verification 

* Remove all existing FAP build artifacts: `rm -rf build/f7-firmware-D/.extapps/`
* Connect Flipper via USB
* Run `./fbt fap_deploy`
* Observe FAPs are built and copied to the connected Flipper

# Checklist (For Reviewer)


scripts/runfap.py
```diff
@@ -15,6 +15,13 @@
 class Main(App):
     def init(self):
         self.parser.add_argument("-p", "--port", help="CDC Port", default="auto")
+        self.parser.add_argument(
+            "-l",
```
hedger: Better change that to a flag meaning "do upload only". Since your approach changes the script's behavior  - "runfap.py" by default no longer runs the FAP itself.

site_scons/extapps.scons
```diff
@@ -109,12 +110,25 @@ if appsrc := appenv.subst("$APPSRC"):
...
     appenv.Alias("launch_app", app_artifacts.validator)
 
+appenv.PhonyTarget(
+    "faps_copy",
+    os.linesep.join(
+        [
+            '${PYTHON3} "${APP_RUN_SCRIPT}"'
+            f' "{app_artifact.compact[0].rstr()}"'
+            f' --fap_dst_dir "/ext/apps/{app_artifact.app.fap_category}"'
+            for app_artifact in appenv["EXT_APPS"].values()
```
hedger: That's not how you create targets in scons.
Let the build system construct the commandline and just provide a list of source nodes in `source=` parameter.

hedger (changes requested):

site_scons/extapps.scons
smkent: I'm happy to get some help here as I wasn't sure what the best way is to run each copy in series. While experimenting, I did have a target with each detected application as an individual build dependency. However, scons ran them in parallel resulting in multiple script invocations attempting to connect to the Flipper via USB simultaneously. Is there a way to provide scons with a list of targets, but have them run one at a time instead of in parallel?

scripts/runfap.py
smkent: Done, I swapped out `-l`/`--launch` with `-n`/`--no-launch`.

site_scons/extapps.scons
hedger: I think I'll look into that myself, probably adding a multi-file mode for runfap.py.

hedger (approved):

smkent:
🎉

PR #2204, opened 2022-12-27
[FL-2435] SD over SPI improvements

DrZlo13:
# What's new

- SPI-DMA
- Rewritten SD-SPI HAL
- Significant SD card speed improvements
- Storage: code optimizations

# Verification 

- All things related to working with the SD card should continue to work, including updating the firmware via usb

# Checklist (For Reviewer)


firmware/targets/f7/fatfs/sd_spi_io.c
```diff
@@ -813,7 +807,7 @@ SdSpiStatus sd_get_card_info(SD_CardInfo* card_info) {
         return status;
     }
 
-    status = sd_spi_get_cid(&(card_info->Cid));
+    status = sd_get_cid(&(card_info->Cid));
```
hramrach, line 811: Hello,

this is likely going to crash.

sd_init  https://github.com/flipperdevices/flipperzero-firmware/pull/2204/files#diff-882800445634166f45106f6d29227981ec54b8102f708886e13e54f7b8c245c6R741 which is also randomly called outside of the storage code initializes the spi handle which sd_get_cid does not, and then there is no guarantee the handle will be initialized. The simple solution is to not rename but rather add a wrapper that does the initialization.

The other option is got get the CID from storage_sd_info which is called anyway. That's API change which I did not want for the minimal implementation but when merging to this rewrite it would make more sense.

DrZlo13: Yep, that's correct.

I have moved the CID-related data into SDInfo struct, and prepared it on the storage service side to make it easier to use.

hramrach: Looks good, thanks

skotopes (approved):

PR #2217, opened 2022-12-29
Gui: change data ownership model in submenu, own text by default (#2217)

nullableVoidPtr:
# What's new

- Rework the Submenu GUI module to use managed `FuriString`s instead.
- Refactor DESFire info menus to use the new submenu API instead of using `nfc->text_store` as a buffer.

# Verification 

- Submenus (in applications, plugins) appear to work as normal.
- DESFire dynamic submenus were tested using samples from documentation, extended to include more files and apps.  

# Checklist (For Reviewer)

### DIFF
--- a/applications/debug/bt_debug_app/bt_debug_app.c
+++ b/applications/debug/bt_debug_app/bt_debug_app.c
@@ -44,13 +44,13 @@ BtDebugApp* bt_debug_app_alloc() {
 
     // Views
     app->submenu = submenu_alloc();
-    submenu_add_item_cstr(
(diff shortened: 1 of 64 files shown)


DrZlo13:
@nullableVoidPtr Thanks for your contribution.
The part where the submenu uses the managed container is nice, but we don't think the submenu needs a separate FuriString* function because we can always get the cstring from the FuriString.

I took the liberty of rewriting this piece of code, I hope you don't mind.

skotopes (approved):

Issue #2185, opened 2022-12-23
Picopass text overflow on failed read


doomwastaken:
### Describe the bug.

![image](https://user-images.githubusercontent.com/72250702/209342813-7dd1aded-7aa7-449b-bc7e-203772ffe067.png)

just text overflow

### Reproduction

Go into picopass and try to read any card, pull the card away in process to get an error message. 
Works on 0.73.1 and 0.74.0-rc / 0.74.1-rc (haven't tested on previous versions)

### Target

f7

skotopes:
@bettse @pcunning Hi guys, can you take a look? )

bettse:
is the font mono-spaced?  if so, I'm surprised we don't see it for all values.  I haven't had time to try to reproduce, but I suspect we just need to bump down the font size a couple points.

bettse:
I couldn't reproduce, but I saw that "1" is super narrow, which can pretty much explain it.

skotopes:
Please check dev or 0.77 and reopen if issue persist

PR #2232, opened 2022-12-31
Move CSN space to prevent overflow

bettse:
# What's new

- Fix for https://github.com/flipperdevices/flipperzero-firmware/issues/2185
- Removes space between colon and first CSN byte.  Should allow enough room to fit a full 00's CSN.

# Verification 

- You could try to reproduce the bug and see it is fixed
- Alternatively just see the space has been moved and estimate that it'll correct the problem from the bug

# Checklist (For Reviewer)


bettse:
cc @pcunning

applications/plugins/picopass/scenes/picopass_scene_read_card_success.c
```diff
@@ -33,7 +33,7 @@ void picopass_scene_read_card_success_on_enter(void* context) {
     uint8_t csn[PICOPASS_BLOCK_LEN];
     memcpy(csn, &AA1->data[PICOPASS_CSN_BLOCK_INDEX], PICOPASS_BLOCK_LEN);
     for(uint8_t i = 0; i < PICOPASS_BLOCK_LEN; i++) {
-        furi_string_cat_printf(csn_str, " %02X", csn[i]);
+        furi_string_cat_printf(csn_str, "%02X ", csn[i]);
```
skotopes, line 36: Nope, not going to fix issue ;-) Space is still space and it's handled when rendering. But even if you trim last space there will be not enough space to fix the line.

skotopes (changes requested):

skotopes:
Is it ok if we remove spaces between?

bettse:
Sorry for the delay, the idea was that the space _sometimes_ was too long, but since it doesn't wrap (as in the failure example) it would be offscreen whitespace.  Inelegant.  I'm 100% in favor or just removing the spaces.  I'll try to get around to that, but feel free to do it and close this out if I'm too slow.

skotopes (approved):

PR #2233, opened 2022-12-31
feat: add missing `const` qualifiers

JarvisCraft:
# What's new

Missing `const` qualifiers were added to the existing APIs (in most cases, getters).

# Verification 

- Recompile old applications using these APIs.

# Checklist (For Reviewer)


skotopes:
Why?

JarvisCraft:
> Why?

What exactly?

skotopes:
> > Why?
> 
> What exactly?

I do agree that const is missing in SPI. But canvas and gui are kinda complex composite entities, why them?

JarvisCraft:
> I do agree that const is missing in SPI. But canvas and gui are kinda complex composite entities, why them?

TL;DR this is the result of [this](https://t.me/flipperzero/103925/144268) thread with @DrZlo13.

In general, the point is to have `const *` whenever it is logical: most commonly, such things as trivial getters, but, generally, all functions which don't mutate pointed data (especially, method-like functions).

While the benefit of this may not be so obvious for pure-C projects (generally, due to the reasons that API users also ignore `const` usually thus having everything be plain `*`), it becomes more important when working with the API from other languages, i.e. creating safe bindings for it. Having stricter contracts here allows foreign callers to:

- better understand semantics of the functions;

- recognize basic invariants upheld by the functions (such as not mutating the internal state means that the caller may rely on the object being kept in the same state);

- recognize logical errors.

My personal example where this appeared is me [writing](https://github.com/flipperzero-rs/flipperzero/pull/290) safe Rust API for these APIs.

Another example is that while analyzing these sources for missing `const`s I found out that `canvas_get_font_params` returned non-`const` pointer to a logically constant data (which I initially treated as mutable in my bindings).

These changes may look non-significant by themselves. but altogether explicit qualifiers should improve overall API usage experience.

skotopes (approved):

PR #2249, opened 2023-01-05
Show region information in sub-GHz app

ntfshard:
# What's new

Information of available for transmission bands could be used as a reference for different purposes, like ignoring forbidden ranges while exploring something (in other words, helps to avoid frustration while you trying to transmit something just captured and got the error msg)

image

# Verification 

- Sub-GHz -> (New menu item) Region Information

# Checklist (For Reviewer)


ntfshard:
Two comments from my side: perhaps I could make it more beautiful, but I have no experience with your UI library. Some guidance/feedback could be very helpful.
And does it make sense add restriction information for capturing functions(read/raw) to show that this range can be only for RX. Like when you selecting the capture settings, it could be something like _NoTx_ near the frequency.

skotopes (dismissed):

skotopes (approved):

Issue #2251, opened 2023-01-05
Use InputKeyRight same way as InputKeyOk in file_browser


zverev-iv:
I think ```InputKeyRight``` event must works symmetrically to ```InputKeyBack``` in file browser.
so we can bind it to same code as ```InputKeyOk```

```diff --git a/applications/services/gui/modules/file_browser.c b/applications/services/gui/modules/file_browser.c
index e03a032c..a4c50568 100644
--- a/applications/services/gui/modules/file_browser.c
+++ b/applications/services/gui/modules/file_browser.c
@@ -617,7 +617,7 @@ static bool file_browser_view_input_callback(InputEvent* event, void* context) {
             browser_update_offset(browser);
             consumed = true;
         }
-    } else if(event->key == InputKeyOk) {
+    } else if(event->key == InputKeyOk || event->key == InputKeyRight) {
         if(event->type == InputTypeShort) {
             BrowserItem_t* selected_item = NULL;
```

1. I think this will help the intuitiveness of the file_browser. 
2. right key is the only one not bound in file_browser.

Is it worth creating a PR?

skotopes:
From UI/UX standpoint we'd like keep patterns simple: always one way to do something. So those changes are not going to be accepted.

PR #2264, opened 2023-01-06
Use InputKeyRight same way as InputKeyOk

zverev-iv:
this must improve UX of the file_browser (#2251):

1. file browser will use all keys events
2. left and right click now works symmetrically

# What's new

press right button will work same way like center button in file browser.

# Verification 

1. open Applications menu on flipper
3. select folder (up/down)
4. press right. -> folder opens.
5. press left for parent folder
6. repeat right click on same folder or any another
7. now right click on application. This will launch the application

# Checklist (For Reviewer)

### DIFF
--- a/applications/services/gui/modules/file_browser.c
+++ b/applications/services/gui/modules/file_browser.c
@@ -617,7 +617,7 @@ static bool file_browser_view_input_callback(InputEvent* event, void* context) {
             browser_update_offset(browser);
             consumed = true;
         }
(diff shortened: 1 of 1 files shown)


skotopes:
There is one more place that uses same UX pattern: archive.
Also @Karator will check UI/UX and we'll come back with review.

skotopes:
Detailed explanation in #2251

PR #2289, opened 2023-01-10
Add lockable capability to Submenu module

giacomoferretti:
# What's new

- Submenu items can now be locked. When clicked, they will display a help message.

# Verification 

- Read new code (or write a simple app that uses this new function)

# Screenshots

lockable1
lock3
lockable2

# Checklist (For Reviewer)

### DIFF
--- a/applications/services/gui/modules/submenu.c
+++ b/applications/services/gui/modules/submenu.c
@@ -1,11 +1,13 @@
 #include "submenu.h"
 
+#include <assets_icons.h>
 #include <m-array.h>
 #include <gui/elements.h>
 #include <furi.h>
 
 struct Submenu {
     View* view;
+    FuriTimer* timer;
 };
 
 typedef struct {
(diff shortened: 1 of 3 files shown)


skotopes:
Use cases are?

giacomoferretti:
One use case might be to show the user the ability to perform a certain action, but not at that time because prerequisites are required (e.g., full card reading).

For example here: [nfc_scene_mf_classic_menu.c#L25](https://github.com/flipperdevices/flipperzero-firmware/blob/0.75.0/applications/main/nfc/scenes/nfc_scene_mf_classic_menu.c#L25)

The "Detect Reader" option may always be visible to inform the user that there is an opportunity to run it, but not at this time because the card has not been read.

Could be useful in external apps too.

skotopes:
Sounds like too over specialized feature.
That is kinda contradicts with idea of small and simple toolkit.

giacomoferretti:
No problem!

PR #2296, opened 2023-01-11
nfc: Fix sector reads when one block is unreadable for MIFARE Classic

RouNNdeL:
# What's new

When a block (or multiple blocks) in a sector are unreadable (access bits `111`) every block after it will fail to read, since auth is performed only once per sector. This PR changes it to auth per block, just like the proxmark does it.

# Verification 

- Get or emulate a card with unreadable block in a sector.
- Attempt to read it using the default 'Read' action.
- All blocks after the unreadable block used to not read - `??` in the dump.

# Checklist (For Reviewer)


RouNNdeL:
I think that, write back to card and update from card need a change as well, since they flatten the block read mask into a single bool for the entire sector. 

https://github.com/flipperdevices/flipperzero-firmware/blob/073fb3861ac310c07bdbbb3fd12fc37fe8d9799c/lib/nfc/nfc_worker.c#L817-L823

If the saved data has block 18 of sector 4 unread, but when writing back we cannot read block 17, since it's in the same sector the check will still pass, while I belive it should fail. I presume this was written to assume if a block in a sector was not read, the entire sector wasn't, which is not always the case.

gornekich:
Hello @RouNNdeL 
Thanks for PR!

I tested it on cards with read-lock blocks and it works just fine. However I was unhappy to see that reading supported cards, previously saved cards and cards with default keys now takes much more time.

Since halting and passing authentication takes some time, I think it's better to do this only if we failed to read previous block. I attach my patch to your code, which optimize re-auth procedure. Let me know your thoughts on it.
[optimization.txt](https://github.com/flipperdevices/flipperzero-firmware/files/10493075/optimization.txt)

RouNNdeL:
I also thought that this was quite unoptimized in the current state. I like your suggestion, but I think I'd like to go further. The approach I have in mind. First abstract away the actual sector read from `mf_classic_read_sector_with_reader` and `mf_classic_read_sector` with the following signature: `(FuriHalNfcTxRxContext * tx_rx, uint8_t sector_num, MfClassicKey key_type, uint64_t key)`. The function would:
- attempt auth with the provided key
- read sector trailer with AC
- read blocks 0-2 **if** the AC allow read of those

Maybe we could also retry a read if it was unsuccessful (up to 3 times)? This approach would be even faster than yours for cards that have the dead block, since we won't attempt reads on those at all. What do you think?

gornekich:
@RouNNdeL I agree that your solution would be the best.

I suggest 2 options for further steps:

1) Since the code that you suggested + my optimization already works better than current dev, we can merge it. Then you or me can implement the approach you suggested.

2) If you think that implementation of suggested approach will not take long, you could proceed work in this PR.

Personally, I prefer an iterative approach.

RouNNdeL:
Sure, we can marge it with your patch, can you apply it or do I need to?

gornekich:
I think I don't have access to your repo. Please, apply it yourself

gornekich (approved):
Thanks!

skotopes (approved):

PR #2299, opened 2023-01-14
LF-RFID: add CRC calculation to paradox protocol

jerji:
The algorithm for the checksum was discovered earlier today on the Proxmark discord by @mwalker33.

I also removed the extra data line as it is already displayed at the top.

# What's new

- New CRC field. Both the one on the card and the calculated one.
- Check card validity in brief view.

# Verification 

- Load up a paradox card ex: (00 04 54 30 7B 70)
- Look at the new field.
- If you change the card, CRC won't match.
- You can write it to a card and see the invalid warning.

# Checklist (For Reviewer)


jerji:
Here is the Proxmark3 version.
https://github.com/RfidResearchGroup/proxmark3/commit/224a83527f6122e758abfc6acec004eaf495b37c

jerji:
Is there something I need to do before this gets reviewed?

skotopes (approved):

PR #2307, opened 2023-01-15
NFC: add MIFARE MINI support

giacomoferretti:
# What's new

- Added support for MIFARE MINI ([datasheet](http://www.orangetags.com/wp-content/downloads/datasheet/NXP/MF1ICS20.pdf))

# Verification 

- Tested on some pre-issued consumer cards.

# Checklist (For Reviewer)


lib/nfc/protocols/mifare_classic.h
```diff
@@ -18,6 +19,7 @@
 #define MF_CLASSIC_ACCESS_BYTES_SIZE (4)
 
 typedef enum {
+    MfClassicTypeMini,
```
hedger: This PR does not include a major API bump and does not require rebuilding .faps -- but public API methods accepting this enum will break. Since the binaries were built with an older version of this enum, with same members having different values. So, please either add the new entry to the end of the enum, or force the API version change.

hedger (changes requested):
Handle the new enum member appropriatlely.

lib/nfc/protocols/mifare_classic.h
giacomoferretti: Oh, you are right. Fixing it! :+1:

giacomoferretti:
Should be ok now.

hedger (dismissed):
LGTM
@gornekich have a look please

applications/debug/unit_tests/nfc/nfc_test.c
```diff
@@ -466,6 +466,10 @@ static void mf_classic_generator_test(uint8_t uid_len, MfClassicType type) {
     nfc_device_free(nfc_keys);
 }
 
+MU_TEST(mf_mini_file_test) {
```
gornekich, line 469: Add 
MU_RUN_TEST(mf_mini_file_test);
in MU_TEST_SUITE(nfc) please

gornekich (changes requested):
Thanks for your PR!
Tested it and it works fine. Please fix unit tests and we will merge it

applications/debug/unit_tests/nfc/nfc_test.c
giacomoferretti: Missed that, thanks!

gornekich (approved):
Thank you!

skotopes (approved):

PR #2317, opened 2023-01-18
nfc: Add mifare classic value block commands

RouNNdeL:
# What's new

Implements the missing mifare classic commands in the emulator: 
- increment
- decrement
- restore
- transfer

Implements writing to initial card that uses value blocks

# Verification 

Emulation verified with a proxmark for all commands and an external reader (only increment + transfer).
Modifying value blocks tested on a physical card and proxmark emulator

# Checklist (For Reviewer)


Astrrra:
Looks good! Will test sometime next week

RouNNdeL:
Please note that the debug log formerly used to get nonces for mfkey32 was removed, since any logging during emulation will cause the readers to timeout. This meant using debug logs for anything surrounding the emulation couldn't not be utilized.

RouNNdeL:
Ideally we would implement a trace buffer that logs everything until the emulation is finished and only then prints the logs to the console.

Karl99999:
Will it be possible to write the nfc code copies from a door key tag to a new nfc key tag with this program ? Now you can only read and emulate with the flipper zero in the original modus ?

augustozanellato:
I'm not a flipper team member so please take my words with a grain of salt: it would be nice if the value block commands were implemented like they are in libfreefare so you have the ability to transfer&restore to different sectors for example. The `mf_classic_inc_dec_res_block` would still be nice to have because it simplifies value block interaction but the separated functions (so `mf_classic_inc_block`, `mf_classic_dec_block`, `mf_classic_transfer_block`, `mf_classic_restore_block`) would be useful for low level interaction with the tag

Karl99999:
The flipper zero copy the nfc mifare classic 1K tags perfect and with using emulation , it's the same as you use the original key

It's copying only the key A and not key B ( the key B on the sector 0 is only written ???????????, and read 16 of 16 blocks and copy only 31 off 32 keys on the tag , but it's impossible to write on a new tag that you just copied from the original tag onto a new tag  that is readable/writable/rewritable, or do you need to format the new nfc mifare classic 1K tags before you write on them with flipper zero ? Or why is the a write function on flipper zero not activated? Where can I buy the full software version too flipper zero that takes away all the blocking in every frequencies and blocking the writing ability in every different signal frequencies? And take away the error message that this is not legal to copy in your region, a regional blocking device that is built inside flipper zero, that you can normally use cyber ghost to change location and to use VPN server to take away or change the regional blocking inside the flipper zero? 
What is the price for the full software version for flipper zero, and where can I buy it 😊🙏💴

RouNNdeL:
> I'm not a flipper team member so please take my words with a grain of salt: it would be nice if the value block commands were implemented like they are in libfreefare so you have the ability to transfer&restore to different sectors for example. The `mf_classic_inc_dec_res_block` would still be nice to have because it simplifies value block interaction but the separated functions (so `mf_classic_inc_block`, `mf_classic_dec_block`, `mf_classic_transfer_block`, `mf_classic_restore_block`) would be useful for low level interaction with the tag

You cannot transfer a value from one sector to the other, since authorization clears the transfer buffer. But I agree that splitting inc/dec/restore and transfer into 2 separate functions does make sense. I don't think splitting the other 3 is necessary, since the only part that's different is the command sent. However, since after inc/dec/restore the only valid command is the transfer command, I think that adding a block number to transfer to as a separate argument would accomplish what you want?

augustozanellato:
> > I'm not a flipper team member so please take my words with a grain of salt: it would be nice if the value block commands were implemented like they are in libfreefare so you have the ability to transfer&restore to different sectors for example. The `mf_classic_inc_dec_res_block` would still be nice to have because it simplifies value block interaction but the separated functions (so `mf_classic_inc_block`, `mf_classic_dec_block`, `mf_classic_transfer_block`, `mf_classic_restore_block`) would be useful for low level interaction with the tag
> 
> You cannot transfer a value from one sector to the other, since authorization clears the transfer buffer. But I agree that splitting inc/dec/restore and transfer into 2 separate functions does make sense. I don't think splitting the other 3 is necessary, since the only part that's different is the command sent. However, since after inc/dec/restore the only valid command is the transfer command, I think that adding a block number to transfer to as a separate argument would accomplish what you want?

That’s not my experience. I tried in some classic 1k both “old” and ev1 (with verified signature so they were legit) and the transfer buffer isn’t cleared during auth. I can freely do the following:

- auth 10 with key a
- restore 10
- auth 20 with key a
- transfer 20

and the value contained in block 10 is transferred to block 20 (ADR bytes included).
If you think that should not be the case can you point me to some resources that say so? I discovered this possibility via some trial&error.

RouNNdeL:
> That’s not my experience. I tried in some classic 1k both “old” and ev1 (with verified signature so they were legit) and the transfer buffer isn’t cleared during auth. I can freely do the following:
> 
> * auth 10 with key a
> * restore 10
> * auth 20 with key a
> * transfer 20
> 
> and the value contained in block 10 is transferred to block 20 (ADR bytes included). If you think that should not be the case can you point me to some resources that say so? I discovered this possibility via some trial&error.

That's really interesting. First of all, there's the [NXP documentation](https://www.nxp.com/docs/en/data-sheet/MF1S50YYX_V1.pdf) for EV1 1K which contains this flow diagram:

image

There is also a [paper on Mifare Classic](https://cs.ru.nl/~wouter/papers/2008-thebest.pdf) which suggests there is an undocumented feature in some cards, that allows to write arbitrary value to a value block, but only in the same sector. Here's the modified diagram, note the exclamation marks: 

image

They also state that

> This can read the contents of another block in the **same sector** into the temporary buffer

and 

> This implies that is is impossible to move the contents of value blocks over the boundary of the individual sector: after authenticating for another sector, first an increment, decrement or restore is required before a transfer can be performed. This prevents the temporary register to serve as a means to move data from one sector to the other.

I have a card that's using 2 value blocks within a sector and the 3rd data block is in the *dead* state (that's how I discovered #2296 btw) just as the paper suggests. You are saying I can write an arbitrary value to that block by using a different sector? I'm also interested, are you performing nested authentication?

augustozanellato:
> I have a card that's using 2 value blocks within a sector and the 3rd data block is in the _dead_ state (that's how I discovered #2296 btw) just as the paper suggests. You are saying I can write an arbitrary value to that block by using a different sector? I'm also interested, are you performing nested authentication?

Not sure if you can overwrite the dead block but you can arbitrary change the value blocks, I'll give you a practical example:
This is a sector 3 dump from one of my tags:
```
F1 FF FF 7F:0E 00 00 80|F1 FF FF 7F:0D F2 0D F2
F1 FF FF 7F:0E 00 00 80|F1 FF FF 7F:0D F2 0D F2
00 00 00 00:00 00 00 00|00 00 00 00:00 00 00 00     <- this has been censored but it doesn't actually matter
00 00 00 00:00 00 37 8C|3C 00 00 00:00 00 00 00     <- keys have been zeroed out to prevent identification :)
```
Checking the ACs it looks like the the first and second blocks are decrement only, the third is read only and ACs are locked, in detail:
```
"AccessConditionsText": {
   "block12": "read AB; decrement transfer restore AB",
   "block13": "read AB; decrement transfer restore AB",
   "block14": "read AB",
   "block15": "read ACCESS by AB",
   "UserData": "00"
}
```
My tag also has an unused sector that has unlocked ACs, I overwrite the original ones in order to have a block that is `read AB; write B; increment B; decrement transfer restore AB`.  Let's call this block `block62`.

Using libfreefare (which I'm not sure if it does nested auth or not but I think it does) I can do:

- auth 62 with key B
- write `7F FF FF FF 80 00 00 00 7F FF FF FF 0D F2 0D F2` to 62
- auth 62 with key B <--- not sure if this auth is needed but haven't tried without it
- restore 62
- auth 12 with key A
- transfer 12

After this process block 12 has value `7F FF FF FF 80 00 00 00 7F FF FF FF 0D F2 0D F2` so I effectively incremented it even if it should be decrement only.

I talked about this ""attack"" in the RRG discord and nobody seemed to really care so I wasn't sure if it was a widely known exploit or not.

RouNNdeL:
I've split the commands into separate function. The existing write command was also renamed to `mf_classic_auth_write_block`. `mf_classic_write_block` now only executes the write command similarly to how `mf_classic_read_block` works.

lib/nfc/protocols/mifare_classic.c
```diff
@@ -391,6 +405,36 @@ void mf_classic_reader_add_sector(
     }
 }
 
+bool mf_classic_block_to_value(int32_t* value, uint8_t* addr, const uint8_t* block) {
+    uint32_t v = *(uint32_t*)&block[0];
+    uint32_t v_inv = *(uint32_t*)&block[4];
+    uint32_t v1 = *(uint32_t*)&block[8];
+
+    bool val_checks =
```
augustozanellato, line 426: it's missing a `&& block[12] == block[14]`, as is `00000000 FFFFFFFF 00000000 00 FF F0 0F` is considered a valid value block

RouNNdeL: Not according to the specification:

image

RouNNdeL: Ok, I misunderstood your comment

RouNNdeL:
After the latest changes the emulation with a real reader has stopped working. Proxmark tests still pass, but that's not enough.

gornekich:
Wonderful work, thanks!
Proxmark doesn't check parity bits, while readers do. I think the problem might be there.
I will test and review your PR once it is ready

RouNNdeL:
Thanks for the parity tip, saved me a bit of time. Ready for review

RouNNdeL:
Since technically halt is 2 bytes I think I will revert it to just the raw bytes tomorrow. But if you have any suggestions how to do it better I'm open to them.

gornekich:
@RouNNdeL , please run linter:
`./fbt format`

skotopes:
Waiting for @gornekich review

lib/nfc/protocols/mifare_classic.c
```diff
@@ -401,6 +418,36 @@ void mf_classic_reader_add_sector(
     }
 }
 
+bool mf_classic_block_to_value(int32_t* value, uint8_t* addr, const uint8_t* block) {
```
gornekich: please, make input arguments come first and output - last

lib/nfc/protocols/mifare_classic.c
```diff
@@ -401,6 +418,36 @@ void mf_classic_reader_add_sector(
...
+    if(value) {
+        *value = (int32_t)v;
+    }
+    if(addr) {
+        *addr = block[12];
+    }
+    return val_checks;
+}
+
+void mf_classic_value_to_block(uint8_t* block, int32_t value, uint8_t addr) {
```
gornekich: change order here too

lib/nfc/protocols/mifare_classic.c
```diff
@@ -996,37 +1115,50 @@ bool mf_classic_emulator(MfClassicEmulator* emulator, FuriHalNfcTxRxContext* tx_
...
 }
 
+void mf_classic_halt(FuriHalNfcTxRxContext* tx_rx, Crypto1* crypto) {
+    furi_assert(tx_rx);
+
+    uint8_t plain_data[2] = {};
+    plain_data[0] = 0x50;
+    plain_data[1] = 0x00;
+
+    nfca_append_crc16(plain_data, 2);
```
gornekich: buffer overflow. should be
```
uint8_t plain_data[4] = {0x50, 0x00};
nfca_append_crc16(plain_data, 2);
```

lib/nfc/protocols/mifare_classic.c
```diff
@@ -996,37 +1115,50 @@ bool mf_classic_emulator(MfClassicEmulator* emulator, FuriHalNfcTxRxContext* tx_
...
+    if(crypto) {
+        crypto1_encrypt(crypto, NULL, plain_data, 2 * 8, tx_rx->tx_data, tx_rx->tx_parity);
+    } else {
+        memcpy(tx_rx->tx_data, plain_data, 2);
+        nfc_util_odd_parity(tx_rx->tx_data, tx_rx->tx_parity, 2);
+    }
+    tx_rx->tx_bits = 2 * 8;
```
gornekich: why 2 bytes? halt+crc = 4 bytes

lib/nfc/protocols/mifare_classic.c
```diff
@@ -1067,17 +1199,188 @@ bool mf_classic_write_block(
...
+            FURI_LOG_D(TAG, "Not ACK received");
+            return false;
+        }
+    } else {
+        FURI_LOG_D(TAG, "Failed to send transfer cmd");
+        return false;
+    }
+
+    return true;
+}
```
gornekich: please, avoid return not in the end of function. Use instead do {} while(false); construction instead.

gornekich (changes requested):
Tested your PR, it works fine! Thank you!

I noticed some problems, please fix them and we will merge

lib/nfc/protocols/mifare_classic.c
```diff
@@ -996,37 +1115,49 @@ bool mf_classic_emulator(MfClassicEmulator* emulator, FuriHalNfcTxRxContext* tx_
...
 }
 
+void mf_classic_halt(FuriHalNfcTxRxContext* tx_rx, Crypto1* crypto) {
+    furi_assert(tx_rx);
+
+    uint8_t plain_data[4] = {0x50, 0x00, 0x00, 0x00};
+
+    nfca_append_crc16(plain_data, 2);
+    if(crypto) {
+        crypto1_encrypt(crypto, NULL, plain_data, 2 * 8, tx_rx->tx_data, tx_rx->tx_parity);
```
gornekich: Looks like you still encrypt 2 bytes instead of 4

lib/nfc/protocols/mifare_classic.c
```diff
@@ -996,37 +1115,49 @@ bool mf_classic_emulator(MfClassicEmulator* emulator, FuriHalNfcTxRxContext* tx_
...
+    } else {
+        memcpy(tx_rx->tx_data, plain_data, 2);
```
gornekich: same here

gornekich (changes requested):
Please, look closely on halt command.

RouNNdeL:
My bad, I was during it in a hurry. By the way do you not have the persmion to modify the branch? I think I selected that checkbox. It would remove the need to go back an forth for those simple things.

RouNNdeL:
Considering in how many places I changed that 2 to 4, maybe it would be better to use `sizeof`?

gornekich:
Yes, please use sizeof() where it's possible.
I have permission to merge dev to your branch. But I can't push commits to your repo

RouNNdeL:
I was looking through the code base to make the `sizeof` usage consistent, and found this in the `read` function: 

https://github.com/flipperdevices/flipperzero-firmware/blob/01a9854f8af8898a9a1138f1e1f496ff74b51b8d/lib/nfc/protocols/mifare_classic.c#L554-L560

I assumed it was a bug, and changed it to `8`. Please correct me if there was any reason to count the parity bits in this instance.

gornekich:
> I assumed it was a bug, and changed it to `8`. Please correct me if there was any reason to count the parity bits in this instance.

Yes, it was a bug. Thanks for fixing it. I will test again PR today

gornekich (approved):
Great work, thank you!

skotopes (approved):

mattiacolombomc:
Can I send these commands with CLI?

PR #2320, opened 2023-01-20
Rename "Detect reader"

team-orangeBlue:
Okay, I'm not quite glad with this function being vaguely named. Obtain MiFare Classic keys seems to be a way better and more descriptive name.

# What's new

1. The MFKey32 function is now named "Obtain MF Classic keys" to avoid misunderstanding and confusion

# Verification 

1. Compile
2. Open the NFC app
3. [new value is there and fits]

# Checklist (For Reviewer)

- [Y] PR has description of feature/bug or link to Confluence/Jira task
- [Y] Description contains actions to verify feature/bugfix
- [Y] I've built this code, uploaded it to the device and verified feature/bugfix
Image below shows all of the changes this PR contains

### DIFF
--- a/applications/main/nfc/scenes/nfc_scene_start.c
+++ b/applications/main/nfc/scenes/nfc_scene_start.c
@@ -23,7 +23,7 @@ void nfc_scene_start_on_enter(void* context) {
 
     submenu_add_item(submenu, "Read", SubmenuIndexRead, nfc_scene_start_submenu_callback, nfc);
     submenu_add_item(
-        submenu, "Detect Reader", SubmenuIndexDetectReader, nfc_scene_start_submenu_callback, nfc);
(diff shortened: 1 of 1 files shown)


Astrrra:
The current version of that feature is not finished. In the future, it will automatically **detect** the type of reader, show information about it, and run the necessary attack (which might not be MFKey32, e.g. Ultralight password sniffing). We understand that it is currently a bit confusing, but renaming it now means that it'll be even more confusing in the future, and will require another renaming back to the original.

team-orangeBlue:
Perhaps keep the "detect reader", and add a "and obtain keys"?

anna-oake:
Anyways, "keys" won't be the correct word here since this feature only collects nonces.

gornekich:
Hi @team-orangeBlue 
As @Astrrra said, Detect Reader will do much more than collecting nonces for mfkey attack in future. And it's better to have the final view of NFC main menu early for people who writes documentation.

PR #2344, opened 2023-01-29
Furi: getter for current thread stdout write callback

Milk-Cool:
# What's new

Added a getter for the current thread stdout callback

# Verification 

Try to get the current callback in an app.

# Checklist (For Reviewer)

### DIFF
--- a/firmware/targets/f7/api_symbols.csv
+++ b/firmware/targets/f7/api_symbols.csv
@@ -1,5 +1,5 @@
 entry,status,name,type,params
-Version,+,11.8,,
+Version,+,11.9,,
 Header,+,applications/services/bt/bt_service/bt.h,,
 Header,+,applications/services/cli/cli.h,,
 Header,+,applications/services/cli/cli_vcp.h,,
@@ -1389,8 +1389,8 @@ Function,+,furi_hal_version_uid_size,size_t,
 Function,-,furi_hal_vibro_init,void,
 Function,+,furi_hal_vibro_on,void,_Bool
 Function,-,furi_init,void,
-Function,+,furi_kernel_is_irq_or_masked,_Bool,
 Function,+,furi_kernel_get_tick_frequency,uint32_t,
+Function,+,furi_kernel_is_irq_or_masked,_Bool,
 Function,+,furi_kernel_lock,int32_t,
 Function,+,furi_kernel_restore_lock,int32_t,int32_t
 Function,+,furi_kernel_unlock,int32_t,
@@ -1515,6 +1515,7 @@ Function,+,furi_thread_get_name,const char*,FuriThreadId
 Function,+,furi_thread_get_return_code,int32_t,FuriThread*
 Function,+,furi_thread_get_stack_space,uint32_t,FuriThreadId
 Function,+,furi_thread_get_state,FuriThreadState,FuriThread*
+Function,+,furi_thread_get_stdout_callback,FuriThreadStdoutWriteCallback,
 Function,+,furi_thread_is_suspended,_Bool,FuriThreadId
 Function,+,furi_thread_join,_Bool,FuriThread*
 Function,+,furi_thread_mark_as_service,void,FuriThread*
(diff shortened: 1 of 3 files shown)


skotopes:
And use cases are?

Milk-Cool:
> And use cases are?

I know this is a little specific, but you may need to set the callback when there already is one (for example, set by the cli service or by yourself earlier).

Milk-Cool:
Here's a code snippet where this can be useful:
```c
static FuriString* execute_command_listen(FuriString* command, Cli* cli){
  FuriString* output = furi_string_alloc_set_str("");
  void cb(const char* data, size_t size){
    UNUSED(size);
    furi_string_cat_str(output, data);
  }
  FuriThread* thread = furi_thread_get_current();
  FuriThreadStdoutWriteCallback old_cb = thread->output.write_callback;
  furi_thread_set_stdout_callback(cb);
  execute_command(command, cli, false);
  furi_thread_set_stdout_callback(old_cb);
  return output;
}
```

This function executes a command and returns its output as a FuriString. However it won't work since the type FuriThread is not public, so we need th replace lines 7 & 8 with `FuriThreadStdoutWriteCallback cb = furi_thread_get_stdout_callback()`.
This function can be called by other commands so I need it to be able to work recursively.

skotopes:
> Here's a code snippet where this can be useful:
> 
> ```c
> static FuriString* execute_command_listen(FuriString* command, Cli* cli){
>   FuriString* output = furi_string_alloc_set_str("");
>   void cb(const char* data, size_t size){
>     UNUSED(size);
>     furi_string_cat_str(output, data);
>   }
>   FuriThread* thread = furi_thread_get_current();
>   FuriThreadStdoutWriteCallback old_cb = thread->output.write_callback;
>   furi_thread_set_stdout_callback(cb);
>   execute_command(command, cli, false);
>   furi_thread_set_stdout_callback(old_cb);
>   return output;
> }
> ```
> 
> This function executes a command and returns its output as a FuriString. However it won't work since the type FuriThread is not public, so we need th replace lines 7 & 8 with `FuriThreadStdoutWriteCallback cb = furi_thread_get_stdout_callback()`. This function can be called by other commands so I need it to be able to work recursively.

Do you have application that already uses this API?

Milk-Cool:
> Do you have application that already uses this API?

Yes I am writing an application, I took the code snippet from its code

Milk-Cool:
If you are wondering if I've tested the function - yes I have

skotopes:
> If you are wondering if I've tested the function - yes I have

Not really. More like all changes to Furi need solid reasoning why. Usually it's some real use examples.

Milk-Cool:
> Not really. More like all changes to Furi need solid reasoning why. Usually it's some real use examples.

Alright so I'm writing an extra commands pack because I felt like the set of commands already included in the firmware lacked a lot of necessary ones. And there's the `if` command which takes two other commands as an argument and if command1 output is non-empty, it executes command2. The catch here is if we want to combine it with the `not` command which is kinda self-explanatory, both use this function and the function eventualy calls itself and if it sets the callback to default cli callback it just breaks everything because the other callback should now be listening to output of the not command but it was replaced by the default callback set by the cli service.

Also sorry for my bad english

skotopes:
> > Not really. More like all changes to Furi need solid reasoning why. Usually it's some real use examples.
> 
> Alright so I'm writing an extra commands pack because I felt like the set of commands already included in the firmware lacked a lot of necessary ones. And there's the `if` command which takes two other commands as an argument and if command1 output is non-empty, it executes command2. The catch here is if we want to combine it with the `not` command which is kinda self-explanatory, both use this function and the function eventualy calls itself and if it sets the callback to default cli callback it just breaks everything because the other callback should now be listening to output of the not command but it was replaced by the default callback set by the cli service.
> 
> Also sorry for my bad english

Sounds like scripting language for CLI.

Ok, we'll accept this PR. But most likely we are not going to accept CLI scripting language.

skotopes (approved):

PR #2350, opened 2023-01-31
[FL-3057] Allow use of any suitable pin for 1-Wire devices

gsurkov:
# What's new

- Decouple 1-Wire classes from iButton HAL
- Add a 1-Wire demo application (DS18B20)

# Verification 

1. Complile and flash.
2. Use the **iButton** applicaton. It should function normally with all types of supported keys in all modes (Read, Write, Emulate).
4. Run the the `example_thermo` application: `./fbt launch_app APPSRC=example_thermo`.
5. Connect a DS18B20 sensor to the Flipper's GPIO in the following configuration:

| DS18B20 | Flipper |
| :-----: | :-----: |
| VDD | 9 |
| GND | 18 |
| DQ  | 17 |

6. Ensure that the temperature is displayed correctly.
7. Change `THERMO_GPIO_PIN` on line 46 in `applications/examples/example_thermo/example_thermo.c` to any of the options listed above it.
8. Exit the application (if running) and repeat step 4.
9. Connect the sensor's DQ pin to the GPIO that is being displayed by the application along with a 4.7k pull-up resistor to GPIO pin 9 (internal pull-up resistors are not enabled).
10. Repeat step 6.

# Caveats
- iButton-related functions (`furi_hal_ibutton_...`, `ibutton_worker`) still work only with GPIO pin 17 because of the specific hardware requirements (RFID part, iButton pogo connector).

# Checklist (For Reviewer)


ImKyleJK:
Would there be a way to "enable" the internal pull-up resistors, or are there none that can be used. As this would be a good feature implementation.

gsurkov:
@ImKyleJK it is possible, but they are not enabled by design. 

A typical value of the internal pull-up resistors is around 40 kOhms, which is almost 10 times the recommended value for most 1-Wire applications. It might work in some cases, but won't in others, so it would be best to force the users to connect an external resistor and thus avoid all potential problems.

skotopes (approved):

PR #2366, opened 2023-02-06
crypto: Erase key from RAM after storing into enclave

Bonstra:
When storing a new unique secret key in the secure enclave, it is temporarily stored in a stack buffer accessible by CPU1. Since it is a secret key, it should not be kept in memory as it could be leaked.

This commit calls the `explicit_bzero()` function from the libc to ensure that the buffer containing the key is cleared. Unlike with `bzero()` and `memset()`, the compiler won't optimize away calls to `explicit_bzero()`.

# What's new

- This commit ensures that the buffer containing the secret key is cleared after use.

# Verification 

- In furi_hal_crypto_store_add_key() comment out the call to `SHCI_C2_FUS_StoreUsrKey()` and return `SHCI_Success` instead, so you do not perform unwanted permanent modification to the enclave slots.
- Add `FURI_LOG_I()` calls to dump the contents of the key buffer after the calls to `explicit_bzero()`.
- Create an external FAP which calls `furi_hal_crypto_store_add_key()` and `furi_hal_crypto_verify_key()`
- Flash the new firmware
- Run the `log` command on the serial console to display the log messages
- Run the test FAP
- Ensure that the keys buffers dumped in the logs are all 0's.
- To double-check, comment the calls to `explicit_bzero()`, rebuild and flash the firmware, and notice that the key buffer content is not all 0's.

# Checklist (For Reviewer)

### DIFF
--- a/applications/services/crypto/crypto_cli.c
+++ b/applications/services/crypto/crypto_cli.c
@@ -276,6 +276,7 @@ void crypto_cli_store_key(Cli* cli, FuriString* args) {
         }
     } while(0);
 
+    explicit_bzero(data, sizeof(data));
     furi_string_free(key_type);
 }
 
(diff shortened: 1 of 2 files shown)


skotopes:
Doesn't make much sense:

- it only wipes memory on stack, but args are also stored in a couple other places
- there is another secure way to deploy keys where key is in encrypted form and decrypted on second core
- small arm cores are not about security, nor operation systems for them

PR #2367, opened 2023-02-07
[FL-3075] Pin Reset

skotopes:
# What's new

- Pin reset 

# Verification 

- Compile and upload
- Set pin
- Forget it
- Reset with new hardware combo: up + back (~30 seconds)

# Checklist (For Reviewer)


DrZlo13 (approved):
It should be clarified that to reset the pin you need to hold the up+back combination until the confirmation window appears after the flipper is reset, and that the flipper must be disconnected from the USB.

PR #2418, opened 2023-02-21
[FL-3123] Reset sector cache when the SD card isn't working

Astrrra:
# What's new

- The SD card cache gets reset on every error

# Verification 

- Use the Flipper as you would normally, check for any abnormal behavior
- Remove/swap/reconnect the SD card in the process of using the Flipper and check that it doesn't slow down or crash

# Checklist (For Reviewer)


firmware/targets/f7/fatfs/sd_spi_io.c
```diff
@@ -230,6 +230,12 @@ static uint8_t sd_spi_wait_for_data_and_read(void) {
     return responce;
 }
 
+SdSpiStatus sd_spi_error(SdSpiStatus status) {
```
DrZlo13: static SdSpiStatus sd_spi_error(void)

firmware/targets/f7/fatfs/sd_spi_io.c
```diff
@@ -230,6 +230,12 @@ static uint8_t sd_spi_wait_for_data_and_read(void) {
...
+    // Reset the sector cache and return the error
+    sector_cache_init();
+    return status;
```
DrZlo13, line 236: return SdSpiStatusError;

DrZlo13 (changes requested):

PR #2429, opened 2023-02-25
Fix animation load fail, size too big

WillyJL:
# What's new

Maybe I'm interpreting the situation wrong, so please correct me if I'm wrong, but:

- Here the `.bm` data is prefixed with either 1 or 2 bytes to identify if it is encoded or not... (`b"\x01\x00"`)
https://github.com/flipperdevices/flipperzero-firmware/blob/eaf965c66f7e57855f2cd9be8247882070bf6453/scripts/flipper/assets/icon.py#L107-L111

- While here it calculates a max `.bm` filesize based on width and height, but only adds 1 byte... (`+ 1`)
https://github.com/flipperdevices/flipperzero-firmware/blob/eaf965c66f7e57855f2cd9be8247882070bf6453/applications/services/desktop/animations/animation_storage.c#L296

While debugging some problematic animations I noticed it sometimes said that the filesize exceeded the max by just 1 byte, and seeing that line of code was my first thought of what's causing it... and indeed from my testing bumping it to `+ 2` allows those problematic animations to be loaded without issues... I had made this change 3 weeks ago and forgot about it, but no one has complained about it causing further issues, and actually I've seen less people asking about animations mysteriously not loading...

# Verification 

- I don't have the original animation that made me look into this at hand anymore, but the explanation above should be sufficient I hope...
- I just have an old log screenshot that shows:
  ```
  [E][AnimationStorage] Filesize 1026, max: 1025 (width 128, height 64)
  [E][AnimationStorage] Load '...' failed, 128x64, size: 1026
  [E][AnimationManager] Can't upload animation described in manifest: '...'
  [I][AnimationManager] Select '(default)' animation
  ```

# Checklist (For Reviewer)


skotopes:
Can you also provide example file?

skotopes:
Fix is incorrect, should be applied on icon compressor side.
Also your branch is not based on our repository and we can not push correct fix to it, so I'm closing this PR.

WillyJL:
> Fix is incorrect, should be applied on icon compressor side.
> Also your branch is not based on our repository and we can not push correct fix to it, so I'm closing this PR.

I agree that the core issue is in the compressor, but at this point making this change would break A LOT of existing animations... this seemed like the easiest solution...

And the branch IS based on your repo? I made the branch on a fork of a fork of you repo, but the branch was based off of flipperdevices/dev. Can make a new one in case if the repo really is an issue (although you were able to push a merge commit so again I don’t understand the issue)

skotopes:
Technically speaking existing animations are incorrectly generated, they are bigger than they should.
There is only one correct fix: regenerate them after we merge this PR: #2434

PR #2431, opened 2023-02-26
Improve vscode clangd experience

dogtopus:
# What's new

- Resolve and use absolute path for the toolchain. This allows clangd to use `compile_commands.json` file without running under fbtenv, simplifies setup for vscode clangd extension (AFAIK clangd solely relies on PATH to look for the compiler drivers instead of having a way to specify the location like cpptools. There seems to be an option to specify a compiler override but it seems to be related to `compile_flags.txt` only and doesn't work for this use case). As a side effect, a rebuild is needed to update `compile_commands.json` after moving the source tree but it shouldn't cause much trouble in practice.
- Add the recommended default settings of the extension to `settings.json`.

# Verification 

- (If have `ms-vscode.cpptools`) Disable `ms-vscode.cpptools`.
- Install `llvm-vs-code-extensions.vscode-clangd`.
- Run `./fbt` and `./fbt vscode_dist LANG_SERVER=clangd`, then open the project in vscode.
- Navigate to `furi/flipper.c` and other files.
- Everything should be correctly indexed including the libc headers.

# Checklist (For Reviewer)


scripts/fbt_tools/crosscc.py
```diff
@@ -15,7 +16,7 @@
 def prefix_commands(env, command_prefix, cmd_list):
     for command in cmd_list:
         if command in env:
-            env[command] = command_prefix + env[command]
+            env[command] = shutil.which(command_prefix + env[command])
```
hedger, line 16: If that's for generating absolute paths in the .json file - it shouldn't be done like that. 
See `COMPILATIONDB_USE_ABSPATH` parameter for `compilation_db` tool.

dogtopus: The name `COMPILATIONDB_USE_ABSPATH` is somewhat misleading. It only makes the paths of source and object files absolute and not the toolchain. In this case only absolute path for the **toolchain** is needed. clangd finds all the **source code** using relative paths just fine ~~since IIRC there's a prefix included in the file~~. It's only the **toolchain** that clangd will try to find under `PATH` and there's no real way that I know to run VSCode clangd under `fbt_env` without using some even nastier hacks, thus this patch.

This quirk is mentioned [in this issue](https://github.com/SCons/scons/issues/3355#issuecomment-630481236) by an SCons developer.

(I guess in the long term we should specify a compiler path prefix with the feature proposed in clangd/vscode-clangd#198 but meanwhile this patch would make clangd immediately usable without needing to resort to wrap VSCode or the whole user VSCode is running on under `fbt_env` for example.)

hedger: We have a fork of compilation_db tool, I think it would be better to implement that within it, without using absolute tool paths everywhere.

Anyway, regarding this PR: I couldn't make it work under Windows. clangd seems to be up and running, but it doesn't load the compilation database from /build/latest. What's your configuration where this works?

Also please add clangd extension to recommended plugins configuration shipped with fbt, and set cpptools as "unwanted" in it.

skotopes:
Please fix all issues and un-draft

scripts/fbt_tools/crosscc.py
dogtopus: > Anyway, regarding this PR: I couldn't make it work under Windows. clangd seems to be up and running, but it doesn't load the compilation database from /build/latest. What's your configuration where this works?

My setup is Linux+VSCodium+vscode-clangd and just loading the project under that.

Are there any indications in the log of what went wrong?

> Also please add clangd extension to recommended plugins configuration shipped with fbt, and set cpptools as "unwanted" in it.

Not sure if we want to move to clangd wholesale. Maybe we should have an alternative config that does what you say?

dogtopus: OK so I dug deeper into this and turns out it's way, way more complicated than I initially thought.

`--query-driver` option expects a fnmatch pattern and it's extremely fragile in terms of handling path separators as everything needs to be prefect or it would refuse to match correctly (and that `${pathSeparator}` doesn't work for some reason) (See clangd/clangd#1518). This also prevented me from using project-relative paths for obvious reasons. I don't see a good way of implementing this other than generating `settings.json` from fbt, which is kind of intrusive and I don't really know much about scons to pull this off confidently.

(There's also the "F\*\*k it" option of just setting query driver to `**` ~~or `${workspaceFolder}**`~~ to effectively bypass it, but there could (?) be security implications.)

As to the absolute tool path, looks like the `compilation_db` just takes pre-built commands from elsewhere and use them, which are a pain to work with, and doing a string replacement isn't a real option here. So looks like the best way would still be providing absolute tool paths to scons from the very beginning (or somehow make clangd resolve extra `PATH` set by fbt without writing absolute paths to `compile_commands` which is like, good luck.)

@hedger thoughts?

dogtopus: > Anyway, regarding this PR: I couldn't make it work under Windows. clangd seems to be up and running, but it doesn't load the compilation database from /build/latest. What's your configuration where this works?
> 
> Also please add clangd extension to recommended plugins configuration shipped with fbt, and set cpptools as "unwanted" in it.

Addressed. See previous comments for tradeoffs made.

For point 1 I still think that changing the tool path from the very beginning seems to be the cleanest approach here.

hedger:
OK, I'll have a second look at this. 

What are the benefits of using clangd over cpptools?

dogtopus:
> 
> What are the benefits of using clangd over cpptools?

It's less about technical benefits and more about giving more choices to the developers, since cpptools is proprietary and doesn't ship with most third-party/FOSS vscode distributions. Meanwhile clangd is on openvsx and therefore is pretty much available everywhere.

That's why I mentioned that we should keep cpptools as the default and offer clangd as an option instead of flat out replacing cpptools.

(Actually there's one: clangd ships with the full capability of clang-tidy so just saving the file would automatically format it to spec.)

hedger (approved):

PR #2446, opened 2023-02-27
[#2446] clock_app-am-to-pm-patch

Ethanol0001:
# What's new
- fixed the issue where the clock would switch from am to pm at 1pm instead of 12pm
- changed the greater than from 12 to 11 so it would switch right as it hits 12 instead of 1 (hour 13 for military time)
# Verification 

- set date and time to 11:55am while updating to change the flippers internal clock and notice if it switches at 12pm instead of 1pm which the release firmware is set to do

- I have uploaded the patched firmware to my flipper to check if it works and everything seems fine
before: 
IMG_8832
after: 
IMG_8833

- firmware to test the update can be found here:
[f7-D.zip](https://github.com/flipperdevices/flipperzero-firmware/files/10856490/f7-D.zip)

# Checklist (For Reviewer)


itsweekly:
seems good to me

skotopes (approved):

PR #2447, opened 2023-02-28
Fix CAME 12bit pilot bit

### LINKED ISSUE #2351: Regression with SubGHZ CAME
### Describe the bug.

A change between 0.74.2 and 0.75.0-rc causes a regression in SubGHZ functionality.
I presume PR #2210 is the culprit but I'm unsure

Before this was merged controlling my fan worked perfectly; however after this commit I have to hold down send for several seconds to get my ceiling fan to respond.

If needed I can take I/Q recordings with an SDR of the original transmission, the transmission before this update, and the transmission after this update.

### Reproduction

1. Record a CAME transmission
2. Attempt to replay it
3. It requires many multiple replays before the appliance reacts

Expected result: a short press should cause the appliance to react

### Target

_No response_

### Logs

_No response_

### Anything else?

_No response_

dani84bs:
This commit fixes a regression on the CAME 12bit protocol, header encoding and fix #2351 .

# What's new

I changed the way the pilot bit is chosen depending on the current protocol
# Verification 

To check that this fix is working I tried it on my gate :D

### DIFF
--- a/lib/subghz/protocols/came.c
+++ b/lib/subghz/protocols/came.c
@@ -120,10 +120,9 @@ static bool subghz_protocol_encoder_came_get_upload(SubGhzProtocolEncoderCame* i
     instance->encoder.upload[index++] = level_duration_make(
         false,
         (((instance->generic.data_count_bit == CAME_24_COUNT_BIT) ||
-          (instance->generic.data_count_bit ==
-           subghz_protocol_came_const.min_count_bit_for_found)) ?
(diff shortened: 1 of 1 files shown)


Skorpionm:
original keys stop working

dani84bs:
Actually I have an original key and didn't work, now it does.
I couldn't try for CAME24 or PRASTEL, because I don't have any, I have an old CAME12 bit fixed code (from 1994)

skotopes:
@dani84bs your issue was fixed in #2443, please check and let us know if it still persist

PR #2451, opened 2023-03-02
Support reseting iCx cards

bettse:
# What's new

- Support resetting iClass cards to iCL/iCS/iCE (https://github.com/iCopy-X-Community/icopyx-teardown/blob/master/tags/README.md#icl)

# Verification 

- Scan iClass card with standard key, see new options.
- Choose one of the options
- Use card with iCopyX (successfully)

# Checklist (For Reviewer)


bettse:
@nvx  / @pcunning Thoughts?

nvx:
Nice. Is it worth adding the iCopy keys to the key search so we can read cards with iCopy keys? (Unless it already does this and I missed the code)

pcunning:
Thanks for implementing my external feedback about moving the menu items into a folder. 

I agree with @nvx that if we are making it this easy to change a key to something it needs to be easy to change back to a standard key. That means we can read cards in a given format they end up in. Biggest issue I see with that is we don’t have a good way to decide what key algorithm we use with the keys. We could just do both or have a way to indicate it in a dictionary file. If we do allow people to input their own key at some point (ui or based on a file) it would be good to make sure that key is also in the user dictionary.

bettse:
So they keys are in the dictionary: https://github.com/flipperdevices/flipperzero-firmware/blob/dev/assets/resources/apps_data/picopass/assets/iclass_elite_dict.txt#L24 but I think it wouldn't find them because they use the standard kdf and not the elite kdf.  Maybe we could run through the system one twice, once elite and once standard?  

I like the idea of having a user input for the key, but that is outside the scope of this to me.

pcunning:
Running through them twice seems good to me.

bettse:
> Running through them twice seems good to me.

Done! 😁

applications/plugins/picopass/scenes/picopass_scene_read_factory_success.c
```diff
@@ -1,6 +1,8 @@
 #include "../picopass_i.h"
 #include <dolphin/dolphin.h>
 
+extern const uint8_t picopass_iclass_key[];
```
skotopes, line 4: The best will be to move this definition to separate .h/.c file

skotopes (approved):

applications/plugins/picopass/scenes/picopass_scene_read_factory_success.c
bettse: I'll try to remember to do that in a future PR, but I'm also OK if someone else does, maybe a good contribution for a new person.

bettse: I was playing with this, but I don't think I'm familiar with the pattern.  Are their examples in the codebase you can point me to?

skotopes: https://github.com/flipperdevices/flipperzero-firmware/blob/0.75.0/firmware/targets/f7/furi_hal/furi_hal_resources.h
https://github.com/flipperdevices/flipperzero-firmware/blob/0.75.0/firmware/targets/f7/furi_hal/furi_hal_resources.c

PR #2458, opened 2023-03-03
[#2458] Archived Holidays Animation + Added Archived Animations Directory

Ethanol0001:
# What's new

- Added back the Happy holidays animation to be archived to make things easier when re-adding it for future holiday seasons. This will be in a separate directory from the external directory in the Dolphin directory. Archiving the animation folder will ensure the animations safety from accidently being lost as it will be hard-built into the code.
- Added the Archived Animations folder for animations to go when they are no longer being used. This will also allow for more customizability for users if they wish to re-add them back.

# Verification 

- check the Dolphin directory to see if the Archived Animations folder is there and that there is nothing new in the external directory and only the Archived Animations folder was added in the same directory as the external folder
- ensure that the manifest file does not include the archived animation while it's in the archived animation folder
- code has been uploaded to a flipper for testing and restarted multiple times and the holidays animation did not show up during testing.
- Update file for testing can be downloaded here: 
[f7-D.zip](https://github.com/flipperdevices/flipperzero-firmware/files/10886642/f7-D.zip)

# Checklist (For Reviewer)


skotopes:
It's literally in the history of this repository ;-)

PR #2464, opened 2023-03-05
Fixed picopass load save file overrun

GuruSteve:
# What's new

When reading and saving an unpersonalized picopass tag, the config block app limit is set to 0xFF.  This causes an issue when subsequently loading the save file, as the code tries to parses the unusually large app limit size.  I have therefore modified the code to limit it to the lesser of the block app limit or PICOPASS_MAX_APP_LIMIT constant.

# Verification 

Read and save a picopass card containing a configuration app limit > 32 (0x20).  Then, try to re-load that saved file.  Prior to this correction, it would error out with "Can not parse file"

# Checklist (For Reviewer)


bettse:
Thanks!

skotopes (approved):

PR #2468, opened 2023-03-07
more than 2 favourite desktop apps

RaZer4lin:
Please consider an option like Primary and Secondary Favorite Apps in the Settings/Desktop Menu but adding the left (long and short press) and right press on desktop to choose other apps there too :).


skotopes:
image

PR #2478, opened 2023-03-11
Picopass standard KDF dictionary

bettse:
# What's new

- Breaks up dictionary based on KDF (I want to use this more later when elite dictionary grows)
- Pass worker down further so the state can be detected and cancelled while keys are being tested

# Verification 

- Read card and hit back button: notice that is doesn't get stuck reading and consistently exited quickly to original menu

# Checklist (For Reviewer)

### DIFF
--- a/assets/resources/apps_data/picopass/assets/iclass_standard_dict.txt
+++ b/assets/resources/apps_data/picopass/assets/iclass_standard_dict.txt
@@ -1,10 +1,16 @@
 
 ## From https://github.com/RfidResearchGroup/proxmark3/blob/master/client/dictionaries/iclass_default_keys.dic
 
+# AA1
+AEA684A6DAB23278
 # key1/Kc from PicoPass 2k documentation
 7665544332211000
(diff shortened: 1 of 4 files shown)


assets/resources/apps_data/picopass/assets/iclass_standard_dict.txt
```diff
@@ -0,0 +1,47 @@
...
+# from loclass demo file.
+5b7c62c491c11b39
+# Kd from PicoPass 2k documentation
+F0E1D2C3B4A59687
+# PicoPass Default Exchange Key
+5CBCF1DA45D5FB4F
+# From HID multiclassSE reader
+31ad7ebd2f282168
+# From pastebin: https://pastebin.com/uHqpjiuU
+6EFD46EFCBB3C875
```
bettse, line 19: I don't know which KDF is used on a few of these, so I'm leaving them in both since there is only a dozen

skotopes (approved):

PR #2480, opened 2023-03-11
BadUSB: Advertise HID device as "boot protocol"

CFSworks:
This is necessary for the BadUSB keyboard to be initialized/used in BIOS/UEFI environments.

In my testing, `USB_HID_PROTO_KEYBOARD` did not appear to make any OS ignore mouse events, but I could have missed something. If mouse events are broken by this change, it will be necessary instead to split the USB interface in two: one for keyboard, one for mouse.

# What's new

- Interface descriptor in FuriHal's keyboard+mouse module uses boot protocol/subclass now

# Verification 

- Connect Flipper to a PC on its BIOS setup menu and use USB remote app to navigate menu
- Use BadUSB scripts to enter BIOS setup and do the same
- Regression-test keyboard **and mouse** (remote/BadUSB) functionality on Linux, macOS, and Windows

# Checklist (For Reviewer)


skotopes:
https://www.usb.org/sites/default/files/documents/hid1_11.pdf is clearly states that descriptor and reporting method should be quite specific for boot subclass, so changes you've made are going to make descriptor invalid.

Also all modern UEFI based systems supports new HID specification and it makes little to no sense to support those legacy systems.

CFSworks:
Not all systems can be upgraded to UEFI and the HID interface must implement the boot protocol for the ones stuck on BIOS. I can take some time to change the descriptors too if necessary.

PR #2486, opened 2023-03-13
[FL-1799] Require the trailing slash for root paths

Astrrra:
# What's new

- Storage paths like "/internal", "/international", "/anyone" etc are not valid anymore.

# Verification 

- storage write /int/ernational
- write anything
- storage stat /international should result in an error
- storage read /International should result in an error as well

# Checklist (For Reviewer)


hedger (changes requested):
INT & EXT are swapped

hedger (dismissed):

DrZlo13 (dismissed):
Will not work with `/int` path. For example `storage list /int`.

skotopes (approved):

PR #2492, opened 2023-03-15
iButton: Add support for Dallas DS1971 v2

baconlive:
# What's new

Added support to one_wire lib to read and write Dallas DS1971 chip. DS1971 is an iButton with 8 bytes fixed key (without CRC and starting with 0x14) and it contains 3 memory areas, one 8 byte application register (writeable only once), a temporary 32 bytes scratchpad and a 32 bytes re-writable memory. The code added allow a user to read key and memory from a DS1971 and save it as file then write memory contents saved into the file on the same iButton or on a different one. Code refactored to use refactored iButton framework.

# Verification 

Read a DS1971, save contents to sd-card, modify it, write it back on the DS1971 (or on a different one) and the read it again to check new data written on the memory.

# Checklist (For Reviewer)


lib/one_wire/ibutton/protocols/dallas/protocol_group_dallas_defs.h
```diff
@@ -7,6 +7,7 @@ typedef enum {
     iButtonProtocolDS1992,
     iButtonProtocolDS1996,
     /* Add new 1-Wire protocols here */
+    iButtonProtocolDS1971,
```
gsurkov, line 9: Please move up 1 line so that the comment stays below it.

lib/one_wire/ibutton/protocols/dallas/protocol_group_dallas_defs.c
```diff
@@ -3,13 +3,15 @@
...
 #include "protocol_ds1996.h"
+#include "protocol_ds1971.h"
 #include "protocol_ds_generic.h"
 
 const iButtonProtocolDallasBase* ibutton_protocols_dallas[] = {
     [iButtonProtocolDS1990] = &ibutton_protocol_ds1990,
     [iButtonProtocolDS1992] = &ibutton_protocol_ds1992,
     [iButtonProtocolDS1996] = &ibutton_protocol_ds1996,
     /* Add new 1-Wire protocols here */
+    [iButtonProtocolDS1971] = &ibutton_protocol_ds1971,
```
gsurkov, line 13: Please move up 1 line so that the comment stays below it.

lib/one_wire/ibutton/protocols/dallas/protocol_ds1971.c
```diff
@@ -0,0 +1,219 @@
...
+static void dallas_ds1971_render_error(FuriString*, const iButtonProtocolData*);
+static bool dallas_ds1971_is_data_valid(const iButtonProtocolData*);
+static void dallas_ds1971_get_editable_data(iButtonEditableData*, iButtonProtocolData*);
+static void dallas_ds1971_apply_edits(iButtonProtocolData*);
+static bool
+    dallas_ds1971_read_mem(OneWireHost* host, uint8_t address, uint8_t* data, size_t data_size);
+
+const iButtonProtocolDallasBase ibutton_protocol_ds1971 = {
+    .family_code = DS1971_FAMILY_CODE,
+    .features = iButtonProtocolFeatureExtData | iButtonProtocolFeatureWriteBlank |
```
gsurkov: You have `iButtonProtocolFeatureWriteBlank` set but the `write_blank` method is NULL. This code will crash when the "Write Blank" menu option is selected.

lib/one_wire/ibutton/protocols/blanks/ds1997.c
gsurkov, line 1: DS1971 is not a blank, so there should not be any `ds1971.*` files in the `blanks` folder. Instead, the `ds1971_write` method should be located either in `protocol_ds1971.c` or in `dallas_common.c` if you think that it could be reused by other protocols.

General info on blanks: these are special (usually 3rd-party) keys that have a writable ROM sector (which normal keys like DS1971 do not). Some of them also posess additional memory to store larger amounts of data.

Right now only 2 types of blanks are supported:
- RW1990 (only allows to write DS1990 keys on it)
- TM2004 (can accept any ROM, but is broken beyond that).

I would advise to leave out the blank functionality for now (which you kinda did already).

gsurkov (changes requested):

skotopes:
Please fix issues mentioned by @gsurkov and un-draft PR

baconlive (commented):
Fixed the issues requested by @gsurkov

lib/one_wire/ibutton/protocols/dallas/protocol_ds1971.c
```diff
@@ -0,0 +1,258 @@
...
+
+#define DS1971_SRAM_DATA_SIZE 32U
+#define DS1971_SRAM_PAGE_SIZE 32U
+#define DS1971_COPY_SCRATCH_TIMEOUT_US 250U
+
+#define DS1971_DATA_BYTE_COUNT 4U
+
+#define DS1971_SRAM_DATA_KEY "Sram Data"
+
+#define DS1971_CMD_READ_MEMORY 0xF0
```
gsurkov: Use `DALLAS_COMMON_CMD_` macros if the required command codes were already defined in `dallas_common.h`

lib/one_wire/ibutton/protocols/dallas/protocol_ds1971.c
```diff
@@ -0,0 +1,258 @@
...
+#define DS1971_FAMILY_CODE 0x14U
+#define DS1971_FAMILY_NAME "DS1971"
+
+#define DS1971_SRAM_DATA_SIZE 32U
+#define DS1971_SRAM_PAGE_SIZE 32U
+#define DS1971_COPY_SCRATCH_TIMEOUT_US 250U
+
+#define DS1971_DATA_BYTE_COUNT 4U
+
+#define DS1971_SRAM_DATA_KEY "Sram Data"
```
gsurkov: According to the [iButton file format docs](https://github.com/flipperdevices/flipperzero-firmware/blob/dev/documentation/file_formats/iButtonFileFormat.md), protocols can specify their own custom data fields.
Since DS1971 is a EEPROM device, this key should be "Eeprom Data" to reflect that. This also should be added to the aforementioned documentation.

lib/one_wire/ibutton/protocols/dallas/protocol_ds1971.c
```diff
@@ -0,0 +1,258 @@
...
+    DS1971ProtocolData* data = protocol_data;
+
+    onewire_host_reset(host);
+    onewire_host_skip(host);
+    // Starting writing from address 0x0000
+    onewire_host_write(host, DS1971_CMD_WRITE_SCRATCHPAD);
+    onewire_host_write(host, 0x00);
+    // Write data to scratchpad
+    onewire_host_write_bytes(host, data->sram_data, DS1971_SRAM_DATA_SIZE);
+    furi_delay_ms(250);
```
gsurkov: `DS1971_COPY_SCRATCH_TIMEOUT_US` is defined, but not used anywhere. However, there are several mentions of 250 milli_seconds here. Please double-check whether the specification calls for 250 micro_ or milli_seconds and use a `#define` constant.

gsurkov (changes requested):
It's getting there! Some changes are still required before we merge it.

baconlive:
@gsurkov fixed the issue you mentioned

lib/one_wire/ibutton/protocols/dallas/protocol_ds1971.c
```diff
@@ -0,0 +1,253 @@
...
+    const DS1971ProtocolData* data = protocol_data;
+    pretty_format_bytes_hex_canonical(
+        result,
+        DS1971_DATA_BYTE_COUNT,
+        PRETTY_FORMAT_FONT_MONOSPACE,
+        data->eeprom_data,
+        DS1971_EEPROM_DATA_SIZE);
+}
+
+void dallas_ds1971_render_brief_data(FuriString* result, const iButtonProtocolData* protocol_data) {
```
gsurkov, line 216: On the info page, it shows "Internal SRAM: 0 Kbit", while it should be something like "Internal EEPROM: 256 bit".

I am suggesting to modify the `dallas_common_render_brief_data` so that it accepts an additional string (SRAM, EEPROM, etc) and calculates the multiplier (none, K, M etc) based on the `data_size` parameter. The parameter names should also be renamed to something like `mem_data` and `mem_data_size` to reflect the universal nature of this method.

documentation/file_formats/iButtonFileFormat.md
```diff
@@ -24,12 +24,13 @@ Changelog:
...
 
-| Name      | Type   | Description                                  |
-| --------- | ------ | -------------------------------------------- |
-| Protocol  | string | Currently supported: DS1990, DS1992, DS1996, DSGeneric*, Cyfral, Metakom |
-| Rom Data  | hex    | Read-only memory data (Dallas protocols only) |
-| Sram Data | hex    | Static RAM data (DS1992 and DS1996 only)
-| Data      | hex    | Key data (Cyfral & Metakom only)              |
+| Name        | Type   | Description                                  |
+| ----------- | ------ | -------------------------------------------- |
+| Protocol    | string | Currently supported: DS1990, DS1992, DS1996, DSGeneric*, Cyfral, Metakom |
```
gsurkov: DS1971 should be listed here as well.

lib/one_wire/ibutton/protocols/dallas/protocol_ds1971.c
```diff
@@ -0,0 +1,253 @@
...
+
+    switch(command) {
+    case DALLAS_COMMON_CMD_SEARCH_ROM:
+        if(data->state.command_state == DallasCommonCommandStateIdle) {
+            data->state.command_state = DallasCommonCommandStateRomCmd;
+            return dallas_common_emulate_search_rom(bus, &data->rom_data);
+
+        } else if(data->state.command_state == DallasCommonCommandStateRomCmd) {
+            data->state.command_state = DallasCommonCommandStateMemCmd;
+            dallas_common_emulate_read_mem(bus, data->eeprom_data, DS1971_EEPROM_DATA_SIZE);
```
gsurkov: This function doesn't work here because DS1971 uses 1-byte addressing. Either implement a separate `dallas_ds191_emulate_read_mem` as you have done with `dallas_ds1971_read_mem` or adapt both `dallas_common_` functions to take an `address_size` argument and use them instead.

gsurkov (changes requested):
@baconlive After testing the code (including emulating a key and reading it with another Flipper), some changes are still needed.

baconlive:
@gsurkov Fixed the last issues you mentioned

gsurkov (approved):
Made some final fixes, looks like it's good to go!
Thanks @baconlive

baconlive:
> Made some final fixes, looks like it's good to go! Thanks @baconlive

@gsurkov do i need to do anything else or it will be merged automatically?

gsurkov:
@baconlive You don't have to do anything, we'll take it from here.

skotopes (approved):

baconlive:
@gsurkov do you think is it possible to add a "ibutton_protocols_get_editable_memory" function to be able to edit memory contents?

gsurkov:
@baconlive 
**short answer**: yes (maybe in a more abstract form, since not all iButtons have a built-in memory as an additional function).

**long answer**: we don't currently have any plans regarding this functionality, as it would require to come up with a GUI editor that would be able to handle up to 64 Kbits of data (such is the size of DS1996 memory, for example) and be usable enough.

If your goal is making such an editor, I would recommend first starting with an [external applicaton](https://github.com/flipperdevices/flipperzero-firmware/blob/dev/documentation/AppsOnSDCard.md).

baconlive:
> @baconlive **short answer**: yes (maybe in a more abstract form, since not all iButtons have a built-in memory as an additional function).
> 
> **long answer**: we don't currently have any plans regarding this functionality, as it would require to come up with a GUI editor that would be able to handle up to 64 Kbits of data (such is the size of DS1996 memory, for example) and be usable enough.
> 
> If your goal is making such an editor, I would recommend first starting with an [external applicaton](https://github.com/flipperdevices/flipperzero-firmware/blob/dev/documentation/AppsOnSDCard.md).

@gsurkov Thanks for the reply. My plan is to use an extenal app as editor but the problem is that with current API i do not have access to memory data...

gsurkov:
@baconlive  After some consideration, I believe that the editor should not use the iButton API at all.

The main reasons for it are:
- Not all key types posess an internal memory which can be edited in a meaningful way;
- The iButton library functionality is mostly unrelated to editing files (we need only the load and save parts from it);
- The code size must be kept as small as possible since iButton is an internal application which is kept in the flash memory.

All that said, I think the best course of action is just reimplement the parts you need in your application. By using the [iButton file format specs](https://github.com/flipperdevices/flipperzero-firmware/blob/dev/documentation/file_formats/iButtonFileFormat.md), you can easily control the way you load and store .ibtn files, by defining custom data structures and methods for their manipulation. It would lead to some code duplication, but it looks like a lesser evil because it would be contained to an external app with little to no relation to the main firmware codebase.

baconlive:
@gsurkov The solution you suggested is what i was thinking about, i'll use onewire api reimplementing iButton code for the specific needs into the external application.

Thank you.

PR #2500, opened 2023-03-17
ibutton: Add read and emulate support for DS2502 (Dell charger chip)

dogtopus:
# What's new

- Read and emulate DS2502 (write is not supported since the device is OTP and requires a special writing voltage to be able to write.)
- A way for `dallas_common` to handle the leading checksum when reading and emulating DS2502.
- Introduce a new key `Otp Data` for dumped OTP data.
- Fix a typo in `dallas_common`

# Verification 

- Connect 2 Flipper's 1W and GND together. Emulate [dell.txt](https://github.com/flipperdevices/flipperzero-firmware/files/10997245/dell.txt) on Flipper A and read on Flipper B.
  - The Flipper B should now have a dump identical to `dell.txt`.
- (If one can source the old Dell barrel jack chargers) Connect the center pin to Flipper 's `1W` pin and **outer** barrel to `GND` and go to `iButton -> Read`.
  - I highly recommend to NOT power on the charger when testing so a slippery probe won't kill the Flipper. OneWire ROMs are self powered so this shouldn't affect the result.
  - Should be able to read the OTP content (a string starting with `DELL00AC`)

# Checklist (For Reviewer)

### DIFF
--- a/documentation/file_formats/iButtonFileFormat.md
+++ b/documentation/file_formats/iButtonFileFormat.md
@@ -29,7 +29,8 @@ Changelog:
 | Protocol    | string | Currently supported: DS1990, DS1992, DS1996, DS1997, DSGeneric*, Cyfral, Metakom |
 | Rom Data    | hex    | Read-only memory data (Dallas protocols only) |
 | Sram Data   | hex    | Static RAM data (DS1992 and DS1996 only)
-| Eeprom Data | hex    | EEPROM data (DS1971 only)
+| Eeprom Data | hex    | EEPROM data (DS1971 only) |
+| Otp Data    | hex    | OTP (One Time Programmable) data (DS2502 and its variants only) |
 | Data        | hex    | Key data (Cyfral & Metakom only)              |
 
 NOTE 1: DSGeneric is a catch-all protocol for all unknown 1-Wire devices. It reads only the ROM and does not perform any checks on the read data. 
--- a/lib/one_wire/ibutton/protocols/dallas/dallas_common.c
+++ b/lib/one_wire/ibutton/protocols/dallas/dallas_common.c
@@ -91,8 +91,8 @@ bool dallas_common_read_mem(
     size_t data_size,
     bool handle_leading_crc) {
     onewire_host_write(host, DALLAS_COMMON_CMD_READ_MEM);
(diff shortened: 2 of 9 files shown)


bbock:
Validated with an old Dell XPS 15 PSU of mine.
Dell PSU
Screenshot-20230317-141546

gsurkov:
After some consideration, we have decided that only iButton-style devices will be accepted into the iButton application (i.e. ones enclosed in a round metal can).

For the rest of 1-Wire devices, it was decided to make an additional **external** application instead. There are many advantages of doing it this way, including not being confined to the workflow of the original iButton app and not being limited by the remaining flash space.

 I will have to close this PR, but stay tuned for the news about the app - your code will come in handy there!

There is no ETA for the 1-Wire app for now, but it'll be a medum-ish priority task.

dogtopus:
@gsurkov Fair enough. Although one thing to note is that this PR also fixed a [bug](https://github.com/flipperdevices/flipperzero-firmware/blob/7de7fa293b6eb8def92ecfecc1c1aea63a269624/lib/one_wire/ibutton/protocols/dallas/dallas_common.c#L88) that caused the wrong address to be sent in `dallas_common_read_mem` (it should be a right shift instead of a greater than). So far this haven't caused any problem due to how these dumpers work but it's better to fix it so it won't cause strange problems later.

gsurkov:
@dogtopus thanks, that's a pretty nasty bug indeed.

dogtopus:
@gsurkov BTW this chip is available in [iButton form factor under part number DS1982](https://www.analog.com/en/products/ds1982.html) as well. Should I repurpose this PR to support this instead or the generic reader for it is enough?

The leading CRC handling implemented here is also needed to read the OTP registers in DS1982 as they share the same protocol and same family code.

gsurkov:
@dogtopus the generic reader would only read the ROM, but not the OTP EEPROM memory (since it doesn't know of its existence).

I would recommend holding off and waiting for the 1-Wire app because of the following reasons:
- None of us has the actual part (DS1982) on hand and therefore, strictly speaking, cannot test it,
- It does not follow the original idea of this PR (which was interfacing with a Dell charger, not reading some abstract iButton),
- There was no demand for DS1982 in the first place.

PR #2517, opened 2023-03-20
Add Mfkey32 application

noproto:
# What's new

- The Mfkey32 extapp allows the Mfkey32 functions of the Flipper companion app and desktop equivalents to be computed locally on the Flipper Zero

# Verification

- Collect (Mifare Classic 1K) nonces with Detect Reader function of NFC app, and attempt to crack them using the Mfkey32 Flipper application


applications/external/mfkey32/mfkey32.c
```diff
@@ -0,0 +1,1346 @@
...
+void napi_mf_classic_nonce_array_free(MfClassicNonceArray* nonce_array) {
+    furi_assert(nonce_array);
+    furi_assert(nonce_array->stream);
+
+    buffered_file_stream_close(nonce_array->stream);
+    stream_free(nonce_array->stream);
+    free(nonce_array);
+}
+
+static void finished_beep() {
```
hedger, line 991: Better use notification service for that

noproto: I updated the beep function to use the notification service instead :+1:

skotopes:
Couple messages from PVS:
image

Also would you like to do some refactoring? It will be nice if application will be split into UI / worker / algo.

Astrrra:
Hi! I've tried running the app, but it shows the "No nonces to crack" error. My /ext/nfc/.mfkey32.log file contains five valid nonce pairs for key A0A1A2A3A4A5, but the app doesn't seem to notice them. Here are the nonces:
```Sec 1 key A cuid 2a234f80 nt0 03f54b83 nr0 ac4f8972 ar0 2fe100da nt1 b981ed97 nr1 94d7cffe ar1 997979e1
Sec 1 key A cuid 2a234f80 nt0 212734a1 nr0 6b122daa ar0 091c482d nt1 6157eed0 nr1 41f80c0b ar1 7dc4d2f5
Sec 1 key A cuid 2a234f80 nt0 9399092a nr0 a38a7e51 ar0 429f186b nt1 7f97f42e nr1 d79bbde4 ar1 e12df280
Sec 1 key A cuid 2a234f80 nt0 0eb54481 nr0 ac3029f9 ar0 1d98b211 nt1 522e7e3b nr1 c2d3379f ar1 a4414c6c
Sec 1 key A cuid 2a234f80 nt0 b6efd0e4 nr0 e79da01c ar0 80050b9b nt1 51dc2dd4 nr1 083ec7ba ar1 ce0b1c7b
```

Is there something that I'm doing wrong?

noproto:
Hi @Astrrra. The reason you are getting the "No nonces to crack" error is because A0A1A2A3A4A5 is already in the default Flipper system dictionary, so Mfkey32 (correctly) identified that no keys needed to be cracked. If you were to scan a card that required A0A1A2A3A4A5, it would scan successfully.

For testing purposes, you may choose to modify your system dictionary to exclude A0A1A2A3A4A5. Then Mfkey32 will attempt to crack the key. I hope this helps.

noproto:
We'll change the error output to explain this better.

noproto:
> Couple messages from PVS: image

In order:

206: False positive, this never exceeds 32 bits. i is never greater than 31, and filter returns 0 or 1.
482: False positive, and I'm not sure why PVS flagged on this.
505: False positive, same as 482.
908: Good catch. Fixed.
955: Mfkey32 needs at least ~80 KB to run. The Flipper documentation indicates applications are provided 140 KB. While we've seen a slight deviation from this number, we haven't seen significantly less. Therefore realloc should not fail here (and if it does, a reboot is most likely necessary due to an external memory leak).
1003: Updated. This would have leaked 8 bytes.
1091: See 955.
1112: Eliminated some logic around that variable. It was only there to make the code more structured.
43: False positive, the SWAPENDIAN macro is only used in instances where it wouldn't be ambiguous.

The PVS errors led us to some missed coverage, so thanks!

> Also would you like to do some refactoring? It will be nice if application will be split into UI / worker / algo.

I would. I hope it's okay if that's a separate PR. For now it is a bit easier to apply the compiler optimizations via pragma when it is all self contained.

And the error message has been updated accordingly @Astrrra.

noproto:
@hedger @skotopes : I've reviewed support requests for this since March. As far as I can tell it is stable. Users only have issues if they use non-official firmware, Mfkey32 has successfully cracked every key it has been tasked with (on-device).

skotopes:
@noproto I'll try to pack it in this release.

applications/external/mfkey32/mfkey32.c
```diff
@@ -0,0 +1,1348 @@
...
+        *lfsr = *lfsr << 1 | BIT(state->even, i ^ 3);
+    }
+}
+
+static inline uint32_t crypt_word(struct Crypto1State* s) {
+    // "in" and "x" are always 0 (last iteration)
+    uint32_t res_ret = 0;
+    uint32_t feedin, t;
+    for(int i = 0; i <= 31; i++) {
+        res_ret |= (filter(s->odd) << (24 ^ i));
```
skotopes: image

skotopes (changes requested):

skotopes:
Un-draft when all issues are fixed

applications/external/mfkey32/mfkey32.c
noproto: These PVS warnings were previously discussed here: https://github.com/flipperdevices/flipperzero-firmware/pull/2517#issuecomment-1480101194. For example, the "always false" conditions that PVS raises issues with on lines 482 and 505 are not always false. We have tested this function extensively and I'm not sure I understand what PVS wants to see changed in our code.

skotopes: In that case all marked places must contain `//-V{CODE}` at the end of the line

noproto: I made the changes in https://github.com/flipperdevices/flipperzero-firmware/pull/2517/commits/2ada46060f5f8452f641742ba37e8b62ce8b092e. Let me know if PVS still warns on the false positives.

noproto: Ah wait, I understand the format of the comments now. Sorry, I misunderstood. Fixing it now.

Edit: Fixed.

noproto (commented):
Should be ready to review and squash merge if PVS is happy.

skotopes (approved):

Issue #2512, opened 2023-03-19
SDK: HMAC-SHA-256 implementation assumes a 32-byte key


str4d:
### Describe the bug.

The `hmac_sha256_init` function exposed in the SDK is undocumented, but internally calls this helper function:

https://github.com/flipperdevices/flipperzero-firmware/blob/b698126c36a6b491420b673683333102a242455d/lib/toolbox/hmac_sha256.c#L49-L55

The internal documentation and code make clear that this is assuming a key length of 32 bytes, matching the output size of SHA-256.

The reference to RFC 6979 is somewhat misleading here; the more pertinent reference is [RFC 2104](https://www.rfc-editor.org/rfc/rfc2104) which specifies HMAC, and states:
> We denote by `B` the byte-length of such blocks (`B=64` for all the above mentioned examples of hash functions), and by `L` the byte-length of hash outputs (`L=16` for MD5, `L=20` for SHA-1).  The authentication key `K` can be of any length up to `B`, the block length of the hash function.  Applications that use keys longer than `B` bytes will first hash the key using `H` and then use the resultant `L` byte string as the actual key to HMAC. In any case the minimal recommended length for `K` is `L` bytes (as the hash output length).

Put in RFC 2104 terms, `hmac_sha256_init` is internally assuming that the provided key is always exactly `L` bytes. However, applications may use a key of any length, and `hmac_sha256_init` only takes a pointer to the key bytes, not an array indicating the length required for the internal implementation assumptions. This leads to two bugs:

- If an application provides a pointer to a key shorter than 32 bytes, the HMAC implementation will extend the key with an out-of-bounds read. This introduces non-determinism and likely means that a different HMAC output will be produced on every invocation.
- If an application provides a pointer to a key longer than 32 bytes, the HMAC implementation will truncate the key to 32 bytes. This will deterministically produce an invalid HMAC output.

[...]

PR #2519, opened 2023-03-20
Remove hmac_sha256 from public API

nminaylov:
# What's new

- hmac_sha256 is removed from public API as it supports only 32-byte key length and is used only for U2F app

# Verification 

- Check U2F app

# Checklist (For Reviewer)


hedger:
Fixes #2512

skotopes (approved):

PR #2534, opened 2023-03-24
Improved thread lifecycle

DrZlo13:
# What's new

- The thread is only joined after the actual thread deletion

# Verification 

- Everything is affected
 
# Checklist (For Reviewer)

### DIFF
--- a/applications/services/cli/cli.c
+++ b/applications/services/cli/cli.c
@@ -37,9 +37,11 @@ char cli_getc(Cli* cli) {
     if(cli->session != NULL) {
         if(cli->session->rx((uint8_t*)&c, 1, FuriWaitForever) == 0) {
             cli_reset(cli);
+            furi_delay_tick(10);
         }
     } else {
         cli_reset(cli);
+        furi_delay_tick(10);
     }
     return c;
 }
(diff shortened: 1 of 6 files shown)


skotopes (dismissed):

skotopes (dismissed):

skotopes (approved):

skotopes:
# !!! Important notice !!! 

This PR got 2 sides:
1) Fixes issues caused by delayed TCB memory release by waiting for deallocation to happen
2) Makes our Realtime Harder([Hard Realtime](https://stackoverflow.com/questions/17308956/differences-between-hard-real-time-soft-real-time-and-firm-real-time)). `furi_thread_join` will fail(loop forever) if CPU time is not given to Idle task.

If you are here, then it was 2nd.

Here is what you need to do next:
- Do not use spin-locks, use appropriate synchronization primitives(Thread flags, Event flags, Mutex, Semaphore, Queue, etc...)
- Make sure that there are no threads in the system allowed to consume all CPU time.
- Make sure that Idle task is scheduled and executed.
- If all conditions are met but your code is still hangin on `furi_thread_join`, then open new github issue.

PR #2537, opened 2023-03-26
Allow calibrating CC1101 crystal frequency

whitequark:
# What's new

I seem to have been particularly unlucky with my device because the crystal ZQ3, specified as 10ppm, is off by 11ppm or so at room temperature. With this commit I can get the carrier frequency to be within measurement accuracy of my equipment (NESDR Nano2+, which has a TCXO with 0.5ppm accuracy).

I was tracking down a frequency error that caused replays of sub-GHz captures to be ignored by my DUT; this error is about 20% of the total frequency offset, with the other 80% being, uh, user error. By the time I figured that out, I implemented most of the functionality already, so I filed a PR. Feel free to close if this functionality is not wanted.

There is no GUI. I thought about adding another scene to Sub-GHz>Test at first but it seems pretty complicated, so for now it's just the CLI. It seems reasonable to expose this in the GUI in some way at least since if this value is set incorrectly for some reason it could send someone on a wild goose chase.

The only way to set the calibration (aside from direct filesystem manipulation) is through `subghz calibrate` CLI command.

# Verification 

- Run `subghz tx_carrier 433920000`. Measure the peak frequency.
- Run `subghz calibrate`. This returns an error since there's no calibration data yet.
- Run `subghz calibrate 100`.
- Run `subghz calibrate`. This should return 100.
- Run `subghz tx_carrier 433920000`. Measure the peak frequency. It should be ~43 kHz higher.
- Run `subghz calibrate -100`.
- Run `subghz tx_carrier 433920000`. Measure the peak frequency. It should be ~43 kHz lower.
- Run `subghz calibrate`. This should return -100.
- Run `subghz calibrate 0` to revert back to normal.

# Checklist (For Reviewer)


skotopes:
Hi. Having calibration option does make sense, but can you tell us a little bit more about your original problem?

PS: also measuring frequency with SDR is not very good idea.

whitequark:
> but can you tell us a little bit more about your original problem?

I was capturing and replaying thermostat (Salus RT510TX) control messages. It uses (as I found out later) a center frequency of 868.312 MHz, but I was capturing at 868.350 MHz (one of the presets on the Flipper). It seemed to demodulate to something reasonable, which is why I expected the replay to work, but the boiler was ignoring me, so I bought an SDR and started digging.

After adjusting the center frequency in the `.sub` file until transmissions from RT510TX and Flipper had a matching center frequency and replays started working. I ran the TX carrier test and noticed that the frequency is significantly off on the SDR spectrum. I considered that it is possible that the SDR is wrong, but the 240 MHz peak that you get around PCs that have USB 2 ports was dead on (it's measured as 240000000±100Hz in my tests). It's not a *great* frequency standard--it's just the best I had--but it's usually pretty accurate, so I thought it's probably Flipper that's off.

At this point I should have probably noticed that the (now known) center frequency of 868.312 MHz doesn't match the 868.350 MHz in the `.sub` file written by Flipper, but I missed that somehow, so I went ahead with this code change.

I think CC1101 has a tracking loop that can correct a small frequency offset for FSK, so it might be a good idea to make the sub-GHz app read FREQOFF_EST and use that, averaged, as the center frequency, or maybe notify the user if there's enough difference between requested and measured frequency. I'm not sure how well that would work in practice.

> PS: also measuring frequency with SDR is not very good idea.

Yeah, I know. Normally I would expect the SDR to be the primary source of error here, but this one has a TCXO and that seems to help.

firmware/targets/f7/furi_hal/furi_hal_subghz.c
```diff
@@ -341,7 +350,8 @@ uint32_t furi_hal_subghz_set_frequency(uint32_t value) {
     }
 
     furi_hal_spi_release(&furi_hal_spi_bus_handle_subghz);
-    return real_frequency;
+
+    return real_frequency * 1000000LL / (1000000 + furi_hal_subghz.calibration);
```
skotopes, line 354: real frequency was already corrected on previous step

firmware/targets/f7/furi_hal/furi_hal_subghz.h
```diff
@@ -178,6 +178,12 @@ float furi_hal_subghz_get_rssi();
  */
 uint8_t furi_hal_subghz_get_lqi();
 
+/** Calibrate frequency
+ *
+ * @param      value  crystal frequency offset in ppm
+ */
+void furi_hal_subghz_calibrate_crystal(int8_t value);
```
skotopes, line 185: `furi_hal_subghz_set_frequency_offset` is better

firmware/targets/f7/furi_hal/furi_hal_subghz.c
```diff
@@ -331,8 +337,11 @@ uint32_t furi_hal_subghz_set_frequency(uint32_t value) {
         furi_hal_subghz.regulation = SubGhzRegulationOnlyRx;
     }
 
+    uint32_t corrected_frequency = value * (1000000LL + furi_hal_subghz.calibration) / 1000000;
```
skotopes, line 340: this formula is incorrect

applications/main/subghz/subghz_cli.c
```diff
@@ -23,6 +23,7 @@
     "299999755...348000000 or 386999938...464000000 or 778999847...928000000"
 
 #define SUBGHZ_REGION_FILENAME "/int/.region_data"
+#define SUBGHZ_CALIBRATION_FILENAME "/int/.subghz.calibration"
```
skotopes, line 26: We'd rather to see it on ext storage in subghz assets folder, but that will require some tests first.

applications/main/subghz/subghz_cli.c
```diff
@@ -519,6 +520,7 @@ static void subghz_cli_command_print_usage() {
         printf("  debug cmd:\r\n");
         printf("\ttx_carrier <frequency:in Hz>\t - Transmit carrier\r\n");
         printf("\trx_carrier <frequency:in Hz>\t - Receive carrier\r\n");
+        printf("\tcalibrate <frequency offset:in ppm>\t - Calibrate crystal\n");
```
skotopes, line 523: Calibrate sounds like an action, but it's not. See furi_hal_subghz comment for more details.

skotopes (changes requested):

skotopes:
Un-draft when ready

firmware/targets/f7/furi_hal/furi_hal_subghz.c
whitequark: This is an inverse transformation--it's to account for `cc1101_set_frequency` doing its own frequency adjustment. It goes like: requested_frequency -(+offset)-> corrected_frequency -(cc1101_set_frequency)-> real_frequency (taking CC1101's discretization into account) -(-offset)-> returned_frequency.

To be honest I'm not sure if the returned value has a lot of utility, I wrote this to preserve the original intent as much as possible.

skotopes: Returned value is quite useful when you want to know what exactly PLL is synthesizing. I think we can keep this part, it make sense.

applications/main/subghz/subghz_cli.c
whitequark: How do you feel about `set_offset`?

firmware/targets/f7/furi_hal/furi_hal_subghz.c
whitequark: Could you elaborate, please?

skotopes: Correct. My bad, was testing with another patch applied.

skotopes:
We'll move configuration into setting_user and reimplement this feature in another PR

whitequark:
Thanks! Sorry I was busy with work...

PR #2639, opened 2023-05-05
Fix storage.py exist_dir logic

perryprog:
# What's new

- The method `exist_dir` would return true if the given path existed as either a directory or a file, which is what the plain `exist` method is for. This led to issues when trying to receive files via storage.py.

# Verification 

- Confirm that receiving a single file via `storage.py receive` does not work on dev (receiving a full directory should work fine)
- Confirm that receiving a single file via `storage.py receive` does work on this PR (receiving a full directory should also still work)

# Checklist (For Reviewer)

### DIFF
--- a/scripts/flipper/storage.py
+++ b/scripts/flipper/storage.py
@@ -323,7 +323,7 @@ class FlipperStorage:
                 return False
             raise FlipperStorageException.from_error_code(path, error_code)
 
-        return True
+        return response == b"Directory"
 
     def exist_file(self, path: str):
         """Does file exist on Flipper"""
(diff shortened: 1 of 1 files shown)


perryprog:
Not too convinced that the test failure is related but I'm also not entirely sure why it's failing. It's definitely not impossible it's due to this change, but I'm not seeing why it could be.

Only thing that it seems like it might be is a failure in storage mkpath but I can't see why that'd be called at that point in the workflow file, and mkpath works fine for me in local testing.

perryprog:
I totally take that back; it was an edge cased I missed. `storage stat` has a different output when on `/ext` or `/int`, and `recursive_send` used isdir on `/ext` which was incorrectly returning false due to my change. Should be good now.

hedger (approved):

PR #2647, opened 2023-05-07
0 Byte PCAP Files

ghost:
Hello,

using Flipper Zero with unleashed 0.48 along with Marauder Companion App 0.34, I have updated the Devboard to latest version "esp32_marauder_v0_10_3_20230414_flipper.bin". 

The App works fine, great job! The scanlogs and all activity are written in logs folder. The Sniff function creates the PCAP Files but they are always 0 byte. All the sniffing creates files in the correct folder, but empty ones.

Maybe can you help?


hedger:
We don't provide support for 3rd-party applications. Ask its author for assistance.

Also you opened a pull request, that's not how you should ask for help anyway.

PR #2688, opened 2023-05-22
nfc: Topaz read support

GMMan:
Note: this is mostly intended for reference and accessibility for those who want this right now. Not anticipating this would be merged before the NFC refactor, and I'll make a new PR after that's done.

# What's new

- Topaz 96 and 512 read support
  - Topaz 96 [datasheet](https://nfc.cards/en/index.php?controller=attachment&id_attachment=17)
  - Topaz 512 [datasheet](http://www.enjoyrfid.com/PDF/BCM20203T512%20wafer%20ds.pdf)

# Verification 

- Use NFC read on Topaz 96 or 512 tag. It should be able to read the tag and show its 7-byte UID, HR0, and HR1 values.
- Scanned tag should be saveable
- Scanned tag should be loadable once saved
- Scanned tag should have HR0 and HR1 showing in Info scene, and should show hex dump when "More" is selected on Info scene
- Topaz should be selectable and usable from "Read Specific Card Type" menu
- All other NFC tag types should work the same as before (regression check)

# Checklist (For Reviewer)

- [ ] PR has description of feature/bug or link to Confluence/Jira task
- [ ] Description contains actions to verify feature/bugfix
- [ ] I've built this code, uploaded it to the device and verified feature/bugfix

PR #2694, opened 2023-05-24
NFC: Skip RATS if not needed

AloneLiberty:
Part of code taken (and improved) from pending PR #2688, but doesn't depend on it

# What's new

- Skip RATS: Mifare Classic (SmartMX Classic/Plus in SL1) can't be read after 14443-4 activation (RATS). This PR removes mandatory sending of RATS, therefore now you can read Mifare Classic via **Read Specific Card Type** -> **Read Mifare Classic**

# Verification 

All read should be done via **Read Specific Card Type** -> **Read Mifare Classic**, should read Mifare Classic

- Read SmartMX with Classic emulation (SAK 20)
- Read Mifare Plus in SL1 on EMV card (again SmartMX?)
- If an EMV application is planned before the NFC refactor: check EMV reading

# Checklist (For Reviewer)


gornekich:
Hello @AloneLiberty 
The idea is good. The problem I can see is that we can't properly identify Mifare Classic type which is iso-14443-4a complient (like Mifare Classic Plus 2k / 4k). We planned to add support for these cards during NFC refactoring.

I convert this PR to draft for now.

AloneLiberty:
> We planned to add support for these cards during NFC refactoring.

Main point of this PR is hot-patch this issue before NFC refactor, as it won't be issue after.

gornekich:
I got your point. But I am afraid that this would raise more questions. When I read Mifare Classic Plus 4k, it is identified and Mifare Classic 1k and not all sectors are read.

I prefer to add full support of these type of cards later, then partially read them and save as incorrect types now

PR #2756, opened 2023-06-09
fix: make `dialog_file_browser_set_basic_options` initialize all fields

JarvisCraft:
# What's new

Previously `dialogs_file_browser_options` did not initialize `hide_dot_files` which led to undefined behaviour in most of its use-cases since it was called on an unitialized structs leaving them in uninitialied state.
This PR fixes this by adding initialization for the remaining field.
This also updates the documentation to ensure that `dialogs_file_browser_options` should initialize all fields and adds a unit-test to assert this behaviour.

# Verification 

- Unit tests should be run, i.e. `dialogs_file_browser_options` suite.

# Checklist (For Reviewer)


skotopes (dismissed):

JarvisCraft:
The test seems to fail, I will take a look.

applications/debug/unit_tests/dialogs/dialogs_file_browser_options.c
```diff
@@ -0,0 +1,25 @@
+#include <dialogs/dialogs.h>
+
+#include "../minunit.h"
+
+MU_TEST(test_dialog_file_browser_set_basic_options_should_init_all_fields) {
```
skotopes, line 5: This test is incorrect: DialogsFileBrowserOptions structure is not packed and filling it with `n` is not going to work

skotopes (changes requested):

applications/debug/unit_tests/dialogs/dialogs_file_browser_options.c
JarvisCraft: Hmm, I forgot about alignment. In this case, I am not sure what the right way to test this (allowing automatic updates in case of futue changes) would be.

JarvisCraft: I've got a small idea to utilize minimal test to ensure some compatibility.

JarvisCraft: @skotopes, here is the alternative approach to the test. It may not handle some specific corner-cases but it is better than nothing.

It adds an assertion on structure size (guess I calculated it correctly) which would get triggered on most changes to `DialogsFileBrowserOptions` (except for ones which fit into alignment) forcing the developer to check the test and thus update initialization logic.

JarvisCraft:
@skotopes, could you please take a look at the changes and re-run the CI?

skotopes (approved):

PR #2201, opened 2022-12-27
picopass read bug fixes

nvx:
# What's new

- fix checking user elite keys - current behaviour checks only the default dictionary
- include calculated Kd when saving cards - current behaviour sets the Kd block to FFFFFFFFFFFFFFFF in the save file (this is an important pre-requisite for emulation which is coming soon)

# Verification 

- Attempt to read an elite keyed iClass card with a key not in the default (`picopass/assets/iclass_elite_dict.txt`) dictionary
- Observe the read fails
- Add the key to the user dictionary (`picopass/assets/iclass_elite_dict_user.txt`)
- Try to read the card again
- Observe the read succeeds
- Save the read card
- Observe the card-specific derived key in block 3 in the saved file rather than all FF's

# Checklist (For Reviewer)

- [x] PR has description of feature/bug or link to Confluence/Jira task
- [x] Description contains actions to verify feature/bugfix
- [x] I've built this code, uploaded it to the device and verified feature/bugfix

PR #2759, opened 2023-06-10
add picopass emulation

nvx:
# What's new

- Adds emulation support to the picopass app
- Adds reader nr/mac collection for loclass attack

# Verification 

- Read an iClass card and save it
- Open the card and select emulate
- Present the flipper to a HID reader that the original card worked on and the reader should act the same

# Checklist (For Reviewer)


nvx:
I've left this as a draft for the moment as there's a a couple of jank issues that need fixing still, but I figured I'd get the PR opened so reviews of the rest of the code can start.

* Exiting emulation is a bit glitchy, sometimes it hangs for a little bit. The flashing stops immediately so it's not an issue capturing the input but I think I've mucked something up with the UI cleanup code. I'm not great Flipper UI code so would be happy if someone can spot what I've done wrong here.
* There should probably be a "capture complete" popup after successfully collecting all the nr/mac pairs for loclass rather than just unceremoneously dumping back to the main menu.

Ping: @bettse @pcunning

Also a huge thanks goes to @g3gg0 for the work on ISO15693 emulation on which this PR heavily relies upon!

nvx:
I haven't tested non-secure card emulation, but in theory it should work. There's a few TODOs for things like checking CRC's and checking auth before allowing reading of blocks that hasn't been implemented which means non-secure card emulation should work. In practice an ideal implementation would be checking these things (along with checking the config block flags to see if it's a secure card or not). From memory though the picopass read code doesn't handle non-secure cards yet anyway though.

I know of at least one example of a gym in France that uses non-secure picopass cards, so there is some utility in eventually supporting non-secure reads. Might start looking at that once this is done and dusted.

nvx:
Note sufficiently old read cards may have all "FF" as the keys (this was fixed in PR #2201). Attempting to emulate with such saved cards will not be able to be read by readers. Re-read the card using a newer version that includes the aforementioned fix and it should work though.

nvx:
The loclass support can be tested by running against an elite keyed reader, then opening `sdcard/apps_data/picopass/.loclass.log` with a text editor and extracting the hex bytes (ignore ts and no fields, those are for informational purposes) once for each CSN (we do each CSN twice to allow for keyroll modes that alternates keys every second read, so you should either do the first of each CSN, or if that fails to reccover the second instance of each CSN) then save those hex bytes as a binary file which can be fed into `hf iclass loclass` in the Proxmark3 client to recover the key.

@bettse has been doing some work in this space to make it easier, but I'll leave it to him to talk more about that (note I added versioning to the log lines so your PoC will need updating).

Unlike other loclass implementations, since the MCU is relatively powerful it specifically checks nr/mac pairs for standard keys and if any are encountered will display on the screen that standard keys have been detected and will ignore them. Note this only detects legacy standard keys, I believe a HID reader configured with Standard-2 keysets (which effectively disables legacy keys) this will not be able to detect such keys and will log as per normal, but then fail to recover elite keys during the offline portion.

applications/external/picopass/picopass_device.h
```diff
@@ -24,6 +25,49 @@
...
+
+// PicoPass command bytes:
+// Low nibble used for command
+// High nibble used for options and checksum (MSB)
+// The only option we care about in 15693 mode is the key
+// which is only used by READCHECK, so for simplicity we
+// don't bother breaking down the command and flags into parts
+// READ: ADDRESS(1) CRC16(2) -> DATA(8) CRC16(2)
+// IDENTIFY: No args -> ASNB(8) CRC16(2)
+#define PICOPASS_CMD_READ_OR_IDENTIFY 0x0C
```
bettse: I think this is the same as `RFAL_PICOPASS_CMD_READ` in `rfal_picopass.h`, except yours has a better explanation.  If they are the same, feels like we shouldn't duplicate, although I'm up in the air as to the 'better' file to put them in.  What do you think?

applications/external/picopass/picopass_i.h
```diff
@@ -22,13 +22,18 @@
...
 #include "views/dict_attack.h"
+#include "views/loclass.h"
 
 #include <storage/storage.h>
 #include <lib/toolbox/path.h>
 #include <picopass_icons.h>
 
 #define PICOPASS_TEXT_STORE_SIZE 128
 
+#define LOCLASS_NUM_CSNS 9
```
bettse, line 33: I was asking myself if this should be in some sort of "loclass" header, but I realize we use the term "loclass" for both the library as well as the attack, so I think this is a good place for the attack specific `define`s since there aren't many

applications/external/picopass/scenes/picopass_scene_emulate.c
```diff
@@ -0,0 +1,58 @@
+#include "../picopass_i.h"
+#include <dolphin/dolphin.h>
+
+void picopass_emulate_worker_callback(PicopassWorkerEvent event, void* context) {
```
bettse, line 4: Look at you, doing GUI code!

applications/external/picopass/scenes/picopass_scene_saved_menu.c
```diff
@@ -26,6 +27,12 @@ void picopass_scene_saved_menu_on_enter(void* context) {
         submenu, "Info", SubmenuIndexInfo, picopass_scene_saved_menu_submenu_callback, picopass);
     submenu_add_item(
         submenu, "Write", SubmenuIndexWrite, picopass_scene_saved_menu_submenu_callback, picopass);
+    submenu_add_item(
+        submenu,
+        "Emulate",
```
bettse, line 32: 🤤

bettse (commented):
I don't claim to understand the emulation, but the parts I do understand look good.  Very excited.

bettse:
> There should probably be a "capture complete" popup after successfully collecting all the nr/mac pairs for loclass rather than just unceremoneously dumping back to the main menu.

I agree, but don't let perfect be the enemy of good.  I think it is amazing it works, and we can polish UX in a future PR.

> I haven't tested non-secure card emulation

picopass, despite the name, has been highly focused in iClass, so I wouldn't stress about non-iClass

> @bettse has been doing some work in this space to make it easier, but I'll leave it to him to talk more about that (note I added versioning to the log lines so your PoC will need updating).

Yup, I've got some code at https://loclass.ericbetts.dev/ , but it needs examples to validate and whatnot.

> I believe a HID reader configured with Standard-2 keysets

I'm not familiar with Standard-2, I'll have to look that up

nvx:
> > There should probably be a "capture complete" popup after successfully collecting all the nr/mac pairs for loclass rather than just unceremoneously dumping back to the main menu.
> 
> I agree, but don't let perfect be the enemy of good. I think it is amazing it works, and we can polish UX in a future PR.

100% agreed, hence submitting this now, but also calling it out if someone gets inspired to throw in a patch for such a thing too. :D

The hang when backing out from emulation is the main reason I've left this as a draft for the second. Oddly I don't recall this happening when I first wrote it though so I'm not sure if it's due to flipper changes, or changes in the 15693 PR code that's biting me.

applications/external/picopass/picopass_device.h
nvx: Herp I missed the ones in rfal_picopass.h. Agreed it's worth collapsing them into one. Will fix this up soon.

nvx: Merged them together, also cleaned up some other duplicated defined (noticed the block size defined in two different places as well)

pcunning:
Just did a successful loclass with an elite keyed Multiclass SE reader and https://loclass.ericbetts.dev. Checking for the standard key is very handy (especially thinking about how people use "detect reader").

Nice work!

applications/external/picopass/picopass_worker.c
```diff
@@ -749,3 +787,484 @@ void picopass_worker_write_key(PicopassWorker* picopass_worker) {
...
+                } else if((loclass_mode) && !loclass_got_std_key && emu_ctx.loclass_got_std_key) {
+                    loclass_got_std_key = true;
+                    picopass_worker->callback(
+                        PicopassWorkerEventLoclassGotStandardKey, picopass_worker->context);
+                } else {
+                    picopass_worker->callback(
+                        PicopassWorkerEventSuccess, picopass_worker->context);
+                }
+            }
+        }
```
bettse, line 1260: Most of the other `while(picopass_worker->state ==` end with a `furi_delay_ms(100);`: Could be the problem you're seeing with leaving emulation.

applications/external/picopass/picopass_worker.c
```diff
@@ -749,3 +787,484 @@ void picopass_worker_write_key(PicopassWorker* picopass_worker) {
...
+                    last_loclass_csn_num = emu_ctx.key_block_num;
+                    picopass_worker->callback(
+                        PicopassWorkerEventLoclassGotMac, picopass_worker->context);
+                } else if((loclass_mode) && !loclass_got_std_key && emu_ctx.loclass_got_std_key) {
+                    loclass_got_std_key = true;
+                    picopass_worker->callback(
+                        PicopassWorkerEventLoclassGotStandardKey, picopass_worker->context);
+                } else {
+                    picopass_worker->callback(
+                        PicopassWorkerEventSuccess, picopass_worker->context);
```
bettse, line 1257: In the other similar blocks they'll have a "break" after something like this.  I don't fully understand the emulation code well enough to say we should have one, but I just wanted to point it out in case it was missed.

applications/external/picopass/picopass_worker.c
nvx: Funnily enough this sleep caused compatibility issues in both picopass and 15693 emulation (for the latter it broke seemingly all Android phones from being able to read with a much smaller delay). Maybe sleeping a few microseconds wouldn't impact it though? Could experiment to see if it helps the responsiveness at any rate.

skotopes:
Picopass was moved to https://github.com/flipperdevices/flipperzero-good-faps closing this PR

PR #2773, opened 2023-06-13
furi_hal_nfc: fix furi_hal_nfc_emulate_nfca

leommxj:
# What's new

call to `rfalTransceiveBitsBlockingTx` in `furi_hal_nfc_emulate_nfca` are passing byte count for buff `buff_rx`, which should be bits count as `rfalTransceiveContext` are expected. 
This will cause `furi_hal_nfc_emulate_nfca` can only handle max 256 bits of data, more data will raise an ERR_NOMEM from `rfalTransceiveBlockingRx`. Only 256 bits of data in `buff_rx` but `buff_rx_len` is the right(bigger) one.

this fixes the problem.

# Verification 

- `furi_hal_nfc_emulate_nfca` can handle data of more than 256 bits correctly.

# Checklist (For Reviewer)

### DIFF
--- a/firmware/targets/f7/furi_hal/furi_hal_nfc.c
+++ b/firmware/targets/f7/furi_hal/furi_hal_nfc.c
@@ -467,7 +467,7 @@ bool furi_hal_nfc_emulate_nfca(
                     buff_tx,
                     buff_tx_len,
                     buff_rx,
-                    sizeof(buff_rx),
+                    rfalConvBytesToBits(buff_rx_size),
                     &buff_rx_len,
                     data_type,
                     RFAL_FWT_NONE);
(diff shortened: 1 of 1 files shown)


gornekich (approved):
Thanks for the fix!

Did you have a problem with incorrect emulation when flipper receives more than 256 bits?

leommxj:
> Thanks for the fix!
> 
> Did you have a problem with incorrect emulation when flipper receives more than 256 bits?

indeed, while testing an NFC relay app I wrote, the flipper acts as a card that will have 256bits real data, and remain parts are all uninitialized `0x5A`. 😅

skotopes (approved):

PR #2774, opened 2023-06-13
[FL-3368] NFC Magic: Change the UI to match the planned one

Astrrra:
# What's new

- NFC magic UI/UX is now matching the one that we planned

# Verification 

- Try writing different dumps to gen4 and gen1 cards, verify that they get written/rejected correctly
- Try writing anything on a non-magic card, you should receive an error saying that it's not a magic card.
- Try writing to a password-protected gen4 card using the special menu [IMPORTANT! My gen4 card is broken and doesn't change its password, I couldn't verify this myself]
- Try writing to a non-password-protected gen4 card using the special menu, you should not get the password input screen (unless the card has bad signal or just appeared in the field, you can't really differentiate between password-protected cards and bad connections, unfortunately)
- Try writing to a non-magic card using the gen4 menu, you should receive an error after inputting any non-standard password
- Try wiping all card types (gen1, unlocked gen4, locked gen4)
- Try changing gen4 password (though some cards apparently can't do that, you can double-check with a proxmark)

# Checklist (For Reviewer)


Astrrra:
Feedback and validation from @AloneLiberty for this would be greatly appreciated!

Relevant graphs: 
image
image

skotopes:
I think we need to discuss it with community and our UI/UX (@Karator). Community version do looks more logic than what we got.

AkechiShiro:
@skotopes could you share a link to how it looks like ? or the repo of the community firmware/application where there is such an UX/UI ?

skotopes:
@AkechiShiro there is a Miro board where we gather everything related to flipper UI/UX. 

Meanwhile I'll close this PR, we decided to leave current UI/UX.

PR #2783, opened 2023-06-19
SLIX2 emulation support / practical use for Dymo printers

g3gg0:
# What's new

## SLIX2 specific
- added SLIX2 specific feature emulation like providing NXP's signature (issue #2781)
- added SLIX2 specific feature reading
## SLIX family
- allow unset keys (also requested in issue #2781)
- if a key is not set, the first password seen will get accepted
- the first accepted password will get stored and from now on the password is set
- clean up duplicate SLIX code by using generic functions
## general
- improve digital_signal performance to send longer packets (impact on picopass @nvx ?)
- clean up digital_signal code
- display card info (blocks/size) as integer, not hexadecimal

# not implemented
- does not read the security bits (yet?)

# Checklist (For Reviewer)


g3gg0:
@nvx can you see some impact on picopass emulation with this patchset?

g3gg0:
current state:
it should read a SLIX2 and be able to emulate it.
all information neeeded will be read, *except* the security status bits for every block. (is this even required?)

@gornekich @nvx ISO15693-3 10.4.1 states that if Option_flag is set, the VICC responds with the block security status.
do you think it is okay to change nfcv_read_blocks() to always add this flag and handle the status byte, if it was returned?
a) i don't want to make an extra READ BLOCK (option 1) readout function just for reading the status byte
b) using "10.4.13 Get multiple block security status" commands would just work with SLIX2, as SLIX-L only supports READ BLOCK (option 1)
c) not implementing it feels wrong :)

nvx:
> * improve digital_signal performance to send longer packets (impact on picopass @nvx ?)

This may actually fix a bug I was chasing. Had some weirdness preventing OmniKeys from reading an emulated picopass. The trace was odd though, during the tags responce to a READ4 command (reads 4 blocks of 8 bytes, plus 2 byte CRC16, so 36 bytes total response) the reader seemed to start talking again before the tag finished replying - which I since discovered was a bug in the Proxmark3 code (fixed in https://github.com/RfidResearchGroup/proxmark3/pull/2010) but hadn't gotten around to circling back to the initial problem I was troubleshooting.

If the current code assumed replies no longer than 32, this could certainly explain the weird behaviour I was seeing and with any luck this PR should fix it. When I get some time might retest with this PR's changes incorporated to see if it makes OmniKeys happier.

nvx:
> ISO15693-3 10.4.1 states that if Option_flag is set, the VICC responds with the block security status. do you think it is okay to change nfcv_read_blocks() to always add this flag and handle the status byte, if it was returned?

Do all tags support this? If not maybe a add a bool to the function definition to control the behaviour?

g3gg0:
> > ISO15693-3 10.4.1 states that if Option_flag is set, the VICC responds with the block security status. do you think it is okay to change nfcv_read_blocks() to always add this flag and handle the status byte, if it was returned?
> 
> Do all tags support this? If not maybe a add a bool to the function definition to control the behaviour?

to my understanding, if not supported the card responds memory content without security byte.
at least "not mentioning" option_flag (previously custom flag) in the '99 standard suggests so.
do currently have no NfcV card except SLIX ones

g3gg0:
> This may actually fix a bug I was chasing.

had chance to try this patch?

skotopes (dismissed):

skotopes (approved):

g3gg0:
Amazing! Thanks!

PR #2801, opened 2023-06-25
Game deeds update

sfjuocekr:
- I changed the DolphinDeedPluginGameStart to DolphinDeedPluginStart because having multiple start "rewards" with the same value is unnecessary.
- Also added win and loss conditions. with reduced XP awarded to them.
- Updated the snake game to reflect the changes, I think we should add DolphinDeedGameWin and DolphinDeedGameLost to it as well in snake_game_process_game_step?

### DIFF
--- a/applications/external/snake_game/snake_game.c
+++ b/applications/external/snake_game/snake_game.c
@@ -346,7 +346,7 @@ int32_t snake_game_app(void* p) {
 
     notification_message_block(notification, &sequence_display_backlight_enforce_on);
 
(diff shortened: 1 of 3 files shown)


skotopes:
Some magic happens here, I didn't close this PR, but it's closed and i can not reopen it.

skotopes:
@sfjuocekr I see that your branch is based on other fork. In that case update from our repo is impossible. Looks like you need to recreate this pr from correct branch and make new PR

Issue #2577, opened 2023-04-12
NFC: Emulating of SAVED Mifare Classic not working


AlexMilender:
### Describe the bug.

Emulation of Mifare Classic produces different results depending on if the Mifare Classic fob has been saved or not.

I am able to successfully read a Mifare Classic key fob using NFC read functionality.
If I do NOT save the read, but instead emulate it directly after the read. The emulation works as expected and i am able to open my key fob lock

If I DO save the read and then try to emulate that saved read. The emulation does not work as expected and I am unable to open my key fob lock

### Reproduction

Successful able to open keyfob lock
1. NFC
2. Read
3. More
4. Emulate

Unsuccessful in opening keyfob lock
1. NFC
2. Read
3. More
4. Save the read
5. NFC
6. Saved
7. Naviatege to saved read from step 4
8. Emulate

### Target

Flipper Zero Release 0.80.1

bettse:
What brand is the lock?  I'd like to see if this is a problem for _all_ emulation, or against a specific lock.

doomwastaken:
Going to research it now on 0.81.1-rc

doomwastaken:
If you could provide additional details about lock and what card you are using, that would be helpful, tried Mifare Classic 1k and 4k and was not succesful so far

Astrrra:
Also, having the full card dump will help out a lot

Indigo94:
I’m experiencing the same issue as OP. The reader is a Schlage lock

Panduhsaur:
I have the same issue as OP.

Here's the Schlage model: https://www.schlage.com/en/home/products/BE467GRWFFF.html?fbclid=IwAR29B_GfcXJUn_cKD3ezzMMTAhIU_PCkSxDqYFbRcEwzHy6_i11rw30EsHI

gornekich:
Could you please try to reproduce the issue on previous release

AlexMilender:
> What brand is the lock? I'd like to see if this is a problem for _all_ emulation, or against a specific lock.
@doomwastaken 

The lock brand is Schlage. The same one @Panduhsaur mentioned (https://www.schlage.com/en/home/products/BE467GRWFFF.html)

Key fob is a Schlage 9651t

Indigo94:
> Could you please try to reproduce the issue on previous release

[...]

PR #2825, opened 2023-07-01
NFC: Improved MFC emulation on some readers

AloneLiberty:
# What's new

- HALT don't disable field (because in the real world the card cannot turn itself off)

- Fixes for faster re-select of card (because we know that in these cases the reader will read again)

- Removed not particularly necessary warning output, in general improves the emulation (`hf mf fchk`, Android with different keys)

- Add trace log if reader send RATS

- Added check if the block is read, if not, it won't access it (won't reply zeros)

- Added check if we access a block we're authorized (don't allow access outside current sector)

# Verification 

- `nfc-mfclassic r a u dump` on ACR122U should work now

- Emulation should work on more Android devices (at least now works on mine)

- `hf mf chk`, `hf mf fchk` and even `hf mf hardnested` proxmark3 commands works stable

- Test for regression on other readers

# Checklist (For Reviewer)


giacomoferretti:
Works for me!
Tested on ACR122U, gonna test it soon with MCT on Android.

Should fix #1862 and maybe even #2577.

lib/nfc/protocols/mifare_classic.c
```diff
@@ -882,12 +882,17 @@ bool mf_classic_emulator(
             break;
         }
 
-        if(cmd == 0x50 && plain_data[1] == 0x00) {
+        if(cmd == NFCA_CMD_HALT && plain_data[1] == 0x00) {
             FURI_LOG_T(TAG, "Halt received");
-            furi_hal_nfc_listen_sleep();
-            command_processed = true;
+            return false;
```
gornekich: Please, avoid return from the body of function. It's not safe if you allocate memory above return and exit from function before freeing that memory. Also it make code less readable.

I suggest rename variable `command_processed` -> `send_nack` and add new variable `need_reset`. Rework every place where you write `return` with filling these variables then call `break`. And return `need_reset` at the and of the function

lib/nfc/protocols/mifare_classic.c
```diff
@@ -902,8 +907,7 @@ bool mf_classic_emulator(
                     access_key = MfClassicKeyA;
                 } else {
                     FURI_LOG_D(TAG, "Key not known");
-                    command_processed = true;
-                    break;
+                    return false;
```
gornekich: Same here and below

gornekich (changes requested):
Hello @AloneLiberty ! Thanks for PR. Tested it and it works good.

Please, make some changes I mentioned and we will merge your PR

gornekich (approved):
Thanks

skotopes (approved):

Issue #2834, opened 2023-07-03
Picopass app key change to change iCE to elite KDF instead of standard KDF


Akury83:
### Description of the feature you're suggesting.

Currently, the Picopass app has a key change feature where the app changes the key on the card.

On the iCE selection, the key is changed to the same key as the iCL key, both using the standard KDF, basically the same raw key is written in either case.

The DRM for the iCE key is the same key as iCL, however it is meant to be encoded using the elite KDF.

Could I please request that the iCE selection uses the elite KDF to change the key when selected?

### Anything else?

As an example, for CSN 2069E013FEFF12E0, the data written in either case when selecting iCE or iCL is 10D7B7245E363973, but should actually be 06E1C3879DBB030E when iCE is selected.

Thank you for all of your hard work on coding this everyone, @bettse in particular. I really appreciate it!!

bettse:
ohhhh.  I kept wondering why the keys were the same.  I was following https://github.com/iCopy-X-Community/icopyx-teardown/blob/master/tags/README.md#ice and I see the `--elite` for ICE that isn't present with the others.  
This should be fixable, but isn't trivial.  I'll look into it when I have some time.

bettse:
@Akury83 Give https://github.com/flipperdevices/flipperzero-firmware/pull/2836 a try and see if that fixes it

Akury83:
All sorted, thank you @bettse, you're a legend!

PR #2836, opened 2023-07-04
Picopass fix ice

bettse:
# What's new

- Fixing issue from https://github.com/flipperdevices/flipperzero-firmware/issues/2834

# Verification 

- Read a card
- Choose "Change key"
- Write iCE 
- Check key is diversified using elite KDF or Use with iCopyX as iCE card

# Checklist (For Reviewer)


Akury83:
That worked @bettse, thank you for your help! There is one issue now that I've noticed but am unsure if it's related or not:

The newly keyed card can't be read via the normal 'Read Card' nor the 'Elite Dict. Attack' method. The key is present in both the `iclass_elite_dict.txt` and `iclass_elite_dict_user.txt` files.

Update: reverted back to old app, does the same thing. I guess this is outside the scope of this feature request, which now works. Thank you kindly!

bettse:
did you update the `iclass_elite_dict.txt` based on this PR?  an iCE keyed card should be read by the Elite Dict. Attack, if it isn't, that seems like a problem.

Akury83:
I didn't, I manually added the key into both `iclass_elite_dict.txt` and `iclass_elite_dict_user.txt` files on the flipper.

Really unusual - if I copy the `iclass_elite_dict.txt`  from the pull request, it seems to work whereas if I manually edit the file myself to also include the key in there per the above, it doesn't seem to work... Anyway, works as expected now, thank you :)

skotopes (approved):

PR #2850, opened 2023-07-06
[FL-3420] Storage: directory sort

DrZlo13:
# What's new

- Storage dir read now returns files in alphabetical order if possible with memory limits.

# Verification 

- Everything is affected

# Checklist (For Reviewer)


applications/services/storage/storage_processing.c
```diff
@@ -260,9 +261,149 @@ static bool storage_process_file_eof(Storage* app, File* file) {
...
+static int storage_sorted_file_record_compare(const void* sorted_a, const void* sorted_b) {
+    SortedFileRecord* a = (SortedFileRecord*)sorted_a;
+    SortedFileRecord* b = (SortedFileRecord*)sorted_b;
+
+    if(a->info.flags & FSF_DIRECTORY && !(b->info.flags & FSF_DIRECTORY))
+        return -1;
+    else if(!(a->info.flags & FSF_DIRECTORY) && b->info.flags & FSF_DIRECTORY)
+        return 1;
+    else
+        return furi_string_cmp(a->name, b->name);
```
digitalcircuit: Would it be better to use `furi_string_cmpi` for a case-insensitive sorting, given FAT is a case-insensitive filesystem?  [`fatsort`](https://fatsort.sourceforge.io/ ) does that by default.

I also use [the "natural order sorting" option](https://en.wikipedia.org/wiki/Natural_sort_order ) with `fatsort`, for example:
* Alphabetical
  1. `z11`
  2. `z2`
* Natural
  1. `z2`
  2. `z11`

I think it's a better result, but I'd understand if that's more than you'd want to handle in this pull request.

digitalcircuit (commented):
Awesome!  This should mean I no longer have to manually run `fatsort` on my Linux laptop after updating the SD card.

applications/services/storage/storage_processing.c
DrZlo13: Yes, case-sensitive sorting makes sense. But sorting in natural order is out of the scope of this pull request, I'll leave that for future improvements.

DrZlo13:
The Moor has done his work, the Moor may go.

skotopes (approved):

PR #2883, opened 2023-07-16
feat: add API for fallible `ViewDispatcher` View operations

JarvisCraft:
# What's new

Fallible APIs (prefixed with `try_`) have been added to `ViewDispatcher` so that it can perform addition/usage/removal of `Views` permitting the operations to fail (depending on the contents of the `ViewSet`).

This API is important for FFI since it allows foreign callers (namely, ust bindings) to provide safe abstractions over `ViewDispatcher` without a need to duplicate IDs tracking on their side.

# Verification 

- `ViewDispatcher`-related tests and code should be re-run

# Checklist (For Reviewer)

### DIFF
--- a/applications/services/gui/view_dispatcher.h
+++ b/applications/services/gui/view_dispatcher.h
@@ -123,12 +123,29 @@ void view_dispatcher_stop(ViewDispatcher* view_dispatcher);
  */
 void view_dispatcher_add_view(ViewDispatcher* view_dispatcher, uint32_t view_id, View* view);
 
+/** Try to add view to ViewDispatcher
+ *
+ * @param      view_dispatcher  ViewDispatcher instance
+ * @param      view_id          View id to register
+ * @param      view             View instance
+ * @return     {@code true} if there is no view by this ID and {@code false} otherwise
+ */
+bool view_dispatcher_try_add_view(ViewDispatcher* view_dispatcher, uint32_t view_id, View* view);
+
 /** Remove view from ViewDispatcher
  *
  * @param      view_dispatcher  ViewDispatcher instance
  * @param      view_id          View id to remove
  */
 void view_dispatcher_remove_view(ViewDispatcher* view_dispatcher, uint32_t view_id);
+
+/** Try to remove view from ViewDispatcher
+ *
+ * @param      view_dispatcher  ViewDispatcher instance
+ * @param      view_id          View id to remove
+*  @return     {@code true} if there was a view by this ID and {@code false} otherwise
+ */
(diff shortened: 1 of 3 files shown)


applications/services/gui/view_dispatcher.h
```diff
@@ -123,13 +123,30 @@ void view_dispatcher_stop(ViewDispatcher* view_dispatcher);
...
 void view_dispatcher_add_view(ViewDispatcher* view_dispatcher, uint32_t view_id, View* view);
 
+/** Try to add view to ViewDispatcher
+ *
+ * @param      view_dispatcher  ViewDispatcher instance
+ * @param      view_id          View id to register
+ * @param      view             View instance
+ * @return     {@code true} if there is no view by this ID and {@code false} otherwise
+ */
+bool view_dispatcher_try_add_view(ViewDispatcher* view_dispatcher, uint32_t view_id, View* view);
```
skotopes, line 133: Hi @JarvisCraft to be honest we don't use `try` pattern. But we'll accept something like `_view_exists()`.

skotopes (changes requested):

skotopes:
Undraft when ready

applications/services/gui/view_dispatcher.h
JarvisCraft: I like the idea of adding `_view_exists`, but I think that having `_try` methods for these specific methods are also essential since theyt allow the caller to not perform value lookup twice (once on check and again on call to the corresponding method).

I am not suggesting adding `_try`-equivalents for each and every use-case, but IMO it is appropriate here since the `check` relies on a "growing" internal state.

skotopes: Can I see real life usage example?

JarvisCraft: Currently, the obvious implementation would look like the following:

image

This, however. has a problem that the ID may already be used which would lead to an error.

---

As a currently available alternative, we could duplicate this IDs tracking on our side:

image

But this would mean having to duplicate data in some heap-allocated struct on our side (also having to keep it in sync for other methods).

---

With the proposed solution, the fixed version would look like:

image

---

The solution based on `_view_exists()` would also be sound:

image

but it would have unreasonable duplicate check.

---

PS: I use result to make the code look canonical, but it could also be `bool` or whatever is more appropriate. The core factor here is that we don't want the API users to observe random crash at firmware level (or UB if assertions are disabled) instead of explicitly having an error indicator (not UB) at our API level

skotopes:
Summary:
- `try` are evil and not aligns with existing stuff, not going to be accepted
- but if some `_exists()` methods are missing - we can add them

PR #2886, opened 2023-07-17
Full MFC emulation , Detect Reader

gornekich:
# What's new

- [ Describe changes here ]

# Verification 

# Checklist (For Reviewer)


AloneLiberty:
@gornekich I don't know where should I correctly report this, but nested authorization doesn't work there. After decryption buffer size is 4 * 8, but both auth commands accept only 2 * 8. You can reproduce this via Mifare Classic Tool on Android, it will read only odd sectors (even sectors will try nested auth and fail with NACK)

gornekich:
@AloneLiberty thanks a lot, will fix it soon

PR #2960, opened 2023-08-08
[FL-3421] Unit bench: multiple attempts to find flipper

doomwastaken:
# What's new

Unit bench: added retry count for finding flipper for reboot/reboot2dfu and power off scripts. Changed returns to sys.exit() since scripts are being used standalone

# Verification 

Unit bench runs

# Checklist (For Reviewer)


scripts/power.py
```diff
@@ -27,41 +31,53 @@ def init(self):
...
 
         flipper = FlipperStorage(port)
         flipper.start()
         return flipper
 
     def power_off(self):
-        if not (flipper := self._get_flipper()):
-            return 1
+        if not (flipper := self._get_flipper(retry_count=10)):
+            sys.exit(1)
```
skotopes: Why? Return is doing exactly the same but gives upper level ability to process it

skotopes (changes requested):

skotopes (approved):

PR #3037, opened 2023-09-03
Add extended I2C HAL functions

agarof:
# What's new

* Add two functions `furi_hal_i2c_tx_ext` and `furi_hal_i2c_rx_ext`, giving the user more flexibility in writing I2C drivers by giving them more choices such as sending RESTART conditions and using multiple buffers in the same transaction.
* Enable transfers larger than 255 bytes by automatically using the reload flag, The flag has been added to existing functions and shouldn't be an API breaking change.
* Make the timeout for multi-transactions functions such as `furi_hal_i2c_trx` shared between all transactions instead of being duplicated.
* Fix an issue where we wait for the timeout to expire when a device is not present on the bus in both the RX/TX functions and `furi_hal_i2c_is_device_ready`.
* Add ten bit addresses support (untested).

I tried to keep the binary size increase to a minimum but this still adds about 300 bytes to the final binary (304 in debug, 296 in compact).

# Verification

To check for basic functionality, simply boot the Flipper. Non working code will cause the communication with the battery control chips and the LED driver to fail, resulting in missing battery readouts and no screen backlight.
More advanced verification must be done using a logic analyzer.
As I have not opened my Flipper, I have only checked the behavior using the external bus.

# Checklist (For Reviewer)


firmware/targets/f7/furi_hal/furi_hal_i2c.c
```diff
@@ -51,177 +51,248 @@ void furi_hal_i2c_release(FuriHalI2cBusHandle* handle) {
...
+
+static uint32_t get_start_signal(I2CBegin begin, bool ten_bit_address, bool read) {
+    switch(begin) {
+    case I2CBeginRestart:
+        if(read) {
+            return ten_bit_address ? LL_I2C_GENERATE_RESTART_10BIT_READ :
+                                     LL_I2C_GENERATE_RESTART_7BIT_READ;
+        } else {
+            return ten_bit_address ? LL_I2C_GENERATE_RESTART_10BIT_WRITE :
+                                     LL_I2C_GENERATE_RESTART_7BIT_WRITE;
```
agarof, line 89: This could be simplified, since both values are equal, depending on how close to ST's LL you want to stick.

firmware/targets/f7/furi_hal/furi_hal_i2c.c
```diff
@@ -51,177 +51,248 @@ void furi_hal_i2c_release(FuriHalI2cBusHandle* handle) {
...
 
-bool furi_hal_i2c_tx(
-    FuriHalI2cBusHandle* handle,
-    uint8_t address,
-    const uint8_t* data,
-    uint8_t size,
-    uint32_t timeout) {
-    furi_check(handle->bus->current_handle == handle);
-    furi_assert(timeout > 0);
+static bool wait_for_idle(I2C_TypeDef* i2c, I2CBegin begin, FuriHalCortexTimer timer) {
```
skotopes: all methods must be prefixed with package name, in this case `furi_hal_i2c_`

firmware/targets/furi_hal_include/furi_hal_i2c.h
```diff
@@ -6,13 +6,39 @@
...
+typedef enum {
+    /*! Begin the transaction by sending a START condition followed by the address */
+    I2CBeginStart,
+    /*! Begin the transaction by sending a RESTART condition followed by the address
+     * @note Must follow a transaction ended with I2CEndAwaitRestart */
+    I2CBeginRestart,
+    /*! Continue the previous transaction with new data
+     * @note Must follow a transaction ended with I2CEndPause and be of the same type (RX/TX) */
+    I2CBeginResume,
+} I2CBegin;
```
skotopes: FuriHalI2cBegin

firmware/targets/furi_hal_include/furi_hal_i2c.h
```diff
@@ -6,13 +6,39 @@
...
+typedef enum {
+    /*! End the transaction by sending a STOP condition */
+    I2CEndStop,
+    /*! End the transaction by clock stretching
+     *  @note Must be followed by a transaction using I2CBeginRestart */
+    I2CEndAwaitRestart,
+    /*! Pauses the transaction by clock stretching
+     *  @note Must be followed by a transaction using I2CBeginResume */
+    I2CEndPause,
+} I2CEnd;
```
skotopes: FuriHalI2cEnd

firmware/targets/furi_hal_include/furi_hal_i2c.h
```diff
@@ -40,34 +66,80 @@ void furi_hal_i2c_release(FuriHalI2cBusHandle* handle);
...
+ */
+bool furi_hal_i2c_tx_ext(
+    FuriHalI2cBusHandle* handle,
+    uint16_t address,
+    bool ten_bit,
+    uint8_t const* data,
+    size_t size,
+    I2CBegin begin,
+    I2CEnd end,
+    FuriHalCortexTimer timer);
```
skotopes: why do we need FuriHalCortexTimer here?

skotopes (changes requested):
Unit tests needed: can be gauge, charger or led driver.

skotopes:
un draft when ready

firmware/targets/furi_hal_include/furi_hal_i2c.h
agarof: To be able to have a common timer for multiple transactions.
For example `furi_hal_i2c_write_mem` uses the same timer both for sending `mem_addr` and the data to write.

agarof:
I have fixed the functions and enums naming and have checked that the `furi_hal` (LED driver communication) and `power` (charger) tests pass.

firmware/targets/furi_hal_include/furi_hal_i2c.h
skotopes: Do we want it to be like that? If i'm doing tx + rx, what exactly I'm specifying in timeout?

skotopes (changes requested):

skotopes:
I've pushed bunch of small edits: docs update, const qualifier usage cleanup, fix with incorrect argument pass. In general everything looks good.

We do need one last thing: unit tests, they must cover newly added features.
You can use any of existing flipper chips or some other popular one that we can buy and add to our unit tests stand.

agarof:
I have added two unit tests, one testing the Restart and Pause / Resume features, the other testing the auto-reload feature.

The first uses the LED driver, same as the existing tests.
The second uses an I2C EEPROM connected to the external bus. It writes 512 bytes 16 at a time then reads them all at once, so the EEPROM needs to be at least 512 bytes large and have a 16 bytes page write buffer. Most I2C EEPROMs I have found, such as [these](https://ww1.microchip.com/downloads/aemDocuments/documents/MPD/ProductDocuments/DataSheets/24AA04-24LC04B-24FC04-4K-I2C-Serial-EEPROM-20001708P.pdf) [two](https://www.onsemi.com/pdf/datasheet/cat24c01-d.pdf), match these characteristics and have the same addressing scheme as the one I have (I don't have a reference for it).

skotopes (dismissed):

skotopes (approved):

firmware/targets/furi_hal_include/furi_hal_i2c.h
skotopes: I've reverted this part, because: 
- For the sake of API simplicity
- CortexTimer is something that I'd rather to not to expose outside, it may be replaced in future
- Some further considerations needed: multithread OS may give time to another thread while we using I2C

firmware/targets/f7/furi_hal/furi_hal_i2c.c
skotopes: We'd rather hide anything related to ST LL inside.

skotopes:
That was some great work, you did well ;-)

PR #3144, opened 2023-10-12
Fix various crashes if debug libraries used

skotopes:
# What's new

- Fix various crashes with LIB_DEBUG=1
- Libs: update ERC and MGG contrast
- Furi: fix double crash caused by `bkpt` use outside of debug session
- FuriHal: enable HSI in stop mode only if we use STOP0, proper SMPS selected clock assert
- Various PVS warning fixes
- FuriHal: fix incorrect CCID statuses returned in some cases
- Update display contrast

# Verification 

-  `./fbt LIB_DEBUG=1 flash_usb_full` 
- full set of tests 

# Checklist (For Reviewer)


firmware/targets/f7/src/update.c
```diff
@@ -76,14 +76,15 @@ static bool flipper_update_load_stage(const FuriString* work_dir, UpdateManifest
     }
     furi_string_free(loader_img_path);
 
+    furi_check(stat.fsize);
```
hedger: may cause boot loops - and if file is empty, we'll bail out later on CRC check

hedger (approved):

PR #3165, opened 2023-10-24
[FL-3637] NFC RC fixes

gornekich:
# What's new

- Remove build settings for NFC lib
- Fix NFC IRQ pin deinit
- Use ASK instead of OOK demodulator in transparent mode
- Add upper page bound for NTAGI2C1K

# Verification 

- Test NFC application
- Confirm iso15693 and SLIX emulation is working

# Checklist (For Reviewer)

### DIFF
--- a/lib/signal_reader/signal_reader.c
+++ b/lib/signal_reader/signal_reader.c
@@ -74,7 +74,7 @@ SignalReader* signal_reader_alloc(const GpioPin* gpio_pin, uint32_t size) {
     SignalReader* instance = malloc(sizeof(SignalReader));
 
     instance->pin = gpio_pin;
-    instance->pull = GpioPullDown;
+    instance->pull = GpioPullNo;
 
     instance->buffer_size = size;
     instance->gpio_buffer = malloc(sizeof(uint16_t) * size * 8);
--- a/lib/signal_reader/signal_reader.c
+++ b/lib/signal_reader/signal_reader.c
@@ -74,7 +74,7 @@ SignalReader* signal_reader_alloc(const GpioPin* gpio_pin, uint32_t size) {
     SignalReader* instance = malloc(sizeof(SignalReader));
 
     instance->pin = gpio_pin;
-    instance->pull = GpioPullNo;
+    instance->pull = GpioPullDown;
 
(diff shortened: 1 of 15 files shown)


gornekich:
@nvx @pcunning 
Thanks a lot for noticing that some readers use ASK modulation. I made some changes and now Flipper parse ASK signal in transparent mode. I will be glad if you test picopass emulation from https://github.com/flipperdevices/flipperzero-good-faps/pull/62  built with SDK from this branch.

Also https://github.com/flipperdevices/flipperzero-good-faps/pull/62 PR is temporary. I will split it to make code review easier soon. Now I just want to be sure that nfc library works correctly.

lib/signal_reader/signal_reader.c
```diff
@@ -74,7 +74,7 @@ SignalReader* signal_reader_alloc(const GpioPin* gpio_pin, uint32_t size) {
     SignalReader* instance = malloc(sizeof(SignalReader));
 
     instance->pin = gpio_pin;
-    instance->pull = GpioPullNo;
+    instance->pull = GpioPullDown;
```
nvx: Is this needed as it looks like `iso15693_parser_alloc` is calling `signal_reader_set_pull()` already.

If you can receive any 15693 signal from the reader then it's working, not having the pull-down set results in complete deafness.

gornekich: You are right, my mistake. Reverted this change.

I can successfully emulate iso15693, slix and picopass tags using proxmark as a reader. But proxmark use OOK modulation for these protocols, and I wonder how flipper emulation works with readers which use ASK modulation

nvx: If you're running an up-to-date version (I recommend running master not the release) of the iceman fork - I added a --shallow option to `hf iclass` commands that will make it use ASK modulation instead of OOK which is better than nothing if you don't have a HID reader that does ASK.

gornekich: Testes with --shallow option - works fine. Thanks!

skotopes (dismissed):

skotopes (approved):

PR #3176, opened 2023-10-30
Mifare Classic low level poller

augustozanellato:
# What's new

- Low-level poller for mifare classic, allows for doing custom command chaining in user applications with a sync api.
- Add support for nested auth in mifare classic poller

# Verification 

Not sure how to provide verification steps. I developed a small test app that I can either link here or add as part of unit tests, but for that I'll need some advice.

# Checklist (For Reviewer)

### DIFF
--- a/lib/nfc/SConscript
+++ b/lib/nfc/SConscript
@@ -38,6 +38,7 @@ env.Append(
         # Sync API
         File("protocols/iso14443_3a/iso14443_3a_poller_sync_api.h"),
         File("protocols/mf_ultralight/mf_ultralight_poller_sync_api.h"),
+        File("protocols/mf_classic/mf_classic_poller_ll_sync_api.h"),
         File("protocols/mf_classic/mf_classic_poller_sync_api.h"),
         # Misc
         File("helpers/nfc_util.h"),
--- a/lib/nfc/protocols/mf_classic/crypto1.c
+++ b/lib/nfc/protocols/mf_classic/crypto1.c
@@ -143,7 +143,8 @@ void crypto1_encrypt_reader_nonce(
     uint32_t cuid,
     uint8_t* nt,
     uint8_t* nr,
-    BitBuffer* out) {
+    BitBuffer* out,
+    bool is_nested) {
     furi_assert(crypto);
     furi_assert(nt);
     furi_assert(nr);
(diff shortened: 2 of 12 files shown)


augustozanellato:
A thing to be noted: calling `mf_classic_poller_ll_free` without having called `mf_classic_poller_ll_halt` before will result in an hang.
That is kinda intended, adding a field to the poller ctx to manage auto halt on free makes the code a lil bit uglier IMHO, and anyone touching something named "ll" as in "low level" should be careful about what/how they do stuff.

gornekich:
Hello @augustozanellato !
I can see that you want to add an opportunity to turn on field, execute some MFC commands and turn off the field after that. Your solution is good, but I had an idea to expose mifare_classic_poller_async* API, so that user could start Iso14443-3a poller and write custom state machine.

Let's think together on improving the API. It would be helpful to see the application you mentioned which works with your changes

augustozanellato:
>I can see that you want to add an opportunity to turn on field, execute some MFC commands and turn off the field after that.

Yeah, that’s the idea.

> I had an idea to expose mifare_classic_poller_async* API, so that user could start Iso14443-3a poller and write custom state machine.

That would also work, in that case I’ll probably close this PR and make another with just the first commit taken from this one in order to allow for nested auth. 

>It would be helpful to see the application you mentioned which works with your changes

It’s doesn’t do much, it was made just to test if the api was working correctly, especially nested auth, I’ll share it tomorrow.

augustozanellato:
@gornekich here's the test app I used to verify that the api was working correctly: https://github.com/augustozanellato/flipper_mfc_ll_test

I also merged in the latest dev changes and removed a leftover debug print statement.

gornekich:
Hello @augustozanellato . Thanks for sending test app. I need some time to think about better API and come back a littler bit later.

augustozanellato:
> I need some time to think about better API and come back a littler bit later.

No problem with that! In case you’d ever need an opinion from the perspective of an api consumer feel free to hit me up, here or on discord.

gornekich:
Hello @augustozanellato !
Just finished working on #3214 . I introduced nfc_poller_start_ex() API, which can be used to start MfClassic poller and process Iso14443_3a events. Also many internal pollers functions were exposed.

I believe that these changes make writing custom pollers for all protocols much easier. I would be glad if you have a look on my PR and let me know if it solves your issues.

From this PR I really want nested authentication implementation. However I would prefer to have separate mf_classic_auth_nested() function rather than add extra parameter in mf_classi_auth() function

augustozanellato:
> Hello @augustozanellato ! Just finished working on #3214 . I introduced nfc_poller_start_ex() API, which can be used to start MfClassic poller and process Iso14443_3a events. Also many internal pollers functions were exposed.
> 
> I believe that these changes make writing custom pollers for all protocols much easier. I would be glad if you have a look on my PR and let me know if it solves your issues.

Sorry for the delay, it has been a busy week. I’ll probably be able to provide you with feedback early next week, probably on Monday. 

> From this PR I really want nested authentication implementation. However I would prefer to have separate mf_classic_auth_nested() function rather than add extra parameter in mf_classi_auth() function

Sure, I’ll close this PR and make another one against updated dev branch next week, probably also on Monday.

PR #3185, opened 2023-11-01
ST25TB poller refining

augustozanellato:
# What's new

- Reworked internal poller API to be much cleaner
- Introduced sync poller API
- Added write API

# Verification 

- Read a ST25TB and verify it's still read correctly

# Checklist (For Reviewer)

### DIFF
--- a/lib/nfc/protocols/st25tb/st25tb_poller.c
+++ b/lib/nfc/protocols/st25tb/st25tb_poller.c
@@ -62,17 +62,30 @@ static void
 }
 
 static St25tbPollerState
+    st25tb_poller_return_to_idle_on_error(St25tbError* error, St25tbPollerState next_state) {
+    if(*error == St25tbErrorNone) {
+        return next_state;
+    } else {
+        return St25tbPollerStateIdle;
+    }
+}
+
+static St25tbPollerState
     st25tb_poller_state_idle_handler(St25tbError* error, St25tbPoller* instance) {
     instance->st25tb_event.type = St25tbPollerEventTypeReady;
+
     *error = st25tb_poller_async_select(instance, NULL);
-    return St25tbPollerStateSelected;
+
+    return st25tb_poller_return_to_idle_on_error(error, St25tbPollerStateSelected);
 }
 
 static St25tbPollerState
     st25tb_poller_state_selected_handler(St25tbError* error, St25tbPoller* instance) {
     instance->st25tb_event.type = St25tbPollerEventTypeReadSuccessful;
+
     *error = st25tb_poller_async_read(instance, instance->data);
-    return St25tbPollerStateRead;
+
+    return st25tb_poller_return_to_idle_on_error(error, St25tbPollerStateRead);
 }
 
 static St25tbPollerState
--- a/lib/nfc/protocols/st25tb/st25tb_poller.c
+++ b/lib/nfc/protocols/st25tb/st25tb_poller.c
@@ -1,5 +1,3 @@
-#include "protocols/nfc_protocol.h"
-#include "protocols/st25tb/st25tb.h"
 #include "st25tb_poller_i.h"
 
 #include <nfc/protocols/nfc_poller_base.h>
@@ -19,6 +17,7 @@ static St25tbPoller* st25tb_poller_alloc(Nfc* nfc) {
     furi_assert(nfc);
 
     St25tbPoller* instance = malloc(sizeof(St25tbPoller));
+    instance->state = St25tbPollerStateIdle;
     instance->nfc = nfc;
     instance->tx_buffer = bit_buffer_alloc(ST25TB_POLLER_MAX_BUFFER_SIZE);
     instance->rx_buffer = bit_buffer_alloc(ST25TB_POLLER_MAX_BUFFER_SIZE);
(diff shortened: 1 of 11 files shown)


gornekich (changes requested):
Please, ignore this change request. I will come with detailed review soon

lib/nfc/protocols/st25tb/st25tb_poller.h
```diff
@@ -14,6 +14,7 @@ typedef struct St25tbPoller St25tbPoller;
 typedef enum {
     St25tbPollerEventTypeError,
     St25tbPollerEventTypeReady,
+    St25tbPollerEventTypeReadSuccessful,
 } St25tbPollerEventType;
```
gornekich, line 18: I am not sure that both Ready and ReadSuccess events are necessary here. For example, if we take Iso14443-3a protocol, there is only Ready event, because there are some protocols based on Iso14443-3a protocol. And if we take Mifare Ultralight protocol, it runs state machine on Iso14443-3a Ready event and spawns ReadSuccess or ReadFail events.
Since there are no other protocols based on St25tb poller, I would leave just ReadSuccess event.

lib/nfc/protocols/st25tb/st25tb_poller.c
```diff
@@ -70,24 +69,28 @@ static NfcCommand st25tb_poller_run(NfcGenericEvent event, void* context) {
...
+            instance->st25tb_event_data.error = error;
+            command = instance->callback(instance->general_event, instance->context);
         } else {
-            instance->st25tb_event.type = St25tbPollerEventTypeReady;
-            instance->st25tb_event_data.error = St25tbErrorNone;
+            instance->st25tb_event.type = St25tbPollerEventTypeError;
+            instance->st25tb_event_data.error = error;
             command = instance->callback(instance->general_event, instance->context);
+            // Add delay to switch context
+            furi_delay_ms(100);
```
gornekich: I can see the problem here. 
On first NfcEventTypePollerReady event you will  be in St25tbPollerStateIdle state and enter callback with St25tbPollerEventTypeReady event.
On second run you will have St25tbPollerStateSelected state and also spawn St25tbPollerEventTypeReady event.
And only on third run you will be in St25tbPollerStateRead state and spawn St25tbPollerEventTypeReadSuccessful event.

Sure, the application will work, but I think it's incorrect logic. I see that you split your poller in 3 states. I suggest adding separate state handlers array and call them in run() function, like it's done in mf_ultralight_poller.c for example. This would be much easier to read and expand.

gornekich: Also, if you do that, it would be nice if "async" functions will not change poller state. 

Firstly, it may cause some problems if you use those functions in different places, in sync API for example. 

Secondly, it would be much easier to understand your state machine. 

You should just call "async" functions and set new poller state based on returned error code

lib/nfc/protocols/st25tb/st25tb_poller_sync_api.c
```diff
@@ -0,0 +1,147 @@
...
+    } read;
+    struct {
+        uint8_t block_num;
+        uint32_t* block;
+    } read_block;
+    struct {
+        uint8_t block_num;
+        uint32_t block;
+    } write_block;
+} St25tbPollerCmdData;
```
gornekich: Please, avoid this kind of definitions. It would be much easier to read if you make an explicit type with typedef for each PollerCmdData and put them in union.

Making a new type is free from memory resources point of view, but makes your code much easier to read, maintain and prevents you from lot's of mistakes

lib/nfc/protocols/st25tb/st25tb_poller_sync_api.c
```diff
@@ -0,0 +1,147 @@
...
+}
+
+St25tbError st25tb_poller_read_block(Nfc* nfc, uint8_t block_num, uint32_t* block) {
+    St25tbPollerContext poller_context = {
+        .cmd_type = St25tbPollerCmdTypeReadBlock,
+        .cmd_data = {
+            .read_block = {
+                .block = block,
+                .block_num = block_num,
+            }}};
```
gornekich: Here and below there is a minor issue with formating. I prefer to put a comma after each variable initialization, and code looks neater

targets/f18/api_symbols.csv
```diff
@@ -1,5 +1,5 @@
 entry,status,name,type,params
-Version,+,44.0,,
+Version,+,44.2,,
```
gornekich: Please, change to 44.1

targets/f7/api_symbols.csv
```diff
@@ -1,5 +1,5 @@
 entry,status,name,type,params
-Version,+,44.0,,
+Version,+,44.2,,
```
gornekich: Same here

gornekich (changes requested):
Thanks for PR!
I found some issues in code which you can fix or we can discuss.

lib/nfc/protocols/st25tb/st25tb_poller.h
augustozanellato: The Ready event is mainly used by the sync api in order to not have to wait for a full tag read before executing commands in case eg detect type or read block is called

lib/nfc/protocols/st25tb/st25tb_poller.c
augustozanellato: I'll try to explain better how the thing works:
An api consumer only sees one of two events:
- St25tbPollerEventTypeReady: an st25tb tag is available in field and has been selected
- St25tbPollerEventTypeReadSuccessful: a tag has been fully read

The first event only exists to accommodate the existence of the sync api, this way it doesn't need to wait for a full read at every call.

I pushed a couple of commits that clean up the poller api and state machine according to your suggestions.

augustozanellato:
Manually merging in latest dev branch changes takes as much effort as redoing them manually, I'll make a new PR with these changes soon.

Issue #3195, opened 2023-11-04
Running build on Windows gives error: `Thu Nov 2 17:28:39 2023 does not match format %a %b %d %H:%M:%S %Y %z`


derskythe:
### Describe the bug.

Clean run `fbt` gives such error:

```
Traceback (most recent call last):
File "D:\_repos\!flipper\official\scripts\[version.py](http://version.py/)", line 176, in <module>
Main()()
File "D:\_repos\!flipper\official\scripts\flipper\[app.py](http://app.py/)", line 41, in call
return_code = self.call()
^^^^^^^^^^^
File "D:\_repos\!flipper\official\scripts\flipper\[app.py](http://app.py/)", line 57, in call
return self.args.func()
^^^^^^^^^^^^^^^^
File "D:\_repos\!flipper\official\scripts\[version.py](http://version.py/)", line 114, in generate
current_info = GitVersion(self.args.sourcedir).get_version_info()
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "D:\_repos\!flipper\official\scripts\[version.py](http://version.py/)", line 48, in get_version_info
commit_date = datetime.strptime(
^^^^^^^^^^^^^^^^^^
File "D:\_repos\!flipper\official\toolchain\x86_64-windows\python\Lib\_strptime.py", line 568, in _strptime_datetime
tt, fraction, gmtoff_fraction = _strptime(data_string, format)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "D:\_repos\!flipper\official\toolchain\x86_64-windows\python\Lib\_strptime.py", line 349, in _strptime
raise ValueError("time data %r does not match format %r" %
ValueError: time data 'Thu Nov 2 17:28:39 2023' does not match format '%a %b %d %H:%M:%S %Y %z'
scons: *** [build\f7-firmware-D\toolbox\version.inc.h] Error 1
```

### Reproduction

1. `git clone`
2. `./fbt`

### Target

Windows

## Update

The problem is OS tolerant and caused by the git config:
```
[log]
   date = local
```

PR #3196, opened 2023-11-04
fix: build on windows gives datetime format error

derskythe:
# What's new

- Fixed build on Windows #3195
- Tested on Windows and Debian

# Verification 

- `git clone`
- run `./fbt`

# Checklist (For Reviewer)

### DIFF
--- a/scripts/version.py
+++ b/scripts/version.py
@@ -45,10 +45,16 @@ class GitVersion:
                 int(os.environ["SOURCE_DATE_EPOCH"])
             )
         else:
-            commit_date = datetime.strptime(
-                self._exec_git("log -1 --format=%cd").strip(),
-                "%a %b %d %H:%M:%S %Y %z",
-            )
+            if os.name == 'nt':
+                commit_date = datetime.strptime(
+                    self._exec_git("log -1 --format=%cd").strip(),
+                    "%a %b %d %H:%M:%S %Y",
+                )
+            else:
+                commit_date = datetime.strptime(
+                    self._exec_git("log -1 --format=%cd").strip(),
+                    "%a %b %d %H:%M:%S %Y %z",
+                )
 
         return {
             "GIT_COMMIT": commit,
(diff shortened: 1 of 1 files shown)


hedger:
That's odd, what's your setup - Windows, Git versions? System locale?

derskythe:
> That's odd, what's your setup - Windows, Git versions? System locale?

Yes, I checked on the different system and can't reproduce.
But, different system has non-English locale.
So:

## Problem Windows 
- Version `10.0.19045 Build 19045`
- Git version: `2.42.0.windows.2`
- Locale: `en-US`
- `git log -1 --format=%cd` output: `Thu Nov 2 17:28:39 2023`

## Working Windows
- Version `10.0.19045 Build 19045`
- Git version: `2.41.0.windows.3`
- Locale: `ru-RU`
- `git log -1 --format=%cd` output: `Thu Nov 2 22:28:39 2023 +0900`

**Is the time zone of latest commit really that important when assembling?**

derskythe:
Found difference, on the problem system `git config -e --global`:
```
[log]
   date = local
```

I can change my PR to check:
```bash
git config --get log.date
```

hedger:
I'm not sure if it's feasible to support all possible date formats. I think it would be better to try parsing the most basic format (omitting the timezone) and, in other cases, throwing an error, suggesting the user to fix their git config.

derskythe:
@hedger Agree, but I think some people don't want to calc local date of each committer and using this setting. 
This is not incorrect settings in git config. 
Why force these users to disable this setting just to compile the firmware?

hedger:
There are other options for setting time format, `(relative|local|default|iso|iso-strict|rfc|short|raw)`. Luckily, `git log` can override the effective time format - no need to check for whatever user has set there.

So, you can just add `--date=default` to `git log` args and leave other parts as they are.

derskythe:
Create PR #3202, this PR is dirty

PR #3229, opened 2023-11-19
Fixed error when reading an iso15693 nfc tag with no blocks

pborsutzki:
Fixed a zero allocation error when reading an iso15693 nfc tag with no additional blocks.

# What's new

When reading an iso15693 nfc tag that reports zero blocks, do not try to allocate memory for reading or try to read the inexistent blocks. This works around a hard crash of the flipper when reading e.g. a FreeStyle Libre 3 tag.

# Verification 

Try to read an iso15693 nfc tag that just has an UID but no additional data. The flipper should not crash. Previously it crashed when trying to allocate 0 bytes through `malloc` in `simple_array_init`.

# Checklist (For Reviewer)


gornekich (approved):
Thanks!

skotopes (approved):

PR #3239, opened 2023-11-21
ST25TB poller refining + write support

augustozanellato:
# What's new

- reworked async poller
- added sync poller
- added write support

This PR should be ~same to #3185 but based on the improved NFC APIs

# Verification 

- Read a ST25TB and verify it's still read correctly

# Checklist (For Reviewer)


lib/nfc/protocols/st25tb/st25tb_poller.c
```diff
@@ -1,13 +1,12 @@
...
+#include "st25tb_poller.h"
 #include "st25tb_poller_i.h"
 
 #include <nfc/protocols/nfc_poller_base.h>
 
-#include <furi.h>
-
 #define TAG "ST25TBPoller"
 
+typedef St25tbPollerState (*St25tbPollerStateHandler)(St25tbError* error, St25tbPoller* instance);
```
gornekich: I would prefer all pollers to have similar handlers signature. In this case I would make
`typedef NfcCommand (*St25tbPollerStateHandler)(St25tbPoller* instance);`
This would be so easier to maintain NFC library if all protocols are written similarly

lib/nfc/protocols/st25tb/st25tb_poller.c
```diff
@@ -60,6 +60,46 @@ static void
...
 }
 
+static St25tbPollerState
+    st25tb_poller_return_to_idle_on_error(St25tbError* error, St25tbPollerState next_state) {
+    if(*error == St25tbErrorNone) {
+        return next_state;
+    } else {
+        return St25tbPollerStateIdle;
+    }
+}
```
gornekich: From a "code reader" point of view it would be much easier to see error processing and following state change directly in state handlers.

lib/nfc/protocols/st25tb/st25tb_poller.c
```diff
@@ -60,6 +60,46 @@ static void
...
+
+static St25tbPollerState
+    st25tb_poller_state_idle_handler(St25tbError* error, St25tbPoller* instance) {
+    instance->st25tb_event.type = St25tbPollerEventTypeReady;
+
+    *error = st25tb_poller_select(instance, NULL);
+
+    return st25tb_poller_return_to_idle_on_error(error, St25tbPollerStateSelected);
```
gornekich: The poller handlers tasks are:
- Exchange data with tag if necessary
- Process errors after data exchange
- Select next state
- Prepare event, call callback to upper layer and return command

gornekich: I would rewrite this handler:
```
static NfcCommand st25tb_poller_select_handler(St25tbPoller* instance) {
    NfcCommand command = NfcCommandContinue;

    do {
        St25tbError error = st25tb_poller_select(instance, NULL);
        if(error != St25tbErrorNone) {
            instance->state = // Fail state;
            break;
        }

        instance->st25tb_event.type = St25tbPollerEventTypeReady;
        command = instance->callback(instance->general_event, instance->context);
        instance->state = // Next state
    } while(false);

    return command;
}

```

lib/nfc/protocols/st25tb/st25tb_poller.c
```diff
@@ -69,25 +109,21 @@ static NfcCommand st25tb_poller_run(NfcGenericEvent event, void* context) {
...
+            instance->st25tb_event_data.error = error;
+            command = instance->callback(instance->general_event, instance->context);
         } else {
-            instance->st25tb_event.type = St25tbPollerEventTypeReady;
-            instance->st25tb_event_data.error = St25tbErrorNone;
+            instance->st25tb_event.type = St25tbPollerEventTypeError;
+            instance->st25tb_event_data.error = error;
             command = instance->callback(instance->general_event, instance->context);
+            // Add delay to switch context
+            furi_delay_ms(100);
```
gornekich: If you rewrite poller handlers as suggested, the NfcEventTypePollerReady event handler would be like this:

```
NfcCommand command = NfcCommandContinue;
if(nfc_event->type == NfcEventTypePollerReady) {
    command = st25tb_poller_state_handlers[instance->state](instance);
}
```

lib/nfc/protocols/st25tb/st25tb_poller.c
```diff
@@ -60,6 +60,46 @@ static void
...
+}
+
+static St25tbPollerState
+    st25tb_poller_state_selected_handler(St25tbError* error, St25tbPoller* instance) {
+    instance->st25tb_event.type = St25tbPollerEventTypeReadSuccessful;
+
+    *error = st25tb_poller_read(instance, instance->data);
+
+    return st25tb_poller_return_to_idle_on_error(error, St25tbPollerStateRead);
+}
```
gornekich: I can see small problem here. The poller states should be "relatively fast". If we read St25tbType04k or St25tbTypeX4k where there are 128 blocks, you will not exit this state until you read all blocks or reading is failed. That leads to UX "freeze" if user wants to stop reading by pressing BACK. To achieve more responsive system we should return from poller states as quickly as possible

gornekich: I would add variable current_block into St25tbPoller. We should reset this variable to 0 at idle state and read handler can be like this:

```
static NfcCommand st25tb_poller_read_handler(St25tbPoller* instance) {
    St25tbError error = St25tbErrorNone;

    do {
        uint8_t total_blocks = st25tb_get_block_count(instance->data->type);
        if(instance->current_block == total_blocks) {
            error = st25tb_poller_read_block(
                instance, &instance->data->system_otp_block, ST25TB_SYSTEM_OTP_BLOCK);
            if(error != St25tbErrorNone) {
                FURI_LOG_E(TAG, "Failrd to read OTP block");
                instance->state = St25PollerStateFail;
                break;
            } else {
                instance->state = St25PollerStateSuccess;
                break;
            }
        } else {
            error = st25tb_poller_read_block(
                instance,
                &instance->data->blocks[instance->current_block],
                instance->current_block);
            if(error != St25tbErrorNone) {
                FURI_LOG_E(TAG, "Failed to read block %d", instance->current_block);
                instance->state = St25PollerStateFail;
                break;
            }

            instance->current_block++;
        }
    } while(false);

    return NfcCommandContinue;
}

```

gornekich (commented):
Hello @augustozanellato , thanks for PR!
First of all - everything works fine. However there are some things I want to discuss.

It would be nice if all pollers in NFC library looks similarly. That would help to maintain NFC subsystem a lot. I left some suggestions in code review.

Also I want to discuss poller states. Since your poller becomes bigger and you already added write capabilities in sync API, I suggest expanding poller itself. I would make these states in poller:

- Idle - reset St25tbPoller instance
- Select - selects card, emit "Ready" event with card type in event data
- Request mode - emit "RequestMode" event and receive mode in event data. Modes are St25tbPollerModeRead and St25tbPollerModeWrite
- Read - reads tag and switches to Fail or Success states
- Write - emits "RequestBlockData" event and receive block and block number in event data. Performs write command and switches to Fail or Success states.
- Fail - emits "Fail" event with error code in event data. Ends with furi_delay_ms(100) to switch context and switches to Idle state.
- Success - emits "Success" event with error code in event data. Ends with furi_delay_ms(100) to switch context and switches to Idle state.

In this case your sync API would be easy to write and we will have easier way to implement "write" in NFC application. You can see how it's done in mf classic poller for example.

Please, let me know what you think about this. We can do this step by step: merge this PR first and then refactore st25tb poller in future

gornekich:
Also please don't forget to lint sources :)
`./fbt format`

augustozanellato:
> Also please don't forget to lint sources :) `./fbt format`

I usually have format on save enabled, I'll figure out why it didn't work this time.

> It would be nice if all pollers in NFC library looks similarly. 

The thing that made me architect this poller different from the other ones was that this one is kinda an exception from the standard poller in the sense that it doesn't have either a parent or any children protocols, but has to implement both "low level" (anticollision, selection etc) and "high level" (data reading and writing).

lib/nfc/protocols/st25tb/st25tb_poller.c
augustozanellato: Oh... I probably misunderstood your comment on #3185, but I think I got it now. I'll come back with a new design in a couple days.

augustozanellato:
> Please, let me know what you think about this. We can do this step by step: merge this PR first and then refactore st25tb poller in future

I think we should just iterate on this pr until we get to a decent shape without breaking API even more times :)

augustozanellato:
It took less time than expected.
There are two deviations from your suggestions:
- there isn't an "idle" state, i think it'd be kinda useless, feel free to correct me on that;
- there isn't a `RequestBlockData` event because I made `RequestMode` handle mode arguments.

gornekich (dismissed):
Well done, thank you! I think poller looks much better now

augustozanellato:
> Well done, thank you! I think poller look much better now

Thanks for your suggestions :)

gornekich:
> there isn't an "idle" state, i think it'd be kinda useless, feel free to correct me on that;

No problems with that. Usually I add Idle state to set all variables in poller instance to default values. Since there are not so many variables in this poller, Idle state is useless. If the poller will grow, we can add this state if necessary.

> there isn't a `RequestBlockData` event because I made `RequestMode` handle mode arguments.

I was thinking about different Write mode behavior - to call "RequestBlockData" event, write blocks until upper layer provides data. When upper layer stops providing data poller switches to Success state. This is quite flexible behavior which allows you to write not only one block after anticollision, but perform multiple write operations in one communication session.
I don't see any reasons to change it right. We can return to this later, if we want to add "Write" option in NFC application. These changes will cause minor API update.

augustozanellato:
> > there isn't an "idle" state, i think it'd be kinda useless, feel free to correct me on that;
> 
> No problems with that. Usually I add Idle state to set all variables in poller instance to default values. Since there are not so many variables in this poller, Idle state is useless. If the poller will grow, we can add this state if necessary.

Yeah that makes sense.

> > there isn't a `RequestBlockData` event because I made `RequestMode` handle mode arguments.
> 
> I was thinking about different Write mode behavior - to call "RequestBlockData" event, write blocks until upper layer provides data. When upper layer stops providing data poller switches to Success state. This is quite flexible behavior which allows you to write not only one block after anticollision, but perform multiple write operations in one communication session. I don't see any reasons to change it right. We can return to this later, if we want to add "Write" option in NFC application. These changes will cause minor API update.

I didn't think about that, but that could be solved by having `st25tb_poller_success_handler` set poller state to `St25tbPollerStateRequestMode` so that in a communication session an arbitrary number of operations can be done. Tell me what do you think about this paradigm.

gornekich:
> I didn't think about that, but that could be solved by having `st25tb_poller_success_handler` set poller state to `St25tbPollerStateRequestMode` so that in a communication session an arbitrary number of operations can be done. Tell me what do you think about this paradigm.

Yes, that's actually a good solution. Will you do that in this PR?

augustozanellato:
> Yes, that's actually a good solution. Will you do that in this PR?

Done, I was just waiting for some feedback :)

gornekich:
Perfect, thanks!

augustozanellato:
The failing unit test is subghz related, I don't think that could be directly related to this pr :thinking:

gornekich:
Yes, the problem is not this PR

augustozanellato:
Looks like extra - read specific card type - st25tb is now broken, I'll hook up a debugger and figure out why.

augustozanellato:
Bug fixed, while investigating that one I also found another one, more details in commit message.
PR should now be ready to review and merge.

skotopes (approved):

PR #3241, opened 2023-11-24
nfc: m1k-based Aime (non-AIC) card support

magicnat:
# What's new

- Added a new NFC supported card plugin for the old red-black colored [Aime cards](https://my-aime.net/en/) (MIFARE Classic 1K based)

# Verification 

- Read an Aime card. Example card: [aime_example.nfc.txt](https://github.com/flipperdevices/flipperzero-firmware/files/13455914/aime_example.nfc.txt)

Note: The example card above does not contain a valid access code and will not work on real arcades. 

# Checklist (For Reviewer)


dogtopus:
Soccer game key finally online

gornekich (approved):
Thanks!

skotopes (approved):

ry4000:
Tested this on my end - I was able to read my Aime Cards; many thanks!

That said, is there a way to provide similar support for [Bandai Namco Passports](https://banapass.net/setlocale/en)?

The MFC BNP access keys are provided in [PR #3314](https://github.com/flipperdevices/flipperzero-firmware/pull/3314).

Ideally, the BNP plugin looks for both Sector 0 Keys:

- Parse as a BNP if S0KA is 6090D00632F5 and S0KB is 574343467632; print the last 10 bytes in S0B2, which is the Access Code;

- Parse as a BNP if S0KA is 6090D00632F5, but S0KB is 019761AA8082; read is still valid, but instead print "Please check the back of your Bandai Namco Passport for your Access Code" because S0B2 would instead be a Value Block.

Unlike the Aime plugin, I am unsure if BNPs have a checksum; however, I do know that the BNP readers look for all Sector Keys in their correct spots, but does not check for correct ACLs [which may be modified without issue].

PR #3273, opened 2023-12-06
Fix T5577 reset function

v0lp3:
# What's new

In the t5577_write_reset the opcode transmitted was 0b10, according to any [T5577 documentation](https://ww1.microchip.com/downloads/en/DeviceDoc/ATA5577C-Read-Write-LF-RFID-IDIC-100-to-150-kHz-Data-Sheet-DS70005357B.pdf) the opcode for reset mode is 0b00. For future implementations I added the possibility to select the page to write.

# Verification 

# Checklist (For Reviewer)


skotopes:
@v0lp3 this PR breaks writing to 5577 completely

Here is simple test set, please ensure that all cards from this set can be written and then read back.
After that reopen PR.

[RFID Test Files.zip](https://github.com/flipperdevices/flipperzero-firmware/files/13650781/RFID.Test.Files.zip)

lib/lfrfid/tools/t5577.c
```diff
@@ -2,20 +2,26 @@
...
-#define T5577_TIMING_START_GAP 30
-#define T5577_TIMING_WRITE_GAP 18
-#define T5577_TIMING_DATA_0 24
-#define T5577_TIMING_DATA_1 56
+
+// Required time to write block, typical value is around 5.6ms
 #define T5577_TIMING_PROGRAM 700
 
+// 1 / 125khz == 8 microseconds; Time = Cycles * 8
+#define CYCLES_TO_TIME(time) ((time) / 125000UL)
```
DrZlo13: At least that place is completely broken, anything less than 125000 will be equal 0.

lib/lfrfid/tools/t5577.c
```diff
@@ -27,16 +33,13 @@ static void t5577_stop() {
 
 static void t5577_write_gap(uint32_t gap_time) {
     furi_hal_rfid_tim_read_pause();
-    furi_delay_us(gap_time * 8);
+    furi_delay_us(gap_time);
```
DrZlo13: Will be better to add something like `furi_delay_125k_clocks()` and count everything in field clocks.

lib/lfrfid/tools/t5577.c
```diff
@@ -27,16 +33,13 @@ static void t5577_stop() {
...
     furi_hal_rfid_tim_read_continue();
 }
 
 static void t5577_write_bit(bool value) {
-    if(value) {
-        furi_delay_us(T5577_TIMING_DATA_1 * 8);
-    } else {
-        furi_delay_us(T5577_TIMING_DATA_0 * 8);
-    }
+    uint32_t timing = (!value) ? T5577_TIMING_DATA_0 : T5577_TIMING_DATA_1;
```
DrZlo13: ```(value) ? T5577_TIMING_DATA_1 : T5577_TIMING_DATA_0```

DrZlo13 (changes requested):

lib/lfrfid/tools/t5577.c
v0lp3: Added `furi_delay_clocks_125khz` function in `furi/core/kernel.*`, idk if it's the right place

lib/lfrfid/tools/t5577.c
v0lp3: 😱 thanks

v0lp3:
> @v0lp3 this PR breaks writing to 5577 completely
> 
> Here is simple test set, please ensure that all cards from this set can be written and then read back. After that reopen PR.
> 
> [RFID Test Files.zip](https://github.com/flipperdevices/flipperzero-firmware/files/13650781/RFID.Test.Files.zip)

I'm sorry. I think this was caused by the problem @DrZlo13 pointed out. Anyway, I don't have a T5577 at the moment because I bought some T5577 rfid tags from Amazon but I'm pretty sure they are em4305 or other (title was T5577/EM4305 idk). 

I made this PR because I was reading T5577 datasheet and flipper's code to understand how to implement em4305 protocol and I noticed the t5577 opcode error, so if you or someone else can test it for me I would be grateful.

v0lp3:
> @v0lp3 this PR breaks writing to 5577 completely
> 
> Here is simple test set, please ensure that all cards from this set can be written and then read back. After that reopen PR.
> 
> [RFID Test Files.zip](https://github.com/flipperdevices/flipperzero-firmware/files/13650781/RFID.Test.Files.zip)

I think this was caused by the problem @DrZlo13 pointed out, now should be ok. Onestly I don't have a T5577 to test read/write at the moment, more than once I bought T5577 rfid tags but every time flipper fails to write them, so if you, or anyone else, could test this code for me I would be grateful.

lib/lfrfid/tools/t5577.c
DrZlo13: No, not really, the 125kHz delay is only applicable to the LF-RFID subsystem and does not belong to the OS kernel.

v0lp3: Ok, I renamed this function `t5577_furi_delay_clocks` and it was place in t5577.c as inline because it's basically just a 'rename' of furi_delay_us. I think we can move it out when we'll support in write more protocols other than t5577.

lib/lfrfid/tools/t5577.c
```diff
@@ -46,46 +58,55 @@ static void t5577_write_opcode(uint8_t value) {
...
-        t5577_write_bit((data >> (31 - i)) & 1);
+        t5577_write_bit(data >> (31 - i));
     }
 
-    // block address
-    t5577_write_bit((block >> 2) & 1);
-    t5577_write_bit((block >> 1) & 1);
-    t5577_write_bit((block >> 0) & 1);
+    // Block address -> All blocks in T5577 are 8: 0 to 7
+    t5577_write_bit(block >> 2);
```
skotopes, line 80: seriously?

skotopes (changes requested):

Issue #3197, opened 2023-11-04
KDF plugins


noproto:
### Describe the enhancement you're suggesting.

A significant enhancement for the NFC app would be an interface to allow libraries/shared objects to be registered (or loaded from a fixed SD card path) which could be used as custom dictionaries for bruteforce attacks. This is particularly useful for key derivation functions (KDFs).

As discussed on the Q&A, we're hoping to identify an ideal location/API to implement this which would be compatible with multiple NFC protocols without duplicating code - such as MIFARE Classic, DESFire, and Ultralight C. Either C or JavaScript functions can be written for use with this handler. The external function is loaded from the SD card and called with the known card parameters (e.g. ATQA, SAK, CUID), which returns an array of zero or more keys. Each key is tested as a part of the overall dictionary attack against the card.

This approach ensures that the firmware remains free from copyrighted or proprietary algorithms, while (often) eliminating more advanced NFC attacks such as MFKey or Nested.

AloneLiberty:
We already have NFC app plugins, it's capable of loading required plugins on demand and exposes all required APIs (?). For simple KDFs that should be enough, but you can't for example display anything or ask for input (don't know in what case it might be necessary). And the only thing we don't have right now is probably application catalog support.

nvx:
> We already have NFC app plugins, it's capable of loading required plugins on demand and exposes all required APIs (?).

This is news to me. Is there an example somewhere that uses this you can point me to?

AloneLiberty:
> Is there an example somewhere that uses this you can point me to?

Plugins in `applications/main/nfc/plugins/supported_cards` and their loader in `applications/main/nfc/helpers/nfc_supported_cards.c`. I didn't know parsers could do that, I also got a hint from @foxushka recently

nvx:
> > Is there an example somewhere that uses this you can point me to?
> 
> Plugins in `applications/main/nfc/plugins/supported_cards` and their loader in `applications/main/nfc/helpers/nfc_supported_cards.c`. I didn't know parsers could do that, I also got a hint from @Foxushka recently

Might be misunderstanding the requirements here:

In:  `nfc_supported_card_plugin.h`
```
 * To add a new plugin, create a uniquely-named .c file in the `supported_cards` directory
 * and implement at least the parse() function in the NfcSupportedCardsPlugin structure.
 * Then, register the plugin in the `application.fam` file in the `nfc` directory. Use the existing
[...]

PR #3286, opened 2023-12-12
[FL-3579, FL-3601, FL-3714] JavaScript runner

nminaylov:
# What's new

- JS scripting engine based on MJS lib
- Plugin-based JS module system
- Initial set of JS modules: BadUSB, UART, dialog, notification, flipper
- file_browser: support for multiple extensions filter
- Running JS scripts from Apps menu or archive

# Verification 

- Run examples from Apps -> Scripts

# Checklist (For Reviewer)


cowchimp:
so excited about this 🤩

vtrenton:
Just noting - JavaScript - while popular is a very heavy language compared to something like C. The flipper isn't exactly a compute powerhouse, so I'm concerned that there will be a substantial performance penalty integrating a high-level interpreted language like this. thoughts?

jessycormier:
> Just noting - JavaScript - while popular is a very heavy language compared to something like C. The flipper isn't exactly a compute powerhouse, so I'm concerned that there will be a substantial performance penalty integrating a high-level interpreted language like this. thoughts?

I believe there are available tooling to convert Js to C. I wonder if this could be part of a "build process" to enable the faster running of the Js.

vtrenton:
> I believe there are available tooling to convert Js to C. I wonder if this could be part of a "build process" to enable the faster running of the Js.

I think that would be a great idea because it allows devs more familiar with JavaScript to write applications for the Flipper platform which as I understand would be the intent of this PR. Obviously, we'd want to build BEFORE reaching the end device - no JIT stuff - because you do not want to compile C on the flipper. So you'd be able to write JS and have that build to a binary `.fap` file. This PR appears to be adding a JS interpreter on the actual Flipper itself so you can run JS natively on the flipper... I personally think performance issues will start coming up quick going this route.

nvx:
This would be perfect for NFC KDF plugins (#3197)

ghost:
Looking forward to this!

dogtopus:
@vtrenton It's also worth noting that MJS is not any known version of JavaScript. It's a proprietary subset that misses critical features to the point that I'm a bit hesitant to call it JavaScript. This also means no TypeScript support out of the box and no npm package transpilation (not just the heavy modules, no, not even a single line hello world program would be allowed) unless one can figure out how to polyfill.

TBH I don't know how usable this really is other than writing some really simple programs with it, since for more complex programs the development overhead might be even bigger than C especially without TypeScript support.

skotopes (approved):

Bachar-official:
Great news! Wanna see docs about this feature.

PR #3287, opened 2023-12-12
nfc_util: little endian bytes2num functions added

Leptopt1los:
# What's new

- nfc_util_bytes2num_little_endian function added

Problem: 
When developing parsers, i often come across cards in which data is stored in little endian. in nfc_util.c there is a function uint64_t nfc_util_bytes2num(const uint8_t* src, uint8_t len), which reads len bytes at the src pointer and returns them as a number. The problem is that it reads them as big endian, and there is no alternative. Accordingly, for each such case I have to write an extra loop and read everything manually

This feature will solve the problem

# Verification 

- test that the function does not go beyond boundaries
- test the correctness of operation (the result is indeed a representation of the little endian number stored in the src)

# Checklist (For Reviewer)

### DIFF
--- a/lib/nfc/helpers/nfc_util.c
+++ b/lib/nfc/helpers/nfc_util.c
@@ -35,6 +35,19 @@ uint64_t nfc_util_bytes2num(const uint8_t* src, uint8_t len) {
     return res;
 }
 
+uint64_t nfc_util_bytes2num_little_endian(const uint8_t* src, uint8_t len) {
+    furi_assert(src);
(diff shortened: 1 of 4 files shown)


hedger (dismissed):
Please sync changes across all targets (`./fbt TARGET_HW=18`).

Leptopt1los:
> Please sync changes across all targets (`./fbt TARGET_HW=18`).

done

gornekich (dismissed):

skotopes (approved):

PR #3349, opened 2024-01-08
Fix #3347 by implementing a read loop until all file bytes are read.

### LINKED ISSUE #3347: NFC: Files from DESFire cards appear truncated; more reads may be necessary
# Summary

The DESFire card "poller" does not appear to be able to adequately dump more than 472 bytes from any DESFire card file, even if the file's metadata indicates it is larger.

# Background

The Mifare DESFire is a particular kind of ISO 14443-compliant "smart card" which, amongst other things, provides several file-based abstractions for saving and retrieving data on the card. The flipperzero NFC framework has wonderful support for querying and interacting with such cards, including an automatic "poller" which attempts to dump as much information as possible from the card when read in the NFC application.

The issue at hand may be related to how the poller attempts to read file data from the card.

# Theories

Tracking down the source to the entire process, I direct your attention to line 403 in https://github.com/flipperdevices/flipperzero-firmware/blob/7eeb60e17ecd31b964cdc169f60ae314c6e9831d/lib/nfc/protocols/mf_desfire/mf_desfire_poller_i.c#L403

wherein the PDU to read the file data is constructed and sent to the card. The PDU requests to read the _entire_ file in one go, which _may_ be an issue due to limitations in PDU response size.

Next, observe where the _response_ to the file read request is processed in https://github.com/flipperdevices/flipperzero-firmware/blob/7eeb60e17ecd31b964cdc169f60ae314c6e9831d/lib/nfc/protocols/mf_desfire/mf_desfire_i.c#L228

Note that the response handler doesn't check (nor have enough information to check) whether the response data actually includes all of the bytes requested.

It may be that underlying ISO 14443 restrictions prevent read responses from exceeding a certain size, and as such, a state machine might have to be set up to read such files in smaller chunks, using multiple read requests, until satisfied.

### Reproduction

1. Acquire a Mifare DESFire card with any type 00 ("standard") file containing 473 or more bytes of data. Example: San Francisco Bay Area "Clipper" transportation card.
2. Using the "Read" function of the FlipperZero's NFC application, read the card.
3. Save the card data to a file (choose "Save", pick a filename).
4. Plug in the Flipper Zero into a Desktop machine using a USB cable.
5. Using the Flipper Desktop application, download the file from the `/nfc/` directory on the SD Card.
6. Using a text editor, inspect the requisite `.nfc` file.
7. Identify files, metadata, and their contents in the file.
Example:
```
Application 9111f2 File 14 Size: 512
Application 9111f2 File 14: 10 00 00 04 00 00 ...
```
8. Notice that the file's metadata notes the file size as, say, 512 bytes.
9. Count the retrieved bytes in the next line and notice that they only amount to **472** bytes.

### Target

_No response_

### Logs

_No response_

### Anything else?

I discovered this issue while attempting to write a "ride history" display in my Clipper card parser at https://github.com/flipperdevices/flipperzero-firmware/pull/3344

ke6jjj:
# What's new

This fixes #3347 by implementing a read loop which reads the file bit by bit until completed.

# Verification 

Tested with a Mifare DESFire card which exhibited the initial issue.

# Checklist (For Reviewer)

### DIFF
--- a/lib/nfc/protocols/mf_desfire/mf_desfire_poller_i.c
+++ b/lib/nfc/protocols/mf_desfire/mf_desfire_poller_i.c
@@ -301,24 +301,32 @@ MfDesfireError mf_desfire_poller_read_file_data(
     size_t size,
     MfDesfireFileData* data) {
     furi_assert(instance);
-
-    bit_buffer_reset(instance->input_buffer);
-    bit_buffer_append_byte(instance->input_buffer, MF_DESFIRE_CMD_READ_DATA);
-    bit_buffer_append_byte(instance->input_buffer, id);
-    bit_buffer_append_bytes(instance->input_buffer, (const uint8_t*)&offset, 3);
-    bit_buffer_append_bytes(instance->input_buffer, (const uint8_t*)&size, 3);
-
+    size_t cur_offset, read_size, num_read;
     MfDesfireError error;
 
-    do {
+    simple_array_init(data->data, size);
+    uint8_t* data_buf = (uint8_t*)simple_array_get_data(data->data);
+
+    for(cur_offset = offset; cur_offset < size; cur_offset += num_read) {
+        read_size = size - cur_offset;
+
+        bit_buffer_reset(instance->input_buffer);
+        bit_buffer_append_byte(instance->input_buffer, MF_DESFIRE_CMD_READ_DATA);
+        bit_buffer_append_byte(instance->input_buffer, id);
+        bit_buffer_append_bytes(
+            instance->input_buffer, (const uint8_t*)&cur_offset, 3); // This is not endian safe.
+        bit_buffer_append_bytes(
+            instance->input_buffer, (const uint8_t*)&read_size, 3); // This is not endian safe.
+
         error = mf_desfire_send_chunks(instance, instance->input_buffer, instance->result_buffer);
 
         if(error != MfDesfireErrorNone) break;
 
-        if(!mf_desfire_file_data_parse(data, instance->result_buffer)) {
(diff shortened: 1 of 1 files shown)


ke6jjj:
This issue likely also exists in the "read file value" and "read file records" code paths. I've only fixed the "read normal file" path.

ke6jjj:
As noted in #3347, I've pushed in some more fixes to this PR.

One fixes an off-by-one counting bug which was causing received responses to be thrown away when they actually were perfectly fine.

The other is a change to the file read polling loop such that it requests no more bytes than can actually fit into the receive buffer.

lib/nfc/protocols/mf_desfire/mf_desfire_poller_i.c
```diff
@@ -301,24 +303,35 @@ MfDesfireError mf_desfire_poller_read_file_data(
     size_t size,
     MfDesfireFileData* data) {
     furi_assert(instance);
+    size_t cur_offset, read_size, num_read;
+    MfDesfireError error;
```
gornekich, line 307: Compiler thinks that we might not get into for loop, where we update error variable. Please, initialize this variable here

gornekich (changes requested):

skotopes:
Un-draft when ready

lib/nfc/protocols/mf_desfire/mf_desfire_poller_i.c
ke6jjj: No problem. I'll look at it this week.

gornekich:
Fixed arrive in dev

PR #3350, opened 2024-01-09
Added a text input that only accepts full numbers (int)

leedave:
# What's new

- I've added a new Keyboard input. It is similar to text_input and byte_input. The purpose of this one is to only allow users to input a number, nothing else. This can be very handy in app designs, as the user input is limited to expected characters.

# Verification 

- #include <gui/modules/int_input.h> in a fap application the same way you would include text_input.h

# Checklist (For Reviewer)


hedger (dismissed):
Please also include the new header in "sdk_headers" for gui module. And update API CSV's for both f7 and f18 afterwards.

leedave:
@hedger Thank you for the feedback. 
Added the requested sdk_headers and CSV entries

leedave:
@hedger are these fail messages normal or am I expected to do something to fix this?

skotopes:
Sorry for delay with your PR. 
Can you also provide example app and rename `IntInput` to `NumberInput`, `int_input` to `number_input`?

leedave:
@skotopes 
Ok, tried my best to keep it as expected. Let me know if something isn't right. 

- The keyboard is renamed
- There is an example app in **applications/examples/example_number_input**

leedave:
https://github.com/flipperdevices/flipperzero-firmware/assets/5908330/7530c74e-8b17-4775-b99e-4c433c50c6b0

skotopes:
```[Switching to Thread 536956528]
0x0801147c in __furi_crash_implementation () at furi/core/check.c:163
163	        RESTORE_REGISTERS_AND_HALT_MCU(debug);
(gdb) bt
#0  0x0801147c in __furi_crash_implementation () at furi/core/check.c:163
#1  0x0807a5cc in view_dispatcher_free (view_dispatcher=0x20014de8) at applications/services/gui/view_dispatcher.c:26
#2  0x200154a6 in example_number_input_free (app=app@entry=0x20015a00) at applications/examples/example_number_input/example_number_input.c:45
#3  0x2001551e in example_number_input (p=<optimized out>) at applications/examples/example_number_input/example_number_input.c:69
#4  0x080548a2 in flipper_application_thread (context=0x20014b20) at lib/flipper_application/flipper_application.c:224
#5  0x08013d6e in furi_thread_body (context=0x20014d78) at furi/core/thread.c:91
#6  0x08013d1a in furi_thread_catch () at furi/core/thread.c:62
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

crashing on exit because some view was not released from view dispatcher

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,389 @@
...
+    furi_assert(number_input);
+    with_view_model(
+        number_input->view,
+        NumberInputModel * model,
+        {
+            model->header = "";
+            model->selected_row = 0;
+            model->selected_column = 0;
+            model->clear_default_text = false;
+            model->text_buffer = "";
```
skotopes: I got feeling that this part was copied from another place. We are not proud with it )

General concept for view modules is that they should own their own buffer and then provide it in callback call.

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,389 @@
...
+    uint8_t selected_column;
+} NumberInputModel;
+
+static const uint8_t keyboard_origin_x = 7;
+static const uint8_t keyboard_origin_y = 31;
+static const uint8_t keyboard_row_count = 2;
+static const uint8_t enter_symbol = '\r';
+static const uint8_t backspace_symbol = '\b';
+
+static const NumberInputKey keyboard_keys_row_1[] = {
```
skotopes, line 39: Since we talking about numbers(int) you may want to provide negative sign too.
Also it will be nice if limits(MIN/MAX) and maybe int size be adjustable

skotopes (changes requested):

skotopes:
Un-draft when ready

skotopes:
also please fix build issues (API mismatch)

skotopes:
@leedave need help?

leedave:
@skotopes 
Can you give me some more hints on the 
`view modules should own their own buffer` issue. 
Tbh I learn how code for Flipper should look by studying the fw you provide, as there is no official developer documentation. I got the basic setup from the other two keyboards (text & hex). If you say you're not proud of what you guys did back then and ask for an update, I'm not sure where I can find something built in the way you want to see it.

skotopes:
@leedave change `model->text_buffer` type to `FuriString` and allocate it on your module allocation

leedave:
@skotopes 
How did you do the check, to see if some views were not removed from view_dispatcher. I added a code update, but lack a way of testing the result

leedave:
So after a long wait, I finally managed to add your requests

- `model->text_buffer` changed to FuriString
- removed number digits length option
- replaced with settings for min/max value allowed
- added possibility to input negative numbers, added button if set to true when loading module

targets/f7/api_symbols.csv
```diff
@@ -20,6 +20,7 @@ Header,+,applications/services/gui/modules/file_browser.h,,
 Header,+,applications/services/gui/modules/file_browser_worker.h,,
 Header,+,applications/services/gui/modules/loading.h,,
```
skotopes, line 21: Bump minor api version

targets/f18/api_symbols.csv
```diff
@@ -13,6 +13,7 @@ Header,+,applications/services/gui/icon_i.h,,
 Header,+,applications/services/gui/modules/button_menu.h,,
 Header,+,applications/services/gui/modules/button_panel.h,,
```
skotopes, line 14: Bump minor api version

applications/services/gui/modules/number_input.h
```diff
@@ -0,0 +1,75 @@
...
+ * @param      max_value          Max number value
+ * @param      useSign            Allows negative values
+ * @param      clear_default_text clear previous entry
+ */
+
+void number_input_set_result_callback(
+    NumberInput* number_input,
+    NumberInputCallback input_callback,
+    void* callback_context,
+    FuriString* text_buffer,
```
skotopes: There is no need for this module to have `FuriString* text_buffer`.

applications/services/gui/modules/number_input.h
```diff
@@ -0,0 +1,75 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** Number input anonymous structure  */
+typedef struct NumberInput NumberInput;
+
+/** callback that is executed on save button press */
+typedef void (*NumberInputCallback)(void* context);
```
skotopes: Pass number to callback

applications/services/gui/modules/number_input.h
```diff
@@ -0,0 +1,75 @@
...
+
+/** callback that is executed when byte buffer is changed */
+typedef void (*NumberChangedCallback)(void* context);
```
skotopes: Pass number or text to callback.
Also what's meaning of this callback? does it works as validator? In that case it should return something to signal validation results.

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+
+typedef struct {
+    const char text;
+    const uint8_t x;
+    const uint8_t y;
+} NumberInputKey;
+
+typedef struct {
+    const char* header;
+    FuriString* text_buffer;
```
skotopes, line 19: Do you need text? At the end you are working with the numbers

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+    bool clear_default_text;
+    int32_t max_value;
+    int32_t min_value;
+    bool useSign;
+    char sign;
```
skotopes: Too many redundant things do you really need them?

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+    NumberInputCallback callback;
+    void* callback_context;
+    int8_t selected_row;
```
skotopes: Can it be negative? if not then size_t. In general there is no sense in using uintX_t types where specialized one exists

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+    const uint8_t text_y = 25;
+
+    elements_slightly_rounded_frame(canvas, 6, 14, 116, 15);
+
+    FuriString* text = model->text_buffer;
+    canvas_draw_str(canvas, text_x, text_y, furi_string_get_cstr(text));
+}
+
+static void number_input_backspace_cb(NumberInputModel* model) {
+    uint8_t text_length = model->clear_default_text ? 1 :
```
skotopes: size_t please

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+                                                      furi_string_utf8_length(model->text_buffer);
+    if(text_length > 0) {
+        furi_string_set_strn(
+            model->text_buffer, furi_string_get_cstr(model->text_buffer), text_length - 1);
+    }
+}
+
+static char* int32_to_string(int32_t value) {
```
skotopes: FuriString got everything you need, why fallback to C-Strings?

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+    } else {
+        model->selected_column = 0;
+    }
+    const NumberInputKey* keys = number_input_get_row(model->selected_row);
+    if(keys[model->selected_column].text == sign_symbol && !model->useSign) {
+        model->selected_column++;
+    }
+}
+
+static void prevent_to_large_number(NumberInputModel* model) {
```
skotopes: In terms of UI/UX this behavior is quite painful: it irreversibly modifies previous input.
Best option is to ignore input. Other systems also have visual/audial alert, but we don't use them for input.

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+    // Added in prevent large, as it would block the input of small positive numbers
+    if(model->sign == '-' &&
+       strtol(furi_string_get_cstr(model->text_buffer), NULL, 10) < model->min_value) {
+        char* str = int32_to_string(model->min_value);
+        furi_string_set_str(model->text_buffer, str);
+        free(str);
+    }
+}
+
+static void prevent_to_small_number(NumberInputModel* model) {
```
skotopes: same

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+    if(strtol(furi_string_get_cstr(model->text_buffer), NULL, 10) < model->min_value) {
+        char* str = int32_to_string(model->min_value);
+        furi_string_set_str(model->text_buffer, str);
+        free(str);
+    }
+}
+
+static void number_input_sign(NumberInputModel* model) {
+    if(model->sign == '-') {
+        model->sign = '+';
```
skotopes: There is a model and there is a representation, what you doing is mixing representation into model.

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+ *
+ * @param      canvas  The canvas
+ * @param      _model  The model
+ */
+static void number_input_view_draw_callback(Canvas* canvas, void* _model) {
+    NumberInputModel* model = _model;
+    uint8_t text_length = model->text_buffer ? furi_string_utf8_length(model->text_buffer) : 0;
+    UNUSED(text_length);
+
+    canvas_clear(canvas);
```
skotopes: clear is done by canvas before each draw call. there is no need clear in your draw call callback

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+    canvas_set_color(canvas, ColorBlack);
```
skotopes: same

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+
+    number_input_draw_input(canvas, model);
+
+    canvas_set_font(canvas, FontSecondary);
+    canvas_draw_str(canvas, 2, 9, model->header);
+    canvas_set_font(canvas, FontKeyboard);
+    // Draw keyboard
+    for(uint8_t row = 0; row < keyboard_row_count; row++) {
```
skotopes: size_t please

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+    furi_assert(number_input);
+    with_view_model(
+        number_input->view,
+        NumberInputModel * model,
+        {
+            model->header = "";
+            model->selected_row = 0;
+            model->selected_column = 0;
+            model->clear_default_text = false;
+            model->text_buffer = furi_string_alloc();
```
skotopes: something really bad happens here

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+
+    return number_input;
+}
+
+void number_input_free(NumberInput* number_input) {
+    furi_assert(number_input);
+    with_view_model(
+        number_input->view,
+        NumberInputModel * model,
+        { furi_string_free(model->text_buffer); },
```
skotopes: and here

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,493 @@
...
+    int32_t max_value,
+    bool useSign,
+    bool clear_default_text) {
+    with_view_model(
+        number_input->view,
+        NumberInputModel * model,
+        {
+            model->callback = callback;
+            model->callback_context = callback_context;
+            model->text_buffer = text_buffer;
```
skotopes: and here

applications/examples/example_number_input/example_number_input.c
```diff
@@ -0,0 +1,77 @@
...
+    view_dispatcher_add_view(
+        app->view_dispatcher,
+        ExampleNumberInputViewIdNumberInput,
+        number_input_get_view(app->number_input));
+    app->show_number = example_number_input_show_number_alloc();
+    view_dispatcher_add_view(
+        app->view_dispatcher,
+        ExampleNumberInputViewIdShowNumber,
+        example_number_input_show_number_get_view(app->show_number));
+    app->view_stack = view_stack_alloc();
```
skotopes: I don't see you using it. Remove anything that you don't use.

applications/examples/example_number_input/scenes/example_number_input_scene_input_number.c
```diff
@@ -0,0 +1,48 @@
...
+    view_dispatcher_send_custom_event(
+        app->view_dispatcher, ExampleNumberInputCustomEventTextInput);
+}
+
+void example_number_input_scene_input_number_on_enter(void* context) {
+    furi_assert(context);
+    ExampleNumberInput* app = context;
+    NumberInput* number_input = app->number_input;
+    size_t enter_name_length = 5;
+    char* str = "Enter a number (0 - 9999)";
```
skotopes: Numbers are clearly off

applications/examples/example_number_input/scenes/example_number_input_scene_input_number.c
```diff
@@ -0,0 +1,48 @@
...
+    const char* constStr = str;
+    number_input_set_header_text(number_input, constStr);
+    number_input_set_result_callback(
+        number_input,
+        example_number_input_scene_input_number_callback,
+        context,
+        app->text_buffer,
+        -128,
```
skotopes: this

skotopes (changes requested):

skotopes:
Sorry for keeping you in limbo for such a long time.

This PR was bugging me a lot when I was trying to review it previous time, It's quite bad in terms of the code and clearly requires more work before we will be able to merge it.

I've highlighted some major issues, un-draft this PR when you fix it. Also feel free to ask if you need any help.

applications/services/gui/modules/number_input.h
leedave: You're right, this is not needed. Will be removed

applications/services/gui/modules/number_input.c
leedave: Yes, the input is a keyboard and therefore it creates a string, which is also shown on screen. 
When I follow your change requests to return a number in callback, I'd convert this string and remove it when the module is closed.

applications/services/gui/modules/number_input.c
leedave: Please explain, for me these are not redundant, they are all needed. 
max_value => highest number allowed
min_value => lowest number allowed
sign => + or - depending if the number is positive or negative
useSign => if true, will show the button that allows changing from positive to negative numbers 
(In my use cases so far I did not want negative numbers and therefore don't want that button, this was a change request from your side)

applications/services/gui/modules/number_input.c
leedave: Your right, as it cannot be negative, there's no need for a sign. 

I'm having trouble figuring out why **size_t** is preferred over **uint8_t**. Can you explain this? From what I can gather it's just uint8_t is alway 8-Bit while the Bit-size of size_t are platform dependent.

applications/services/gui/modules/number_input.c
leedave: ok

applications/services/gui/modules/number_input.c
leedave: Ok, since I always use this to manipulate a FuriString, I'll attempt to use furi_string_printf directly in the function in place of snprintf.

applications/services/gui/modules/number_input.c
leedave: It is the same line... or are you refering to something else?

applications/services/gui/modules/number_input.c
skotopes: It's all about machine word size and architecture. If you have 32bit Arm MCU then it will operate with 32bit wide registers. So basically you place your 8bit value to 32bit register do things and then transfer it back. That's not going to be faster then doing same thing with 32bit wide variable.

Also 8 bit is kinda small by itself, make no sense to put such constrains on software that meant to be portable and future proof.

Keep

applications/services/gui/modules/number_input.c
skotopes: default color on draw call is black

leedave:
@skotopes 
Thank you for taking the time to review my code. Looks like I still have a lot to learn, but I'll stay at it. 

As chance has it, just like last time your review collided with the first day of my vacation. Meaning I can't update it before I get back. I did some of the small changes already, the rest will need to wait.

skotopes:
@leedave no problem, I'll wait ;-)

applications/examples/example_number_input/scenes/example_number_input_scene_input_number.c
leedave: The option is that you can set your own text in your app. But I will adjust this, so that the text transferred is dynamic matching the min/max numbers.

targets/f7/api_symbols.csv
leedave: Not sure if I'm doing this correctly... I assume I need to run `.\fbt` , fix entries with ? and then commit?

leedave:
@skotopes 
I've done some refactoring to solve the issues you mentioned and some more. 

Main additions
- Abilitity to adjust min and max values in the demo app for easier testing of edge-cases
- Invalid number inputs are ignored
- In some cases you need to be able to write an invalid number to be able to finish off with a valid number later (min: 50, type in 9 then 0 for 90). A Lock button replaces the save button in this case to symbolize invalid input and prevent saving.
- Back button no longer saves entry, but loads back the original value
- Removed sign & useSign, as no longer needed with min/max settings
- replaced all const char* strings in module with furiString. Only used in module and destroyed on exit. 
- Callback returns number, no longer uses string

I hope this fits better. Let me know if something is still not good.

skotopes:
image
tehe

applications/examples/example_number_input/scenes/example_number_input_scene_input_max.c
```diff
@@ -0,0 +1,49 @@
...
+        app->view_dispatcher, ExampleNumberInputCustomEventTextInput);
+}
+
+void example_number_input_scene_input_max_on_enter(void* context) {
+    furi_assert(context);
+    ExampleNumberInput* app = context;
+    NumberInput* number_input = app->number_input;
+
+    int32_t max = 2147483646; //has no effect if larger
+    static char str[50];
```
skotopes: why static?

applications/examples/example_number_input/example_number_input.c
```diff
@@ -0,0 +1,75 @@
...
+    furi_assert(context);
+    ExampleNumberInput* app = context;
+    return scene_manager_handle_custom_event(app->scene_manager, event);
+}
+
+static ExampleNumberInput* example_number_input_alloc() {
+    ExampleNumberInput* app = malloc(sizeof(ExampleNumberInput));
+    app->gui = furi_record_open(RECORD_GUI);
+    app->notification = furi_record_open(RECORD_NOTIFICATION);
+    app->text_buffer = furi_string_alloc();
```
skotopes: dead code?

applications/examples/example_number_input/example_number_input.c
```diff
@@ -0,0 +1,75 @@
...
+    app->current_number = 5;
+    app->min_value = -128;
+    app->max_value = 2345;
+    furi_string_set(app->text_buffer, "5");
+    //Turn backlight on, believe me this makes testing your app easier
+    notification_message(app->notification, &sequence_display_backlight_on);
```
skotopes: not needed anymore

applications/examples/example_number_input/example_number_input.c
```diff
@@ -0,0 +1,75 @@
...
+
+int32_t example_number_input(void* p) {
+    UNUSED(p);
+    ExampleNumberInput* app = example_number_input_alloc();
+
+    view_dispatcher_attach_to_gui(app->view_dispatcher, app->gui, ViewDispatcherTypeFullscreen);
+
+    scene_manager_next_scene(app->scene_manager, ExampleNumberInputSceneShowNumber);
+
+    furi_hal_power_suppress_charge_enter();
```
skotopes: why?

applications/examples/example_number_input/scenes/example_number_input_scene_input_max.c
```diff
@@ -0,0 +1,49 @@
...
+    view_dispatcher_send_custom_event(
+        app->view_dispatcher, ExampleNumberInputCustomEventTextInput);
+}
+
+void example_number_input_scene_input_max_on_enter(void* context) {
+    furi_assert(context);
+    ExampleNumberInput* app = context;
+    NumberInput* number_input = app->number_input;
+
+    int32_t max = 2147483646; //has no effect if larger
```
skotopes: INT32_MAX

applications/examples/example_number_input/scenes/example_number_input_scene_input_min.c
```diff
@@ -0,0 +1,49 @@
...
+    view_dispatcher_send_custom_event(
+        app->view_dispatcher, ExampleNumberInputCustomEventTextInput);
+}
+
+void example_number_input_scene_input_min_on_enter(void* context) {
+    furi_assert(context);
+    ExampleNumberInput* app = context;
+    NumberInput* number_input = app->number_input;
+
+    int32_t min = -2147483647; //has no effect if smaller
```
skotopes: INT32_MIN

applications/examples/example_number_input/scenes/example_number_input_scene_input_min.c
```diff
@@ -0,0 +1,49 @@
...
+    static char str[50];
```
skotopes: why static?

applications/examples/example_number_input/scenes/example_number_input_scene_input_min.c
```diff
@@ -0,0 +1,49 @@
...
+    } else if(event.type == SceneManagerEventTypeCustom) {
+        scene_manager_previous_scene(app->scene_manager);
+        return true;
+    }
+    return consumed;
+}
+
+void example_number_input_scene_input_min_on_exit(void* context) {
+    ExampleNumberInput* app = context;
+    UNUSED(app);
```
skotopes: UNUSED(context)

applications/examples/example_number_input/scenes/example_number_input_scene_input_number.c
```diff
@@ -0,0 +1,48 @@
...
+    view_dispatcher_send_custom_event(
+        app->view_dispatcher, ExampleNumberInputCustomEventTextInput);
+}
+
+void example_number_input_scene_input_number_on_enter(void* context) {
+    furi_assert(context);
+    ExampleNumberInput* app = context;
+    NumberInput* number_input = app->number_input;
+
+    static char str[50];
```
skotopes: why static?

applications/examples/example_number_input/views/example_number_input_show_number.c
```diff
@@ -0,0 +1,150 @@
...
+
+void example_number_input_show_number_exit(void* context) {
+    furi_assert(context);
+}
+
+void example_number_input_show_number_enter(void* context) {
+    furi_assert(context);
+}
+
+ExampleNumberInputShowNumber* example_number_input_show_number_alloc() {
```
skotopes: you are inventing DialogEx here

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,487 @@
...
+    NumberInputCallback callback;
+    void* callback_context;
+    size_t selected_row;
+    size_t selected_column;
+} NumberInputModel;
+
+static const size_t keyboard_origin_x = 7;
+static const size_t keyboard_origin_y = 31;
+static const size_t keyboard_row_count = 2;
+static const size_t enter_symbol = '\r';
```
skotopes: those are chars

applications/services/gui/modules/number_input.c
```diff
@@ -0,0 +1,487 @@
...
+ * @param      canvas  The canvas
+ * @param      _model  The model
+ */
+static void number_input_view_draw_callback(Canvas* canvas, void* _model) {
+    NumberInputModel* model = _model;
+
+    number_input_draw_input(canvas, model);
+
+    canvas_set_font(canvas, FontSecondary);
+    canvas_draw_str(canvas, 2, 9, furi_string_get_cstr(model->header));
```
skotopes: model->header can be null: either allocate on start or add check

targets/f18/api_symbols.csv
```diff
@@ -13,6 +13,7 @@ Header,+,applications/services/gui/icon_i.h,,
 Header,+,applications/services/gui/modules/button_menu.h,,
```
skotopes, line 13: Bump API version minor

skotopes (changes requested):

skotopes:
See comments. Un-draft when ready.

applications/examples/example_number_input/example_number_input.c
leedave: I generally hate it when loading an app via fbt and the backlight is not on. Its hard to see whats on screen without the light and it requires a keypress to activate. If the keypress already does something, you miss the initial screen. 

I'll leave it in there, but comment it out because you don't want it.

leedave: Ok I just noticed this was recently changed, it now turns on without the command... so yeah, not needed anymore, will remove

applications/examples/example_number_input/example_number_input.c
leedave: It's a habbit, as in some apps I want do things with the LED. Not needed here, so I will remove it.

targets/f18/api_symbols.csv
leedave: Please double check if I'm doing this correctly

leedave:
@skotopes 
Your requested updates are made, ready for review.

targets/f7/api_symbols.csv
skotopes: Change `Version,+,68.1,,` to `Version,+,68.2,,`

applications/services/gui/modules/number_input.h
```diff
@@ -0,0 +1,68 @@
...
+ * @param      max_value          Max number value
+ */
+
+void number_input_set_result_callback(
+    NumberInput* number_input,
+    NumberInputCallback input_callback,
+    void* callback_context,
+    int32_t current_number,
+    int32_t min_value,
+    int32_t max_value);
```
skotopes, line 58: set max value to 0 and then try to input negative number

skotopes (changes requested):

skotopes:
@leedave you are close to the finishing line, just a little bit more ;-)

leedave:
Thanks again for making me a better programmer

applications/examples/example_number_input/example_number_input.c
```diff
@@ -0,0 +1,79 @@
...
+        number_input_get_view(app->number_input));
+
+    app->dialog_ex = dialog_ex_alloc();
+    view_dispatcher_add_view(
+        app->view_dispatcher,
+        ExampleNumberInputViewIdShowNumber,
+        dialog_ex_get_view(app->dialog_ex));
+
+    app->current_number = 5;
+    app->min_value = INT32_MIN;
```
skotopes, line 41: I've made couple changes in example code:
- Simplified it as much as possible
- Changed min max values to int32_t specification

Looks like you incorrectly handling INT32_MIN and INT32_MAX in number_input module. Try to input something bigger or smaller that INT32_MIN or INT32_MAX.

skotopes (changes requested):

skotopes:
Un-draft when fixed

applications/examples/example_number_input/example_number_input.c
leedave: There's a reason I used +/- 1 on the INT32_MIN/INT32_MAX specifications. 

You removed these lines and replaced it with the MIN/MAX. That's why it no longer works. 
`
int32_t max = INT32_MAX - 1; // 2147483646, limit has no effect if larger
int32_t min = INT32_MIN + 1; // -2147483647 limit has no effect if smaller
`

The module needs to make an int of the input string and compare it with the min/max values. If using the max allowed int32 values the number resets if bigger/smaller and the comparison doesn't work. 

If the solution of reducing the limit by 1 is not ok, please advise me how I can fix this in a better way.

skotopes: Normally you have to check `errno` for any errors that happened in conversion. However `errno` use is prohibited because it is not thread safe. Another options is to use `strtoll` and then compare int64_t values.

I've pushed bunch of fixes that solves all remaining issues.

skotopes (approved):

skotopes:
That was quite a journey, couple things for you to take away:
- Read documentation
- Plan object ownership and allocate/release cycle
- Keep things simple
- Test

applications/examples/example_number_input/example_number_input.c
leedave: Great, thank you for the update and solution.

leedave:
Yes that was a journey, but I do appreciate the help as at the time I started developing things for Flipper there was no documentation I could read, only other code from the FW I could learn from. 

I've just noticed that in the mean time documentation has started appearing ... and I'm surprised to see that there's actually a page related to [this Pull Request] (https://developer.flipper.net/flipperzero/doxygen/md__2github_2workspace_2applications_2examples_2example__number__input_2_read_me.html)

skotopes:
As long as you document your code in doxygen format it will appear in developer documentation

PR #3356, opened 2024-01-10
NFC: system dict skip when user dict is skipped fix

Leptopt1los:
# Issue
At the moment, when performing a dictionary attack on MF Classic, when you try to skip the user dictionary, the system dictionary is also skipped after it. This bug is not reproduced if you attach the card to the flipper after reading has started, or if you remove it and reapply it while reading with a user dictionary.

The bug is due to the fact that when an NfcCustomEventDictAttackSkip event occurs with the DictAttackStateUserDictInProgress state, instance->nfc_dict_context.is_card_present is false (see line 226 on nfc_scene_mf_classic_dict_attack.c)

##Issue verification:
- Create manually MF Classic 1k 4byte UID dump
- Write it to magic card
- Try to read it
- Skip user dict

System dict will also be skipped

# What's new

The issue described above is fixed

# Verification 

Same as issue verification. After user dict skipping, system dict attack must be started

# Checklist (For Reviewer)

### DIFF
--- a/applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
+++ b/applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
@@ -173,6 +173,7 @@ void nfc_scene_mf_classic_dict_attack_on_enter(void* context) {
 
     instance->poller = nfc_poller_alloc(instance->nfc, NfcProtocolMfClassic);
     nfc_poller_start(instance->poller, nfc_dict_attack_worker_callback, instance);
+    instance->nfc_dict_context.is_card_present = true;
 }
 
 static void nfc_scene_mf_classic_dict_attack_notify_read(NfcApp* instance) {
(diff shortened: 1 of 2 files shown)


gornekich:
Hello @Leptopt1los 
Thanks for PR. I reproduced the bug but the problem is in different place. `instance->nfc_dict_context.is_card_present` variable should be updated on MFC poller event, but event is not emitted because of incorrect MFC poller state initialization.

Please, consider attached patch. If it works fine for you, please apply this patch to this PR and we will merge it.
[mfc_poller_fix_patch.txt](https://github.com/flipperdevices/flipperzero-firmware/files/13902994/mfc_poller_fix_patch.txt)
Applying patch: `git apply <path_to_patch_file>`

gornekich (changes requested):
Not completely correct fix

gornekich (approved):
Thanks

skotopes (approved):

PR #3360, opened 2024-01-11
fbt/ufbt: Ensure POSIX paths are passed to GDB on all platforms

CookiePLMonster:
# What's new

GDB heavily dislikes forward slashes from Windows paths and strips them internally instead of normalizing them. Account for this by passing POSIX paths explicitly.

Fixes an issue where GDB was given paths with forward slashes and later stripped them internally, leaving to errors like
`OSError: Error reading file 'H:devflipper-zero.ufbtcurrentscriptsdebug/STM32WB55_CM4.svd': failed to load external entity "H:devflipper-zero.ufbtcurrentscriptsdebug/STM32WB55_CM4.svd"`

Couple of notes:
* This is the first time I see a SCons library in the wild, but far as I can tell, the use of `SCons.Node.FS.ProxyEntry` is correct here.
* I have not been able to test the entire debugging flow, as only later I realized that debugging through USB without a dev board is not possible. However, I still wanted to fix the issue I spotted. Therefore, my list of fixes may not be exhaustive.

# Verification 

* Attempt to debug via `ufbt debug` on Windows.

# Checklist (For Reviewer)


CookiePLMonster:
Note - I got my devboard today and ufbt debugging over blackmagic works fine without this PR too - therefore, I'm not sure how severe those issues truly are. I assume Blackmagic's debugging is not actually equivalent to `ufbt debug` but rather `ufbt blackmagic`.

hedger:
@CookiePLMonster could you please try out the current implementation?

CookiePLMonster:
These changes look good but I'll try it soon and get back to you!

skotopes (approved):

PR #3363, opened 2024-01-13
NFC: Custom UID entry when adding manually

TollyH:
# What's new

- Implemented a UID entry field for all the NFC types under the "Add Manually" submenu

# Verification 

- Go to NFC > Add Manually and choose any of the options with the new UID entry field (i.e. not the two generic NFC-A options)
- Enter a UID, and the UID of newly created card should match the entered UID

# Checklist (For Reviewer)


lib/nfc/helpers/nfc_data_generator.c
```diff
@@ -25,18 +26,15 @@ static const uint8_t default_data_ntag215_216[] = {0x03, 0x00, 0xFE};
...
 static const uint8_t default_config_ntag_i2c[] = {0x01, 0x00, 0xF8, 0x48, 0x08, 0x01, 0x00, 0x00};
 
-static void nfc_generate_mf_ul_uid(uint8_t* uid) {
-    uid[0] = NXP_MANUFACTURER_ID;
-    furi_hal_random_fill_buf(&uid[1], 6);
-    uid[3] |= 0x01; // To avoid forbidden 0x88 value
-    // I'm not sure how this is generated, but the upper nybble always seems to be 8
-    uid[6] &= 0x0F;
-    uid[6] |= 0x80;
-}
```
gornekich: Removing these UID construction rules may result in incorrect Mf tag emulation. At some point these rules should be preserved in UID generation

gornekich (commented):
Hello, thanks for PR.

I think editing UID is useful feature. But there are some UID constraction rules that all Mifare tags follow, and we should help user to build correct UID.

Personally I think it would be better to leave UID generation as it was and let user edit UID from saved card menu. Let me ask out design team what they think about it and I will come back to you with more details.

gornekich:
I talked to design team and we came up with a solution. Please have a look on a picture attached.

Editing UID is not a common action. That's why it's not correct to make user enter UID manually on every card generation. Adding explicit menu option in following menu is better solution.

From implementation point of view, we should add "Change UID" option in nfc_protocol_support_scene_read_menu_on_enter() function only if we came to this scene after NfcSceneGenerateInfo. scene_manager_has_previous_scene() must be applied.

Please, let me know if you will make this changes by yourself or if you need more information
image

TollyH:
Thank you for the feedback. I can start working on those changes myself and will undraft the PR when complete.

TollyH:
I've added that menu option and restored the original random UID feature as discussed. In order to account for the UID being changed after the initial generation of the card data, I altered the behaviour of `mf_classic_set_uid` and `mf_ultralight_set_uid` so that when the UID is updated, the relevant data in the card is also updated instead of only modifying the iso14443_3a UID.

gornekich (dismissed):

gornekich:
Thanks!

skotopes (approved):

PR #3365, opened 2024-01-15
[FL-3661] Troika layout fixes

Astrrra:
# What's new

- Support for different Troika layouts and sub-layouts (currently, layout 0xE and sub-layouts 0x3 and 0x5 are supported). If the parser encounters an unsupported layout/sub-layout, it will exit to avoid showing garbage data.
- Fixes for 4K Troika cards (see Verification section for more info)
- Enabling Debug in system settings will result in the parser now showing additional data in results: layout, sub-layout, and data block number. It will also disable the check whether layout is supported or not to simplify diagnosing unknown cards that don't work correctly.  

# Verification 

- Check parsing of 1K and 4K cards with layout 0xE and sub-layouts 0x3 and 0x5 (4 variations in total). 
- 4K Troika cards seem to have had a bug with the wrong data block being used, at least all the 4K cards I have access to (with the same layout/sub-layout) have a different data block. This makes me a bit suspicious, therefore further experimentation is required on a bigger sample size:
    - Check all available 4K Troika cards with Debug **disabled**, verify that they either show correct data or aren't parsed at all (due to yet unsupported layout). If any of the cards gets parsed but shows incorrect data, please provide a full dump of the card for further testing. **Ideally, that shouldn't happen.**
    - Check all available 4K Troika cards with Debug **enabled**.  All the cards should trigger the parser and show _some_ results, possibly wrong. Note all the cards with wrong results and provide dumps of them as well, this will help with support for more layouts. **This is OK and expected, as we don't yet support all the layouts**.

# Checklist (For Reviewer)


gornekich (dismissed):

applications/main/nfc/plugins/supported_cards/troika.c
```diff
@@ -76,6 +90,132 @@ static bool troika_get_card_config(TroikaCardConfig* config, MfClassicType type)
...
+    uint8_t start_block_num,
+    TroikaLayout layout,
+    TroikaSubLayout sub_layout) {
+    furi_assert(data);
+
+    // In layout 0x3 balance in bits 188:209 ( from sector start, length 22).
+    // In layout 0x5 balance in bits 165:185 ( from sector start, length 20).
+
+    if(layout == TroikaLayoutE && sub_layout == TroikaSublayout3) {
+        const uint8_t* temp_ptr = &data->block[start_block_num + 1].data[7];
```
hedger: use a  common function in both branches?

applications/main/nfc/plugins/supported_cards/troika.c
```diff
@@ -76,6 +90,132 @@ static bool troika_get_card_config(TroikaCardConfig* config, MfClassicType type)
...
+static TroikaLayout troika_get_layout(const MfClassicData* data, uint8_t start_block_num) {
+    furi_assert(data);
+
+    // Layout is stored in byte 6 of block, length 4 bits (bits 52 - 55), second nibble.
+    const uint8_t* layout_ptr = &data->block[start_block_num].data[6];
+    const uint8_t layout = (*layout_ptr & 0x0F);
+
+    TroikaLayout result = TroikaLayoutUnknown;
+    switch(layout) {
+    case 0x2:
```
hedger: pointless - enum values are equal to switch cases, we can use them as-is.
So, just check for known values and cast to enum, return Unknown otherwise

gornekich (dismissed):

hedger (approved):

Issue #3380, opened 2024-01-17
Furi: Apps can easily starve the timer thread, breaking input processing


CookiePLMonster:
### Describe the bug.

# Intro

The following example app I created to familiarize myself with the Flipper Zero ends up breaking the input processing. 90% of the time it is not possible to terminate this app gracefully with the Back button, and Flipper remains unresponsive until rebooted (with Left+Back):
https://gist.github.com/CookiePLMonster/a95e7974b287cbc11714e234a477a871

While this app doesn't follow the conventional structure of a Flipper app, it is (to the best of my knowledge) valid and attempts to keep a steady ~60FPS refresh rate on the rendered checkerboard content. This is done by ensuring that the View Dispatcher's tick function executes 60 times per second, and updates the View in order to trigger a Draw callback on it consistently.

The issue is that with this setup, the app appears to consistently starve input processing, as proven by the `exit_app` log message not appearing at all under the broken scenario. As outlined in the code, any attempts to add extra logging in the app shift timing just enough for the bug not to manifest.

# Breakdown

Together with @kbembedded, we've troubleshooted this issue on the Flipper Zero Discord to the best of our abilities and came to the following conclusions:
* Under the broken scenario, the firmware gets stuck waiting for the input timer to stop. The `while` loop in this place never exits, despite being entirely correct usage-wise: https://github.com/flipperdevices/flipperzero-firmware/blob/027ea9ea36da137144548295c016d99255af53c3/applications/services/input/input.c#L115-L117
* Raising the priority of times via a `furi_timer_set_thread_priority(FuriTimerThreadPriorityElevated);` call mitigates this issue.
* Yielding from the app's tick callback **also** mitigates the issue.
* Replacing the View Dispatcher tick callback with a dedicated timer **also** mitigates the issue, unless the timer interval is set too low. In my tests, setting it to 1 tick resulted in an identical deadlock issue.

Based on the following, we have a reasonable suspicion that the issue relates to the thread starvation. The timer thread runs at a priority 2, unless boosted by the API call, which has been proven to help. The main thread that executes the View Dispatcher's tick callback, as well as the GUI thread, both run at a priority 16. Therefore, it's quite possible that those threads never give the timer thread a chance to execute under those broken conditions, preventing the stop request from being acknowledged and thus leading to a deadlock.

## What makes sense?

Those observations support the theory of this issue being caused by improper thread priorities:
1. Increasing the timer priority "helps".
[...]

PR #3381, opened 2024-01-18
furi/core/timer: resolve timer handle use-after-free post deletion

kbembedded:
# What's new

While helping investigate #3380, I stumbled upon invalid behavior in `furi_timer_delete()` where it continues to use a timer handle to detect if the timer has been deleted. This loop checking if the timer was running was introduced in #1716.

This behavior is fine for static timers (declared with `xTimerCreateStatic()` and using non-dynamically allocated memory), but is dangerous in situations where the timer is dynamically allocated. The `xTimerDelete()` call is responsible for freeing the timer handle. Due to this, there is always potential for `free()ed` memory to be reallocated and initialized while `furi_timer_free()` is spinning and waiting for the timer to stop which could lead to a crash or deadlock.

This PR addresses that by first safely removing the FURI callback handle from the TimerID of the timer. At this point, if the timer fires, `TimerCallback()` in FURI timers would be called, but with a `NULL`  TimerID nothing further would happen.

Now, the allocated callback memory can be freed without issue. And finally the call to `xTimerDelete()` will submit the timer for deletion in such a way that FURI no longer needs to wait for it to be removed.

EDIT: While investigating other parts of FreeRTOS, I noted that `vPortFree()` does actually zero out the memory region, and can confirm that the `furi_timer_free()` loop pattern currently in use absolutely accesses `free()`ed and zero'ed memory and the fix from #1716 only works coincidentally:

```
Breakpoint 1, furi_timer_free (instance=instance@entry=0x2000b878) at furi/core/timer.c:62
62	    furi_assert(!furi_kernel_is_irq_or_masked());
(gdb) print *(Timer_t*)0x2000b878
$5 = {
  pcTimerName = 0x0,
  xTimerListItem = {
    xItemValue = 925375,
    pxNext = 0x2000aa94,
    pxPrevious = 0x20001458 <xActiveTimerList1+8>,
    pvOwner = 0x2000b878,
    pxContainer = 0x20001450 <xActiveTimerList1>
  },
  xTimerPeriodInTicks = 250,
  pvTimerID = 0x2000ae19,
  pxCallbackFunction = 0x801435b <TimerCallback>,
  uxTimerNumber = 0,
  ucStatus = 5 '\005'
}
(gdb) c
Breakpoint 3, furi_timer_is_running (instance=0x2000b878) at furi/core/timer.c:132
132	    furi_assert(!furi_kernel_is_irq_or_masked());
(gdb) print *(Timer_t*)0x2000b878
$7 = {
  pcTimerName = 0x0,
  xTimerListItem = {
    xItemValue = 925375,
    pxNext = 0x2000aa94,
    pxPrevious = 0x20001458 <xActiveTimerList1+8>,
    pvOwner = 0x2000b878,
    pxContainer = 0x20001450 <xActiveTimerList1>
  },
  xTimerPeriodInTicks = 250,
  pvTimerID = 0x2000ae19,
  pxCallbackFunction = 0x801435b <TimerCallback>,
  uxTimerNumber = 0,
  ucStatus = 5 '\005'
}
(gdb) c
Continuing.

Breakpoint 3, furi_timer_is_running (instance=0x2000b878) at furi/core/timer.c:132
132	    furi_assert(!furi_kernel_is_irq_or_masked());
(gdb) print *(Timer_t*)0x2000b878
$8 = {
  pcTimerName = 0x0,
  xTimerListItem = {
    xItemValue = 0,
    pxNext = 0x0,
    pxPrevious = 0x0,
    pvOwner = 0x0,
    pxContainer = 0x0
  },
  xTimerPeriodInTicks = 0,
  pvTimerID = 0x0,
  pxCallbackFunction = 0x0,
  uxTimerNumber = 0,
  ucStatus = 0 '\000'
}
```

The `furi_timer_is_running()` loop accesses an invalid and `free()`ed pointer, and manages to work because `ucStatus` of `0` lines up with the timer not running. As noted above, if this range is re-allocated while `furi_timer_free()` is looping, this pointer could potentially show the timer still running.

# Verification 

Unfortunately, I don't have any application code where this actually causes an issue. This is not an issue I experienced in any real scenario. I only noticed it while doing a deep read of the FURI timer code.

EDIT: As the original test in #1716 seemed to be the snake game, I've run and quit the game dozens of times with the change active and it behaves as expected.

# Checklist (For Reviewer)

### DIFF
--- a/furi/core/common_defines.h
+++ b/furi/core/common_defines.h
@@ -21,6 +21,10 @@ extern "C" {
 #define FURI_PACKED __attribute__((packed))
 #endif
 
+#ifndef FURI_ALWAYS_STATIC_INLINE
+#define FURI_ALWAYS_STATIC_INLINE __attribute__((always_inline)) static inline
+#endif
+
 #ifndef FURI_IS_IRQ_MASKED
 #define FURI_IS_IRQ_MASKED() (__get_PRIMASK() != 0U)
 #endif
--- a/targets/f7/furi_hal/furi_hal_gpio.c
+++ b/targets/f7/furi_hal/furi_hal_gpio.c
@@ -129,11 +129,9 @@ void furi_hal_gpio_init_ex(
         LL_GPIO_SetPinMode(gpio->port, gpio->pin, LL_GPIO_MODE_INPUT);
         LL_SYSCFG_SetEXTISource(sys_exti_port, sys_exti_line);
         if(mode == GpioModeInterruptRise || mode == GpioModeInterruptRiseFall) {
-            LL_EXTI_EnableIT_0_31(exti_line);
             LL_EXTI_EnableRisingTrig_0_31(exti_line);
         }
         if(mode == GpioModeInterruptFall || mode == GpioModeInterruptRiseFall) {
-            LL_EXTI_EnableIT_0_31(exti_line);
             LL_EXTI_EnableFallingTrig_0_31(exti_line);
         }
         if(mode == GpioModeEventRise || mode == GpioModeEventRiseFall) {
@@ -149,6 +147,7 @@ void furi_hal_gpio_init_ex(
         if(LL_SYSCFG_GetEXTISource(sys_exti_line) == sys_exti_port &&
            LL_EXTI_IsEnabledIT_0_31(exti_line)) {
             LL_EXTI_DisableIT_0_31(exti_line);
+            LL_EXTI_ClearFlag_0_31(exti_line);
             LL_EXTI_DisableRisingTrig_0_31(exti_line);
             LL_EXTI_DisableFallingTrig_0_31(exti_line);
         }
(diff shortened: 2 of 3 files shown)


furi/core/timer.c
```diff
@@ -67,17 +67,29 @@ void furi_timer_free(FuriTimer* instance) {
...
 
-    furi_check(xTimerDelete(hTimer, portMAX_DELAY) == pdPASS);
-
-    while(furi_timer_is_running(instance)) furi_delay_tick(2);
-
+    /* If callback memory was dynamically allocated, there is potential
+     * for a race condition between issuing xTimerDelete, the timer actually
+     * being freed, and freeing the allocated TimerID that could lead to a
+     * use-after-free situation. Therefore, we first need to set the TimerID
+     * to NULL; this is wrapped in a critical secion in vTimerSetTimerID, and
```
CookiePLMonster:
```suggestion
     * to NULL; this is wrapped in a critical section in vTimerSetTimerID, and
```

furi/core/timer.c
```diff
@@ -67,17 +67,29 @@ void furi_timer_free(FuriTimer* instance) {
...
+     */
     if((uint32_t)callb & 1U) {
+        /* Set TimerID to NULL to not have a use-after-free */
+        vTimerSetTimerID(hTimer, NULL);
+
         /* Callback memory was allocated from dynamic pool, clear flag */
         callb = (TimerCallback_t*)((uint32_t)callb & ~1U);
 
         /* Return allocated memory to dynamic pool */
         free(callb);
```
skotopes: vTimerSetTimerID doesn't guarantee that timer thread is not running this callback right now. If it is then there will be condition race where you freed callb memory before it was processed.

skotopes: So you still have to ensure that timer stopped before removing callback

skotopes (changes requested):

furi/core/timer.c
kbembedded: That is a great point. Looks like the correct way to handle teardown in that case be to stop the timer, wait for it to be no longer active, free the callback memory, and then issue a delete request.

I'll get that updated, do some testing, and push a change to the PR in the next day or so.

PR #3384, opened 2024-01-18
NFC: Display unread Mifare Classic bytes as question marks

TollyH:
# What's new

- Before, the Mifare Classic data view displayed bytes that hadn't been read from the card as `00`, now they display as `??`

# Verification 

- Open Mifare Classic files in the NFC app that are missing different parts, such as a key, a single block, or an entire sector
- Verify that the bytes are correctly displayed as `??` and line up with the known bytes

# Checklist (For Reviewer)

### DIFF
--- a/applications/main/nfc/helpers/protocol_support/mf_classic/mf_classic_render.c
+++ b/applications/main/nfc/helpers/protocol_support/mf_classic/mf_classic_render.c
@@ -18,59 +18,48 @@ void nfc_render_mf_classic_info(
     furi_string_cat_printf(str, "\nSectors Read: %u/%u", sectors_read, sectors_total);
 }
 
+static void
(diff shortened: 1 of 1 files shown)


gornekich (changes requested):
Hello @TollyH ! Thanks for PR!

I like the idea of showing '??' instead of 00, when we didn't read data. I can see small issues in implementation. I think there is a better way to render unknown data.

Please consider my patch attached. You can apply it and check. If everything is fine, please commit this patch, update PR and we will merge it.

To apply patch run `git apply <path_to_patch>`
[mf_classic_render.txt](https://github.com/flipperdevices/flipperzero-firmware/files/14027530/mf_classic_render.txt)

TollyH:
Thanks @gornekich, I've gone ahead and applied that patch and all seems to working fine

gornekich (approved):
Thanks!

hedger (approved):

PR #3394, opened 2024-01-24
FBT/uFBT: Enable C++20/GNU23 in VSCode IntelliSense

CookiePLMonster:
# What's new

This PR corrects two VSCode configuration issues:
1. `ms-vscode.cpptools` can't be a configuration provider to itself, so it has been removed. See also: https://github.com/microsoft/vscode-cpptools/issues/11889#issuecomment-1911159040
2. #3451 upgraded the codebase to C++20/GNU2x, but the VSCode templates remained at C++17/GNU17. This has been corrected. GNU2x is a deprecated alias for GNU23, but it's not supported yet by GCC in use, yet the VSCode extension requires it. Those should be aliases anyway so despite different naming, there should be no behaviour disparity.

**Original PR below:**

~~According to Microsoft, `"compilerPath"` overrides `compile_commands.json`, so just set it to be empty. Do note this is different to not setting `"compilerPath"` at all, as this will make VSCode fall back to the platform's default compiler, while an empty path will specifically tell it not to try to do anything.~~

~~For a detailed discussion on the VSCode extension, see this GitHub issue, especially the last few posts:
https://github.com/microsoft/vscode-cpptools/issues/11889~~

~~Notably:~~
> The current behavior of the compilerPath overriding the compiler in compile_commands.json **is expected and not a defect**.

# Verification 

~~1. Use `./fbt vscode_dist` or `ufbt vscode_dist`.
3. Observe that VSCode's IntelliSense now uses system defines specific to Cortex M4:
image~~

# Checklist (For Reviewer)

### DIFF
--- a/.vscode/example/cpptools/c_cpp_properties.json
+++ b/.vscode/example/cpptools/c_cpp_properties.json
@@ -5,7 +5,6 @@
             "compilerPath": "${workspaceFolder}/toolchain/x86_64-windows/bin/arm-none-eabi-gcc.exe",
             "intelliSenseMode": "gcc-arm",
             "compileCommands": "${workspaceFolder}/build/latest/compile_commands.json",
-            "configurationProvider": "ms-vscode.cpptools",
             "cStandard": "gnu17",
             "cppStandard": "c++17"
         },
@@ -14,7 +13,6 @@
             "compilerPath": "${workspaceFolder}/toolchain/x86_64-linux/bin/arm-none-eabi-gcc",
             "intelliSenseMode": "gcc-arm",
             "compileCommands": "${workspaceFolder}/build/latest/compile_commands.json",
-            "configurationProvider": "ms-vscode.cpptools",
             "cStandard": "gnu17",
             "cppStandard": "c++17"
         },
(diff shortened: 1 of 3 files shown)


CookiePLMonster:
This has been identified as a bug in the vscode extension, and will be fixed in an upcoming Insiders release 1.19.3: https://github.com/microsoft/vscode-cpptools/issues/11889#issuecomment-1911209593

I'll keep this PR as a draft until this is fixed, especially since it's also been identified that one of the properties in `c_cpp_properties.json` is used incorrectly: https://github.com/microsoft/vscode-cpptools/issues/11889#issuecomment-1911159040
> Also, I see this invalid entry present: `"configurationProvider": "ms-vscode.cpptools"` Which you may want to remove also. The C/C++ Extension is not a provider of custom configurations to itself. This field would be for custom configuraiton providers such as CMake Tools, or Makefile Tools.

It's a tiny nitpick, but might as well fix it since I paid attention to this file anyway. That said, the current change to `"compilerPath"` will not be necessary anymore once the extension is fixed.

hedger:
Thank you for digging into that. Looking forward to fixes from MS.

CookiePLMonster:
> Thank you for digging into that. Looking forward to fixes from MS.

Looks like Microsoft fixed the issue in a pre-release 1.19.3:
image

Therefore, `compilerPath` can now stay as-is. Do you wish me to edit this PR to edit the `"configurationProvider"` line out as per the guidance from the issue I linked in the cpptools repo, or not bother and just close it as invalid?

Hopefully it's not long before these changes hit the stable extension, as they have not updated the stable branch since November.

CookiePLMonster:
Hey @hedger, I went ahead and applied the `configurationProvider` fix, as well as a fix for the C++/C standard level to repurpose this PR into something more useful than a stylistic one-liner.

hedger (approved):

PR #3272, opened 2023-12-06
ble: profile rework

hedger:
# What's new

**General BLE**:
- ble: reworked profile subsystem
- now applications can define custom BLE profiles
- profiles define BLE advertising parameters and own service set

**BLE internals**:
- combined SHCI & HCI event processing threads
- reworked event dispatching to services to support new way of defining profiles and overcome fixed wpan handler count limit
- battery service (&others): now supports multiple instances
- bt service: now only init BLE pairing dialog when needed to save memory
- moved BLE HID profile from furi_hal to lib/ble_profile as an example (custom profiles can also be defined in app's private libs)

**Extra beacon**:
- added API for controlling additional, non-connectable BLE advertisements (BLE beacon)
- that beacon works in background, independently from main BLE subsystem
- extra beacon has status indication on desktop
- added example app for configuring and testing extra beacon feature

**Misc**:
- Moved example_custom_font app from debug to examples
- Improved "ps" CLI command output: added thread priorities
- Added FURI_PACKED macro for packed structures
- hid_app: cleaner code split for USB/BLE transports
- fbt: better handling of malformed target.json files

# Verification 

- test that update from older versions and downgrade do not break BLE pairing
- run BLE remote app, check that old pairings work & app itself works
- run BLE stress test from companion apps
- run BLE Beacon app in various combinations with BLE on/off, RPC active/not
- check out cool "ps"
- idk

# Checklist (For Reviewer)

- [x] PR has description of feature/bug or link to Confluence/Jira task
- [x] Description contains actions to verify feature/bugfix
- [x] I've built this code, uploaded it to the device and verified feature/bugfix

PR #3408, opened 2024-01-30
Bluetooth: Add OpenHaystack implementation

thejudge156:
# What's new

- Modify gap.c to add OpenHaystack-related defines, such as the key path and length
- Add the max BLE packet length to bt.h (Do I change that location and/or name?)
- Add an advertisement data template for OpenHaystack
- Add a bool for when the keyfile is detected, and therefor OpenHaystack gets enabled.
- Change the mac address to the first 6 key bytes, with a bitmask over the first. (Required by AirTags)
- Add a flag for disabling the OpenHaystack implementation for a profile, OpenHaystack for the HID profile for example is disabled.

# Verification

- A lot of people can't use OpenHaystack because of the expensive hardware surrounding it, but using [this project](https://github.com/dchristl/macless-haystack), you just need Docker and Python
- Clone the repo, run generate-keys.py from the repo with Python
- ./generate-keys.py -p haystack
- Find the out directory in the repo, copy the keyfile and the json file to a safe location
- Copy the haystack_keyfile to ext root
- Reboot the Flipper Zero, enable Bluetooth
- Follow the instructions on the repo as detailed [here](https://github.com/dchristl/macless-haystack?tab=readme-ov-file#set-up-endpointfetch-location-server)
- Import the JSON file into [here](https://dchristl.github.io/macless-haystack/)
- If reports do not appear after a refresh, see the last note below

# Checklist (For Reviewer)

# Note
Current issue that I could use help finding a way around; You can't connect to the mobile app because the service UUID and the broadcasting name are not set while Haystack is on, and the mac address is not the same to stock. You *can* connect to the Flipper if you reenable service UUID in gap.c, connect to the mobile app, then disable services again. It will even reconnect even with the service disabled.

One more thing, until [this issue](https://github.com/biemster/FindMy/issues/40) gets resolved, iOS 17 devices will not report the "AirTag" frequently. It will see it, and give you the "someone is tracking you" notification, but I guess the Apple Servers don't like it? Either way, without a iOS <= 16 phone around, this is practically useless. Mostly posting this now so when the issue *is* fixed, all issues with this PR will be resolved and can be merged.

### DIFF
--- a/applications/services/bt/bt_service/bt.h
+++ b/applications/services/bt/bt_service/bt.h
@@ -8,6 +8,7 @@ extern "C" {
 #endif
 
 #define RECORD_BT "bt"
+#define BT_LE_MAX_LENGTH 31
 
 typedef struct Bt Bt;
 
--- a/targets/f7/ble_glue/gap.c
+++ b/targets/f7/ble_glue/gap.c
@@ -5,7 +5,8 @@
 
 #include <furi_hal.h>
 #include <furi.h>
-#include <bt/bt_service/bt_i.h>
+#include <bt/bt_service/bt.h>
+#include <storage/storage.h>
 
 #define TAG "BtGap"
 
--- a/targets/f7/ble_glue/gap.c
+++ b/targets/f7/ble_glue/gap.c
@@ -5,11 +5,15 @@
 
 #include <furi_hal.h>
 #include <furi.h>
+#include <bt/bt_service/bt_i.h>
 
 #define TAG "BtGap"
 
 #define FAST_ADV_TIMEOUT 30000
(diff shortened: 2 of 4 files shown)


skotopes:
See https://github.com/flipperdevices/flipperzero-firmware/pull/3272

thejudge156:
Would you prefer I add this functionality into that system when that PR gets merged?

skotopes:
@thejudge156 yes

thejudge156:
Closing until [3272](https://github.com/flipperdevices/flipperzero-firmware/pull/3272) gets merged

skywinder:
@thejudge156  #3272  is merged  alredy!🙌

PR #3410, opened 2024-01-31
lfrfid/em4100: added support for different bit rates

Mrkvak:
# What's new
The current em4100 code supports only 64 clock cycles per bit.

This PR adds support for different bit rates (32 and 16), instead of default 64 clock cycles per bit.
According to EM4100 datasheet available at https://docs.rs-online.com/6ee7/0900766b8161beae.pdf, used combinations are:
 * A6: Manchester 64cl/bit
 * A5: Manchester 32cl/bit
 * B6: Bi-phase 64cl/bit
 * B5: Bi-phase 32cl/bit
 * C4: PSK, 16cl/bit

This change should make the Manchester 32clock/bit variant and possibly also 16clock/bit PSK variant.
The code is not very nice, but I'm unfamiliar with the codebase to think of something better right now. Suggestions are welcome!

# Verification 
* Get EM4100 card with 32 clock cycles / bit (or program T5577 in that way)
* Reading it should fail before applying this PR and succeed after applying this PR.

# Checklist (For Reviewer)


xMasterX:
Hi!

I've tested with two different T5577 tags and pm3

64 clock - works fine
32 clock - works fine
16 clock - flipper can't read them at all but can emulate properly, and write on T5577 (no success screen cuz no read but the tag is written successfuly and can be read with proxmark)
Manual read in ASK or PSK modes doesnt help, it can't see RF/16 tag at all

skotopes (dismissed):

skotopes (approved):

PR #3431, opened 2024-02-07
furi/core/timer: resolve timer handle use-after-free post deletion (was #3381)

kbembedded:
_Note: I managed to completely break PR #3381. I believe I understand what happened and have adjusted my workflow to prevent it from reoccurring. The sole commit here addresses the changes that were requested. Below is the original description, with edits as needed, in order to not need to reference the original PR. My apologies for the additional noise._

# What's new

While helping investigate #3380, I stumbled upon invalid behavior in `furi_timer_delete()` where it continues to use a timer handle to detect if the timer has been deleted. This loop checking if the timer was running was introduced in #1716.

This behavior is fine for static timers (declared with `xTimerCreateStatic()` and using non-dynamically allocated memory), but is dangerous in situations where the timer is dynamically allocated. The `xTimerDelete()` call is responsible for freeing the timer handle. Due to this, there is always potential for `free()ed` memory to be reallocated and initialized while `furi_timer_free()` is spinning and waiting for the timer to stop which could lead to a crash or deadlock.

This PR addresses that by first stopping the timer if dynamically allocated memory was used, waiting for confirmation of the timer being stopped, freeing the memory, and then deleting the timer without any additional waiting. If a timer is ever instantiated without the dynamically allocated memory used for callback, the timer is simply just deleted.

The `vPortFree()` call does actually zero out the memory region, and can confirm that the `furi_timer_free()` loop pattern currently in use absolutely accesses `free()`ed and zero'ed memory and the fix from #1716 only works coincidentally:

```
Breakpoint 1, furi_timer_free (instance=instance@entry=0x2000b878) at furi/core/timer.c:62
62	    furi_assert(!furi_kernel_is_irq_or_masked());
(gdb) print *(Timer_t*)0x2000b878
$5 = {
  pcTimerName = 0x0,
  xTimerListItem = {
    xItemValue = 925375,
    pxNext = 0x2000aa94,
    pxPrevious = 0x20001458 <xActiveTimerList1+8>,
    pvOwner = 0x2000b878,
    pxContainer = 0x20001450 <xActiveTimerList1>
  },
  xTimerPeriodInTicks = 250,
  pvTimerID = 0x2000ae19,
  pxCallbackFunction = 0x801435b <TimerCallback>,
  uxTimerNumber = 0,
  ucStatus = 5 '\005'
}
(gdb) c
Breakpoint 3, furi_timer_is_running (instance=0x2000b878) at furi/core/timer.c:132
132	    furi_assert(!furi_kernel_is_irq_or_masked());
(gdb) print *(Timer_t*)0x2000b878
$7 = {
  pcTimerName = 0x0,
  xTimerListItem = {
    xItemValue = 925375,
    pxNext = 0x2000aa94,
    pxPrevious = 0x20001458 <xActiveTimerList1+8>,
    pvOwner = 0x2000b878,
    pxContainer = 0x20001450 <xActiveTimerList1>
  },
  xTimerPeriodInTicks = 250,
  pvTimerID = 0x2000ae19,
  pxCallbackFunction = 0x801435b <TimerCallback>,
  uxTimerNumber = 0,
  ucStatus = 5 '\005'
}
(gdb) c
Continuing.

Breakpoint 3, furi_timer_is_running (instance=0x2000b878) at furi/core/timer.c:132
132	    furi_assert(!furi_kernel_is_irq_or_masked());
(gdb) print *(Timer_t*)0x2000b878
$8 = {
  pcTimerName = 0x0,
  xTimerListItem = {
    xItemValue = 0,
    pxNext = 0x0,
    pxPrevious = 0x0,
    pvOwner = 0x0,
    pxContainer = 0x0
  },
  xTimerPeriodInTicks = 0,
  pvTimerID = 0x0,
  pxCallbackFunction = 0x0,
  uxTimerNumber = 0,
  ucStatus = 0 '\000'
}
```

The `furi_timer_is_running()` loop accesses an invalid and `free()`ed pointer, and manages to work because `ucStatus` of `0` lines up with the timer not running. As noted above, if this range is re-allocated while `furi_timer_free()` is looping, this pointer could potentially show the timer still running.

# Verification 

The original test in #1716 was an issue with the snake game upon exit. In verifying this fix I've confirmed that repeated start/exit of the snake game correctly tears down the timer. There are no leaks of heap memory during that cycling. And walking through code it will correctly ensure the timer is fully stopped and has no chance of being fired off while memory is being freed and the timer is being deleted.

# Checklist (For Reviewer)

### DIFF
--- a/furi/core/timer.c
+++ b/furi/core/timer.c
@@ -67,17 +67,22 @@ void furi_timer_free(FuriTimer* instance) {
 
     callb = (TimerCallback_t*)pvTimerGetTimerID(hTimer);
 
-    furi_check(xTimerDelete(hTimer, portMAX_DELAY) == pdPASS);
-
-    while(furi_timer_is_running(instance)) furi_delay_tick(2);
-
     if((uint32_t)callb & 1U) {
+        /* If callback memory was allocated, it is only safe to free it with
+         * the timer inactive. Send a stop command and wait for the timer to
+         * be in an inactive state.
+         */
+        furi_check(xTimerStop(hTimer, portMAX_DELAY) == pdPASS);
+        while(furi_timer_is_running(instance)) furi_delay_tick(2);
+
         /* Callback memory was allocated from dynamic pool, clear flag */
         callb = (TimerCallback_t*)((uint32_t)callb & ~1U);
 
         /* Return allocated memory to dynamic pool */
         free(callb);
     }
+
+    furi_check(xTimerDelete(hTimer, portMAX_DELAY) == pdPASS);
 }
 
 FuriStatus furi_timer_start(FuriTimer* instance, uint32_t ticks) {


skotopes:
I'll merge this one, but it's not good: xTimerIsTimerActive doesn't know if there are other command in the control queue, which leads to some uncertainty. 

Good solution is to throw away FreeRTOS timer subsystem and write one with predictable control flow.

skotopes (approved):

CookiePLMonster:
> I'll merge this one, but it's not good: xTimerIsTimerActive doesn't know if there are other command in the control queue, which leads to some uncertainty.

Are you thinking of sequences like `stop -> start -> stop -> free` where `free` would consider the timer permanently stopped after the first `stop` command gets processed, before it later attempts to re-start? That sounds tricky to account for, especially since RTOS docs specifically say they make the queue opaque:

> [Software timer](https://www.freertos.org/RTOS-software-timer.html) functionality is provided by a timer service/daemon task. Many of the public FreeRTOS timer API functions send commands to the timer service task through a queue called the timer command queue. **The timer command queue is private to the RTOS kernel itself and is not directly accessible to application code.**

kbembedded:
> I'll merge this one, but it's not good: xTimerIsTimerActive doesn't know if there are other command in the control queue, which leads to some uncertainty.

> Good solution is to throw away FreeRTOS timer subsystem and write one with predictable control flow.

If there are other `furi_timer_...` calls made after `furi_timer_free` is called on a timer instance, that would certainly be a use bug.

FURI timer API isn't really set up to handle enforcing that at the moment (e.g. a struct member meant to invalidate the timer for any other `furi_timer_...` calls if `furi_timer_free` was called on that timer); but if you are ripping out the existing timers at some point adding that enforcement may not be necessary.

CookiePLMonster:
> If there are other `furi_timer_...` calls made after `furi_timer_free` is called on a timer instance, that would certainly be a use bug.

If I understand the issue at hand correctly, it's "easier" to hit. Imagine a sequence like

```cpp
furi_timer_stop(timer);
furi_timer_start(timer, /* whatever */);
furi_timer_stop(timer);

furi_timer_free(timer);
```

If my assumptions are correct, FURI could assume that the timer has been permanently stopped after the first scheduled stop command was processed, and be unaware that there are also start and a second stop commands right after. However, considering that not even RTOS has a way to combat this despite them **explicitly** stating the timer needs to wait to be stopped before being deleted, maybe this is an edge case so minimal not even RTOS is worried?

kbembedded:
> > If there are other `furi_timer_...` calls made after `furi_timer_free` is called on a timer instance, that would certainly be a use bug.
> 
> If I understand the issue at hand correctly, it's "easier" to hit. Imagine a sequence like
> 
> ```c++
> furi_timer_stop(timer);
> furi_timer_start(timer, /* whatever */);
> furi_timer_stop(timer);
> 
> furi_timer_free(timer);
> ```

The timer queue is processed sequentially. So if you stop, start, stop, start, ..., free, then the free will always be processed last (which will now stop then delete the timer handle). If there is an application using a timer handle in such a way that it would attempt to do anything on the timer after free is called, then that is the application's fault.

Unless I'm misunderstanding what use case you are talking about?

> If my assumptions are correct, FURI could assume that the timer has been permanently stopped after the first scheduled stop command was processed, and be unaware that there are also start and a second stop commands right after. However, considering that not even RTOS has a way to combat this despite them **explicitly** stating the timer needs to wait to be stopped before being deleted, maybe this is an edge case so minimal not even RTOS is worried?

A timer does not need to be stopped before being deleted. That requirement is a result of how FURI is handling the callback processing. The timer queue handling for most commands will first remove the timer from an active list (most commands will be starting, restarting, stopping, etc.) before performing the action. In the case of STOP, the timer is removed from the active list and the handle is marked as inactive. In the case of DELETE, the timer is removed from the active list and then the handle is freed (which zero's the memory, which led to why #1716 worked).

Do you have a link to where they state a timer needs to be stopped first? I don't see anything readily in the documentation and that is not a quirk I remember in my uses of FreeRTOS.

CookiePLMonster:
> Do you have a link to where they state a timer needs to be stopped first? I don't see anything readily in the documentation and that is not a quirk I remember in my uses of FreeRTOS.

To be precise, this was about the memory allocated for the timer - aka the main issue you've solved here.

PR #3435, opened 2024-02-09
Compile-time APP_DATA_PATH() and APP_ASSETS_PATH()

WillyJL:
# What's new

- `APP_DATA_PATH()` and `APP_ASSETS_PATH()` use `appid` from compile time instead of runtime `thread_id`
- Fixes incorrect paths when these app storage macros are used in timer / service context (eg. timer/draw callback)
- Less runtime overhead due to less string replacing to generate correct paths
- Code not compiled as FAP will fail since FAP_APPID is not defined
- No user intervention required, this is drop in fix, only recompile is needed
- To function, adds C define `FAP_APPID`, could be useful to app makers aswell

# Verification 

- Open app that uses `APP_DATA_PATH()` or `APP_ASSETS_PATH()` in draw/timer callback
- Path is still correct `/ext/apps_.../appid`, instead of ending as `/ext/apps_.../system` or `/ext/apps_.../gui`

# Checklist (For Reviewer)


WillyJL:
Just occurred to me that this would need an equivalent in ufbt, I don't use that myself so I forgot... I can make a PR there as well if we go through with this PR

hedger:
Thanks for the PR.
Overall, it's well-implemented. However, it has a certain side-effect. 

Currently, if a .fap file is renamed or copied on the device, it gets a separate app data folder, according to its new file name. That way, you can create multiple isolated environments for a single app. However, there are no apps that would make use of that at the moment - at least to my knowledge. With current implementation from this PR, app is confined to its appid. Surely, that approach has its own benefits and enables a predictable file location.

This PR also introduces a system-wide define with "system" name at a fairly low level - maybe its scope could also be reduced to app loader level. I'll look into that and see whether that can be adjusted.

WillyJL:
Agreed on the define in `furi.h`, the naming was rushed just to get something working lol. As you say yeah maybe Loader or Storage services could handle this. Before making this PR I had originally fully removed `furi_thread_get_appid()` since it wasn't really used elsewhere, so that's part of why I put it in `furi.h`, but then chose against removing it so here we are.

As for renaming `.fap` files, you make a very good point... at the very least then the behavior should be consistent for `APP_ASSETS_PATH()`. Currently, it extracts to folder based on fap filename, and then the app looks based on compiled appid... easiest solution I see would be making the assets extract to folder based on fap header appid, so after rename it is same folder path. The other option would be finding a way for loader to tell the app what it's appid is based on fap filename, but then we are back to the issue of threads, and it becomes overcomplicated i think.

WillyJL:
> easiest solution I see would be making the assets extract to folder based on fap header appid, so after rename it is same folder path

i was looking around to implement this but realized that fap header has app name but not appid... not sure what the best idea would be then...

skotopes:
@Willy-JL Some proxy object that is initialized at the app start and got it name and then all your internal references goes to it. That is not something that should be on firmware side, but something that lives in app bss

WillyJL:
so then APP_ASSETS_PATH() and APP_DATA_PATH() will remain half broken? and apps will need to make their own system? 

or do you mean an alternative way of doing it that is not drop-in, but replaces the existing macros?

WillyJL:
ah i think i get what you mean, i was unfamiliar with the concept of .bss... so then, at app launch, loader would initialize this global static variable that is located in bss with the app's id, and the app code references to it?

sounds like then we could provide helper macros that construct a path using this variable, but still not handled by firmware. like some sort of printf in app code that uses this appid variable, but inside a macro so app developers dontn eed to worry about the variable itself?

would mean this is not strictly drop-in, but close enough, same idea but using a printf or similar under the hood...

skotopes:
yep

WillyJL:
im new to those concepts but ill try to figure something out, if you have some examples i could look at that would be greatly appreciated!

also i drafted for now, but ill perhaps close this and open a new pr when i have something working

WillyJL:
@skotopes i was looking into how to implement this with bss, and had a counter argument.

i feel like the behavior of renaming a fap file changing storage location as @hedger described is quite weird. yes it has a *very* niche use case, but i think it instead leads to a lot more confusion for people messing around with faps installed manually, than the small advantages it provides.

instead, i was considering that maybe a better solution would be sticking with what i have in this PR thus far (save for moving where the define is made), so paths are defined at compile time (and backwards compatible with existing code), and instead add a new elf section (or attribute in fap header) for the appid, so that assets extractor doesnt need to rely on fap name and can instead use the same appid that was used at compile time to hardcode the paths.

this would fix the disjunction between runtime and compile time paths, as well as as avoid unnecessarily complex logic to store the appid in bss and construct paths on app side at runtime with printf. as well as be backwards compatible, requiring one simple recompile.

CookiePLMonster:
FWIW from the user's point of view this PR is worth it even for this

> Fixes incorrect paths when these app storage macros are used in timer / service context (eg. timer/draw callback)

point alone.

skotopes:
Put implementation aside for a moment.

Do you guys feel like you want to have ability to have multiple app versions with isolated spaces or not?

WillyJL:
While I see the value in that option in some niche use cases, to the average user it can be more of a pain when the app doesn't work as intended when they renamed a file. Say the download a fap, and they download it again on a newer version, but they didn't clean the downloads folder. Now it's `appid (1).fap`. Suddenly, the app lost all its configuration. The user doesn't even have the slightest indication why, since the app itself works, just the configuration is lost.

also, there's the point that compile time macros are easier to use. No need for furi strings or printf to insert appid at runtime, easier to use, and not a breaking change compared to current api.

Personally, I'm 100% for the option of compile time, appid is hardcoded in fap, not chosen by filename.

WillyJL:
Also, I recently discovered there is compile time macro "FAP_VERSION". It only makes logical sense to offer "FAP_APPID". I wasn't aware at the time of making this PR, hence the "system" and "FURI_APPID". "FAP_APPID" would work great and be super intuitive, and "APP_DATA_PATH()" can rely on FAP_APPID then?

skotopes:
Ok, we'll discuss it with @hedger and @DrZlo13 today and then decide

applications/system/js_app/js_app.c
```diff
@@ -98,7 +98,9 @@ static void js_app_free(JsApp* app) {
 int32_t js_app(void* arg) {
     JsApp* app = js_app_alloc();
 
+#define FAP_APPID "js_app"
     FuriString* script_path = furi_string_alloc_set(APP_ASSETS_PATH());
+#undef FAP_APPID
```
WillyJL: using same system as FAP_VERSION means FAP_APPID is only added to apps configured as external, from what i understand. in momentum, we have js_app as external, which would fix this. otherwise, this workaround seems to be fine

applications/services/storage/storage.h
```diff
@@ -15,12 +15,16 @@ extern "C" {
...
 #define STORAGE_ANY_PATH_PREFIX "/any"
-#define STORAGE_APP_DATA_PATH_PREFIX "/data"
-#define STORAGE_APP_ASSETS_PATH_PREFIX "/assets"
 
 #define INT_PATH(path) STORAGE_INT_PATH_PREFIX "/" path
 #define EXT_PATH(path) STORAGE_EXT_PATH_PREFIX "/" path
 #define ANY_PATH(path) STORAGE_ANY_PATH_PREFIX "/" path
+
+#define STORAGE_APPS_DATA_STEM EXT_PATH("apps_data")
+#define STORAGE_APPS_ASSETS_STEM EXT_PATH("apps_assets")
```
WillyJL: moved from storage_i.h "APPS_DATA_PATH" and "APPS_ASSETS_PATH" to here to avoid duplicated definition with "apps_data" explicitly in string. not sure if its fine to be exposed to public api, if it gives wrong idea

WillyJL:
i feel like this is satisfactory and achieves what we discussed. however, there are 2 parts i am not sure what to do about:
- unit tests `test_storage_data_path_apps` and `test_storage_data_path` are not relevant anymore, since its not based on thread appid anymore. rather, only thing i can think to test is that accessing `/ext/apps_data/any_appid` will create this folder correctly.
- since apps use hardcoded assets paths now, they need to be extracted to hardcoded path too, not based on filename. fap header contains app name, not appid. first solution i can think of is adding `appid` to fap header, and bumping api version (but also check header version in case of same api version).

i will implement those solutions in the meantime waiting for comments

WillyJL:
tested with ufbt and apps that use both assets and data paths, seems to be working well to me :D

applications/system/js_app/js_app.c
WillyJL: Actually now that I think about it, this doesn't make much sense... js app doesn't come with assets, and plugins are in app data path, this line should use `EXT_PATH("apps/Scripts")`

hedger (changes requested):
unit tests are failing.
`./fbt flash_usb_full FIRMWARE_APP_SET=unit_tests`
in cli: 
`unit_tests` / `unit_tests test_storage`
```
test_storage_data_path failed:
	applications/debug/unit_tests/tests/storage/storage_test.c:468: storage_dir_open(file, "/data")
test_storage_data_path_apps()
test_storage_data_path_apps failed:
	applications/debug/unit_tests/tests/storage/storage_test.c:443: 1 expected but was 0
```

WillyJL:
Yep, as I mentioned above those tests aren't really relevant anymore. Should I just remove them?

hedger:
Yes - if they are obsolete.

WillyJL:
@hedger done )

hedger (approved):

applications/services/storage/storage_processing.c
```diff
@@ -520,45 +520,26 @@ static FS_Error storage_process_sd_status(Storage* app) {
...
         // Create app data folder if not exists
-        if(create_folders &&
-           storage_process_common_stat(app, apps_data_path_with_appsid, NULL) != FSE_OK) {
-            furi_string_set(apps_data_path_with_appsid, APPS_DATA_PATH);
-            storage_process_common_mkdir(app, apps_data_path_with_appsid);
-            furi_string_cat(apps_data_path_with_appsid, "/");
-            furi_string_cat(apps_data_path_with_appsid, furi_thread_get_appid(thread_id));
-            storage_process_common_mkdir(app, apps_data_path_with_appsid);
+        if(create_folders && storage_process_common_stat(app, apps_data_appid, NULL) != FSE_OK) {
+            FuriString* apps_data = furi_string_alloc_set(STORAGE_APPS_DATA_STEM);
```
hedger, line 529: we're checking existence of one path, but creating a different one

hedger (changes requested):
in certain cases, when uploading files to literal paths in /ext/apps_data/..., alias processing fails. See unit tests log

applications/services/storage/storage_processing.c
WillyJL: it first makes /ext/apps_data here, then 2 lines below it makes /ext/apps_data/appid. it is just ensuring that apps_data exists, previous implementation did this too

WillyJL:
@hedger unit tests are failing in CI setup step, not in unit test itself it seems. when uploading files for apps_data/nfc, complains that nfc folder already exists, which makes sense because now all literal accesses to /ext/apps_data/anything will create such folder. i guess, update CI to accept "already exist" as success, instead of fail?

skotopes:
@hedger @Willy-JL @DrZlo13 PR with int storage drop was merged, so we ready to merge this one. Please finalize and let me know when we can merge it

hedger (dismissed):

applications/services/storage/storage_processing.c
```diff
@@ -513,46 +513,26 @@ static FS_Error storage_process_sd_status(Storage* app) {
...
 
         // Create app data folder if not exists
-        if(create_folders &&
-           storage_process_common_stat(app, apps_data_path_with_appsid, NULL) != FSE_OK) {
-            furi_string_set(apps_data_path_with_appsid, APPS_DATA_PATH);
-            storage_process_common_mkdir(app, apps_data_path_with_appsid);
-            furi_string_cat(apps_data_path_with_appsid, "/");
-            furi_string_cat(apps_data_path_with_appsid, furi_thread_get_appid(thread_id));
-            storage_process_common_mkdir(app, apps_data_path_with_appsid);
+        if(create_folders && storage_process_common_stat(app, apps_data_appid, NULL) != FSE_OK) {
```
hedger, line 528: This code incorrectly handles paths like `"/ext/apps_data/nfc"`
See https://github.com/flipperdevices/flipperzero-firmware/actions/runs/10283770489/job/28458336904#step:7:89

@Willy-JL @DrZlo13

WillyJL: seems like its still the error i mentioned above, setting up the CI unit tests, not running them. code that causes the issue is not in flipper code, but python helper is failing when creating a folder that already exists. i had tried in 887cd4b94b84e61587cbfb0ccab9968807930927 to make it ignore an "already exists" error but seems like this fix had some syntax issue.

i will fix this, but maybe could consider making it so checking for dir exist in "/ext/apps_data/*" will say that dir already exists?

hedger (dismissed):

applications/services/storage/storage_processing.c
hedger: I think maintaining compatibility and existing behaviour from API is quite important - looking forward to hearing @DrZlo13's opinion on these changes.

skotopes:
image

Need some more work, @hedger will provide details

WillyJL:
> Need some more work, @hedger will provide details

looks like he is writing a novel with all the details :P

hedger:
I think the key points from our past discussion are:
- avoiding side-effects from API calls, especially those which are getter-like (~ creating intermediate paths when doing path existence checks);
- maintaining existing behaviour for sequences of storage API calls;
- redesigning the `storage_process_alias()` function so it does not lead to similar cases.

The issue was exposed by unit tests runner script, when modifications were needed to maintain previous behaviour.

WillyJL:
@skotopes @hedger i have an idea that might make this respect existing behavior while fixing the main issue this pr wanted to address (wrong appid depending on thread and fap filename).

current implementation has `APP_DATA_PATH("file.txt")` evaluate to `/data/file.txt`

my original proposal was to change it to evaluate to `/ext/apps_data/appid/file.txt` at compile time, but brings issues when its not an app invoking that path but instead for example a script copying files

how about merging the ideas? at compile time `APP_DATA_PATH("file.txt")` will evaluate to `/data:appid/file.txt` (and `/assets:appid/file.txt` for assets).
- can keep using the `FAP_APPID` define to make this work
- still need to include appid in .fap_meta header section, so assets are extracted to correct place
- will mean there is still overhead in storage service for translating `/data:` and `/assets:` paths, but not a huge deal
- old apps using `/data/` and `/assets/` will error due to different api version (will bump it) and also fap header anyway, no need to handle cases for `/data/` paths without `:appid`, can just return an error if for whatever reason storage service receives such a path
- the folder is automatically created only when using `/data:appid/`, not `/ext/apps_data/appid/`, which is same behavior as now

PR #3456, opened 2024-02-16
Parser for Santiago, Chile BIP transit card

rbasoalto:
# What's new

- Parser for Bip Card (Santiago, Chile public transit)

# Verification 

- Try it with any of [these saved cards](https://gist.github.com/rbasoalto/398decb957efe8d93a825e034ab25fcf) (old dumps of some cards that I got permission to share, converted to Flipper NFC file format)

# Checklist (For Reviewer)


applications/main/nfc/plugins/supported_cards/bip.c
```diff
@@ -0,0 +1,372 @@
...
+            BipTransaction* charge = &bip_data.charges[i];
+            parse_bip_timestamp(&data->block[SECTOR_BLOCK_OFFSET(11, i)], &charge->timestamp);
+            parse_uint16_txn_amount(
+                &data->block[SECTOR_BLOCK_OFFSET(11, i)].data[10], &charge->amount);
+        }
+
+        parsed = true;
+    } while(false);
+
+    if(parsed) {
```
gornekich: Could you please move all logic below to the `do {} while(false)` structure above? I see that if we have parsed == true here, there is no chance that plugin wouldn't parse anything.

You can just move all code below to `do{} while(false)` structure above, leave `parse = true` at the end and return parsed value: `return parsed`

applications/main/nfc/plugins/supported_cards/bip.c
```diff
@@ -0,0 +1,372 @@
...
+            if(key != bip_keys_a[i]) {
+                parsed = false;
+                break;
+            }
+            key = bit_lib_bytes_to_num_be(sec_tr->key_b.data, 6);
+            if(key != bip_keys_b[i]) {
+                parsed = false;
+                break;
+            }
+        }
```
gornekich: Is it really necessary to verify all the keys? The execution of this code is quite long and it leads to a little "freeze" before we can see parsed data.

gornekich: I would just check the key that we use in `verify()` function, if it's OK for you

gornekich (changes requested):
Hello @rbasoalto , thanks a lot for PR!

I tested and everything works fine. Please make some changes and we will merge your PR.

applications/main/nfc/plugins/supported_cards/bip.c
rbasoalto: I suppose if we already read the card, then the other keys should match. The only unlikely edge case I can think of is loading a stored card where the first sector keys' match but not the others, though that seems unlikely enough.

applications/main/nfc/plugins/supported_cards/bip.c
rbasoalto: Done! Would you like keeping it as it is now, where we first unpack everything into the `bip_data` struct and only print at the end? Or would you rather print as we unpack each bit of data?

gornekich: It's fine how it's done now. Thanks

gornekich (approved):
Thank you!

PR #3457, opened 2024-02-17
NFC: Fix greedy verify function of washcity plugin

### LINKED ISSUE #3422: NFC: regression in reading MFClassic since 0.98.0-rc
### Describe the bug.

Today, on `0.98.2` I re-read a MFClassic 1K Tag and to my surprise it only found 14/32 keys. However it used to be 30/32 !
I also noticed the screen is not showing the progress bar where it tries all keys of user-dict and system-dict. It only shows "Dont move" for a little longer than usual but way shorter than trying all my ~3800 keys (user+sys combined)

Comparing the new dump with my old, shows that it doesnt even have KeyA for alot sectors, but the key is default key `FF FF FF FF FF FF`
Furthermore, one section, which I have the sniffed key in user-dict, is also missing.

Now I went back Firmware versions and re-tried:
- 0.98.2  bad
- 0.98.1-rc bad
- 0.98.0-rc bad
- 0.97.1 good, gets all 30/32 keys again

~I wonder if it could be related to https://github.com/flipperdevices/flipperzero-firmware/pull/3356 which got introduced in 0.98.0-rc~

philicious:
washcity plugin verify func only checks authing a single block with a single key. and that key being a very commonly used one. thus potentially breaking reading of alot MFC cards by dicts

fixes #3422 

# What's new

- use two blocks and keys for verification
- avoid using a commonly used key at all

# Verification 

- read a MFC with 0xA0A1A2A3A4A5 A-key for sector0 thats not a WashCity card: should now use user/system dict
- WashCity cards should still work and use the plugin

# Checklist (For Reviewer)

### DIFF
--- a/scripts/toolchain/fbtenv.sh
+++ b/scripts/toolchain/fbtenv.sh
@@ -203,7 +203,7 @@ fbtenv_show_unpack_percentage()
 fbtenv_unpack_toolchain()
 {
     echo "Unpacking toolchain to '$FBT_TOOLCHAIN_PATH/toolchain':";
-    rm $FBT_TOOLCHAIN_PATH/toolchain/current || true;
+    rm "$FBT_TOOLCHAIN_PATH/toolchain/current" || true;
     tar -xvf "$FBT_TOOLCHAIN_PATH/toolchain/$TOOLCHAIN_TAR" -C "$FBT_TOOLCHAIN_PATH/toolchain" 2>&1 | fbtenv_show_unpack_percentage;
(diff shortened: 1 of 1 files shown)


philicious:
@gornekich if you could have the bot create a FW based on this branch, I could verify. and also WashCity-card author

I can cleanup codes/commits if I know its safe to re-use instances of `MfClassicKey` `MfClassicAuthContext` `MfClassicError` for secondary reads with different key

PR #3464, opened 2024-02-19
Moscow social card parser

assasinfil:
# What's new

- Parser based on Troyka layouts with additional data from soc card.

# Verification 

- Try to read Moscow social card and validate result with "Транспортные карты Москвы"

# Checklist (For Reviewer)


applications/main/nfc/plugins/supported_cards/social_moscow.c
```diff
@@ -0,0 +1,254 @@
...
+    {.a = 0xa229e68ad9e5, .b = 0x49c2b5296ef4}, //40
+};
+
+static bool social_moscow_get_card_config(SocialMoscowCardConfig* config, MfClassicType type) {
+    bool success = true;
+    if(type == MfClassicType1k) {
+        config->data_sector = 15;
+        config->keys = social_moscow_1k_keys;
+    } else if(type == MfClassicType4k) {
+        config->data_sector = 15;
```
philicious, line 91: if I see it correctly, this will use key `0xa0a1a2a3a4a5`, also for verification. 
which then introduces the same false-positive problem that I just fixed in https://github.com/flipperdevices/flipperzero-firmware/pull/3467 for the wash-city parser:
the 0xa0.. key is very commonly used and if any plugins verify func uses only that key, chances are high for false-positive , essentially breaking reading of alot cards

assasinfil: I chose block 15 for verification as it contains the basic data of the social card, I agree that both keys A and B need to be checked at once.

philicious: yes. I would suggest using two non-common keys for verification if its the right card. doesnt have to be the actual data-section, just unique enough to identify the proper card.

(if possible, add sanity-checks in the parse function = make sure it returns false if the parsed data doesnt make sense)

great work besides that remark 👍 :)

assasinfil:
@gornekich I hope that I fixed the problem with catching other cards and the parser is now working. Please check that the parser does not interfere with others.

Astrrra (approved):

skotopes (approved):

PR #3467, opened 2024-02-19
NFC: Fix washcity plugin verify function being to greedy

### LINKED ISSUE #3422: NFC: regression in reading MFClassic since 0.98.0-rc

philicious:
washcity plugin verify func only checks authing a single block with a single key. and that key being a very commonly used one. thus potentially breaking reading of alot MFC cards by dicts

**I have built this with `./fbt && ./fbt flash_usb_full` and verified it working**

fixes #3422 
obsoletes #3457 with proper code

# What's new

- use two blocks and keys for verification
- avoid using a commonly used key at all

# Verification 

- read a MFC with 0xA0A1A2A3A4A5 A-key for sector0 thats not a WashCity card: should now use user/system dict ✅

# Checklist (For Reviewer)

### DIFF
--- a/applications/main/nfc/plugins/supported_cards/washcity.c
+++ b/applications/main/nfc/plugins/supported_cards/washcity.c
@@ -57,37 +57,17 @@ static bool washcity_verify(Nfc* nfc) {
     bool verified = false;
 
     do {
-        MfClassicAuthContext auth_context;
+        const uint8_t verify_sector_number = 1;
+        const uint8_t verify_block_number =
+            mf_classic_get_first_block_num_of_sector(verify_sector_number);
+        FURI_LOG_D(TAG, "Verifying sector %u", verify_sector_number);
+
         MfClassicKey key = {0};
-        uint8_t verify_sector_number = 0;
-        uint8_t verify_block_number = 0;
-        MfClassicError error = MfClassicErrorNone;
-
-        // verify first key thats not a commonly used key
-        verify_sector_number = 1;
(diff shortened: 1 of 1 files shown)


gornekich:
@yaba could you confirm that it's ok to verify washcity cards by trying to auth to 1 sector?

gornekich (approved):
@philicious thanks for PR.

I changed verify function to verify just 1st sector. I think it's enough

yaba:
Apologies for the delay, yes, I should be looking for both A/B keys, instead of just one.

skotopes (approved):

PR #3495, opened 2024-03-04
Show the wrong PIN Attempt count on the login screen

FireFly7386:
# What's new

- It now shows how often you entered the wrong PIN when the flipper is locked

# Verification 

- Just enter the wrong PIN and the counter should increase
Verification_Show_Wrong_Pin_Attempts

# Checklist (For Reviewer)


applications/services/desktop/scenes/desktop_scene_pin_input.c
```diff
@@ -127,12 +128,12 @@ bool desktop_scene_pin_input_on_event(void* context, SceneManagerEvent event) {
         case DesktopPinInputEventResetWrongPinLabel:
             desktop_scene_locked_light_red(false);
             desktop_view_pin_input_set_label_primary(desktop->pin_input_view, 0, 0, NULL);
-            FuriString* enter_pin_string = furi_string_alloc();
-            furi_string_set_str(enter_pin_string, "Enter PIN (Wrong Attempts: ");
+            state->enter_pin_string = furi_string_alloc();
```
CookiePLMonster: Remove this line, `on_enter` has already allocated the string. You only need to modify it, not allocate again.

applications/services/desktop/scenes/desktop_scene_pin_input.c
```diff
@@ -98,7 +110,7 @@ bool desktop_scene_pin_input_on_event(void* context, SceneManagerEvent event) {
     Desktop* desktop = (Desktop*)context;
     bool consumed = false;
     uint32_t pin_timeout = 0;
-
+    DesktopScenePinInputState* state = malloc(sizeof(DesktopScenePinInputState));
```
rdefeo: You shouldn't need to reallocate a new state here. Instead, retrieve the state that was originally created in the on_enter method, by using scene_manager_get_scene_state function. This will return a ptr to the originally allocated struct, and thus you won't need to reallocate the enter_pin_string below.

applications/services/desktop/scenes/desktop_scene_pin_input.c
```diff
@@ -121,8 +133,19 @@ bool desktop_scene_pin_input_on_event(void* context, SceneManagerEvent event) {
         case DesktopPinInputEventResetWrongPinLabel:
             desktop_scene_locked_light_red(false);
             desktop_view_pin_input_set_label_primary(desktop->pin_input_view, 0, 0, NULL);
-            desktop_view_pin_input_set_label_secondary(
-                desktop->pin_input_view, 44, 25, "Enter PIN:");
+            uint32_t attempts = furi_hal_rtc_get_pin_fails();
+            state->enter_pin_string = furi_string_alloc();
```
rdefeo: See the above comment - this allocation is not needed.

applications/services/desktop/scenes/desktop_scene_pin_input.c
```diff
@@ -121,8 +133,19 @@ bool desktop_scene_pin_input_on_event(void* context, SceneManagerEvent event) {
...
+            furi_string_set_str(state->enter_pin_string, "Enter PIN (Fails: ");
```
rdefeo: This can be cleaned up a bit by doing something like:

```
furi_string_set_str(state->enter_pin_string, "Enter PIN (Fails: %lu):");
furi_string_printf(state->enter_pin_string, attempts);
```

And it makes the full string easier to read!

applications/services/desktop/scenes/desktop_scene_pin_input.c
```diff
@@ -121,8 +133,19 @@ bool desktop_scene_pin_input_on_event(void* context, SceneManagerEvent event) {
...
+            furi_string_cat_printf(state->enter_pin_string, "%lu", attempts);
+            furi_string_cat_str(state->enter_pin_string, "):");
+            const char* enter_pin_cstr = furi_string_get_cstr(state->enter_pin_string);
```
rdefeo: You only use the var `enter_pin_cstr` in the if block, so it should be declared within there.

FireFly7386: Thanks for the infos! Will probably look at it tommorow :)

skotopes:
@FireFly7386 you will need review from our UI?UX team first. 

@Karator FYI

Karator:
Thanks for the interesting idea!
It would be better if the text of the attempts were displayed centered at the bottom of the screen, like this:
wrong_pin

FireFly7386:
Isn't there the OK button or do you mean to show the attempts if it is currently hidden?

applications/services/desktop/scenes/desktop_scene_pin_input.c
FireFly7386: thanks for the helps, did that :)

applications/services/desktop/scenes/desktop_scene_pin_input.c
FireFly7386: couldn´t just leave it away, did the things below, but thanks for pointing out

FireFly7386:
The Text is now down here:
Image1

And if you input an arrow it disappers so it doesn´t overlap with any other texts:
Image2

It only updates the text on the DesktopPinInputEventResetWrongPinLabel event at the moment, should i make it update at the DesktopPinInputEventUnlockFailed event or are you fine with that?

applications/services/desktop/scenes/desktop_scene_pin_input.c
```diff
@@ -123,6 +132,14 @@ bool desktop_scene_pin_input_on_event(void* context, SceneManagerEvent event) {
             desktop_view_pin_input_set_label_primary(desktop->pin_input_view, 0, 0, NULL);
             desktop_view_pin_input_set_label_secondary(
                 desktop->pin_input_view, 44, 25, "Enter PIN:");
+            uint32_t attempts = furi_hal_rtc_get_pin_fails();
+            furi_string_set_str(state->enter_pin_string, "");
```
skotopes: furi_string_reset

applications/services/desktop/scenes/desktop_scene_pin_input.c
```diff
@@ -88,6 +89,13 @@ void desktop_scene_pin_input_on_enter(void* context) {
     desktop_view_pin_input_hide_pin(desktop->pin_input_view, true);
     desktop_view_pin_input_set_label_button(desktop->pin_input_view, "OK");
     desktop_view_pin_input_set_label_secondary(desktop->pin_input_view, 44, 25, "Enter PIN:");
+    uint32_t attempts = furi_hal_rtc_get_pin_fails();
+    state->enter_pin_string = furi_string_alloc();
+    furi_string_cat_printf(state->enter_pin_string, "Wrong Attempts: %lu", attempts);
+    if(attempts > 0) {
+        const char* enter_pin_cstr = furi_string_get_cstr(state->enter_pin_string);
+        desktop_view_pin_input_set_label_tertiary(desktop->pin_input_view, 30, 60, enter_pin_cstr);
```
skotopes: I'd align text to center and then use center middle of the screen for anchoring

skotopes (changes requested):

skotopes:
I'd say that this thing behaves quite strange. @Karator you'd want to see it.

FireFly7386:
Are you talking about that even tough attempts = 0 it sometimes executes the if block anyways? Currently trying to figure out why.

applications/services/desktop/scenes/desktop_scene_pin_input.c
FireFly7386: are you fine with those changes?

FireFly7386:
did that fix the wierd behavior you were talking about?

skotopes (approved):

PR #3507, opened 2024-03-11
Api Symbols: replace asserts with checks

skotopes:
# What's new

- Api Symbols: replace asserts with checks
- Unavoidable voids in functions with no arguments
- Update PVS config and make it happy
- Various fixes: incorrect args, incorrect function signatures and etc...

# Verification 

- Full set of tests

# Checklist (For Reviewer)


CookiePLMonster:
Fun sidenote - while this might technically seem like these changes add overhead to the FW code, `furi_check` is a noreturn function, and therefore the optimizer now "gains" extra info about the variables being validated. This opens up additional optimization opportunities in the rest of the function body, thus potentially generating more optimal code overall.

For example:
```cpp
void some_func(int x)
{
   furi_check(x > 0);

   // The optimizer now knows that x is positive, and removes this entire code path
   if (x == -1) {
    // Do something
  }
}
```

With GCC's `__builtin_assume`, asserts could theoretically do the same, just without a runtime check in the release builds.

skotopes:
@CookiePLMonster it is going to cost us 12kb anyway

CookiePLMonster:
@skotopes Sure, size wise it'll be a loss (because you can't not pay the cost of having the check body in there), but with more info for the optimizer the function bodies can _technically_ be more optimal. If the failure case in `furi_check` is compiled as an unlikely case (as in, the branch is **taken** for the check failure, and it's not taken for the normal case) then the runtime cost of that check can be lower than the potential gains from extra optimizations.

This might only be an issue if you are starved for code size, and I don't know how much more budget you have on that front.

hedger (dismissed):

hedger (approved):

PR #3517, opened 2024-03-14
refactor(scripts): use match-case, add minor cosmetic touches

seppzer0:
# What's new

- Minor refactoring of existing Python scripts, primarily in the scope of using `match-case` instead of `if-elif` structures
 
NB! : minimum Python version would have to be bumped to 3.10 .

# Verification 

- Nothing is changed functionally; the modified code should behave the same when launching it.

# Checklist (For Reviewer)


skotopes:
@seppzer0 to be honest this PR type is always very confusing to me: it kinda got some improvements in it, but not on the level expected. Like, why specifically those files were updated and not the other? Is it really looks better now? Is it going to improve code maintenance in a long term? Looks like answer for all those questions is "not really".

Sorry, but this PR is going to be rejected.

seppzer0:
> @seppzer0 to be honest this PR type is always very confusing to me: it kinda got some improvements in it, but not on the level expected. Like, why specifically those files were updated and not the other? Is it really looks better now? Is it going to improve code maintenance in a long term? Looks like answer for all those questions is "not really".
> 
> Sorry, but this PR is going to be rejected.

@skotopes ,

> Like, why specifically those files were updated and not the other?

Well, because there weren't any other `if-elif` constructions anywhere else in Python code that could be replaced with `match-case`. If there were, please kindly let me know.

> Is it really looks better now?

It is always subjective. To me, it's newer and cleaner, so it does. However small the change is.

> Is it going to improve code maintenance in a long term?

I'm not sure what to answer to this. Wouldn't say it will, but also wouldn't say otherwise.

> Sorry, but this PR is going to be rejected.

It's ok. PRs are PRs to *propose* improvements to maintainers, not enforce them.

PR #3522, opened 2024-03-18
Added the js command

Milk-Cool:
# What's new

- Added the js command, which allows the user to run JS scripts from the Flipper CLI.

# Verification 

- Flash the firmware onto your flipper (obviously)
- Try running `js /ext/apps/Scripts/delay.js` in the CLI (`./fbt cli`, `ufbt cli` or https://lab.flipper.net/)

# Checklist (For Reviewer)


Milk-Cool:
just found out that importing modules does not work for some reason. do i have to manually load the modules in some way?

it's just that in `js_app.c` i don't see any code that would do that. i thought these were loaded automatically.

Milk-Cool:
also, maybe it would be better if we load the `js_app` app with the script path as the argument? we wouldn't get any output in the console but other than that everything would be much simpler and cleaner

Milk-Cool:
> just found out that importing modules does not work for some reason.

image
wait, nevermind. it doesn't work either way. is loading modules broken on the `dev` branch?

also should i open an issue on this?

Milk-Cool:
ok i didn't update the apps on the flipper, i had to use `./fbt flash_usb_full` instead of `./fbt flash_usb`

importing modules still does not work with the `js` command though...

Milk-Cool:
> also, maybe it would be better if we load the js_app app with the script path as the argument? 

Nevermind... it does not automatically exit when the script is done

nminaylov:
Hi, the proper method here is to implement CLI handler inside js_app.c. You need a startup hook that registers CLI command handler (you can see "ibutton_start" as an example)

Milk-Cool:
hello, it seems like i won't have enough time to do that before i go to school, and i don't really have much motivation too. i will be able to add the `js` command in mid-june, but for now... do whatever you want with this pr

nminaylov:
Ok, I can do it myself later

nminaylov:
Moved to #3539

PR #3550, opened 2024-03-29
Create Dockerfile

ferryman0608:
# What's new

**Type of Change:** Feature
**Explanation:** In an effort to support the project's reach and user experience, I've developed a Dockerfile with the aim of streamlining the setup process and ensuring consistent environments across different platforms. This initiative is part of a larger project I'm working on, which leverages large language models to automatically generate Dockerfiles tailored for specific projects.

# Verification 

**Usage:**
Place the Dockerfile in the root directory of the project;
Run the following command in the root directory of the project to build the image: docker build -t <image_name> . ;
Run the following command to run the container: docker run -it <image_name> bash;

Attached to this PR is the Dockerfile I've proposed. I’m fully open to discussing any aspects of it and am willing to make necessary adjustments to better align with your project's goals and practices.

Thank you very much for considering this contribution.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,51 @@
+# Use 64-bit Ubuntu as the base image
+FROM ubuntu:20.04
+
+# Set environment variable to make the installation non-interactive
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Set the working directory inside the container
+WORKDIR /flipperzero-firmware
+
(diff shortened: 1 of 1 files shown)


skotopes:
There is no need to use docker for firmware development. We've been designing build system and toolchain in a way that it doesn't require any additional dependencies from user OS. Even more using docker will cripple it quite badly.

PR #3553, opened 2024-03-30
Hide unlock with reader for MFU-C

bettse:
# What's new

- Hide the "unlock with reader" for MFU-C, which use 3des and not a 32bit password (https://en.wikipedia.org/wiki/MIFARE#MIFARE_Ultralight_family)

# Verification 

- Read MFU-C: confirm "Unlock with Reader" is hidden
- Read MFU EV1, NTAG21x, etc: confirm "Unlock with Reader" is shown

# Checklist (For Reviewer)

### DIFF
--- a/applications/main/nfc/scenes/nfc_scene_mf_ultralight_unlock_menu.c
+++ b/applications/main/nfc/scenes/nfc_scene_mf_ultralight_unlock_menu.c
@@ -20,12 +20,17 @@ void nfc_scene_mf_ultralight_unlock_menu_on_enter(void* context) {
     uint32_t state =
         scene_manager_get_scene_state(nfc->scene_manager, NfcSceneMfUltralightUnlockMenu);
     if(nfc_device_get_protocol(nfc->nfc_device) == NfcProtocolMfUltralight) {
-        submenu_add_item(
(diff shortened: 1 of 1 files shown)


skotopes:
@gornekich are you ok with this kinds of type check? As i understand we can make it through `data->type == MfUltralightTypeMfulC`?

bettse:
I did see `MfUltralightTypeMfulC`, `mf_ultralight_get_type_by_version`, and `instance->data->type`, but seems I'd need an `MfUltralightPoller`, and I'm still not very good at the NFC api and opaque types and casting stuff and whatnot, so I wasn't sure how to get from the `nfc_device` I saw to there.

gornekich:
Hello @bettse . The best way to check ultralight type here is to cast nfc_device data to MfUltralight pointer and check it's type. Please look at patch attached.
[check_mfu_type.txt](https://github.com/flipperdevices/flipperzero-firmware/files/14856050/check_mfu_type.txt)

bettse:
Thanks! That works (tested with amiibo and hotel card) and is much more semantic

skotopes (approved):

PR #3568, opened 2024-04-04
IR: Fix crash on duty_cycle=1

WillyJL:
# What's new

- `infrared_signal_is_raw_valid()` says that `duty_cycle > 1` is invalid
- `infrared_worker_set_raw_signal()` checks `duty_cycle < 1`
- therefore, `duty_cycle == 1` fails the check, when really it should be accepted

# Verification 

- send an infrared raw with `duty_cycle: 1`

# Checklist (For Reviewer)

### DIFF
--- a/lib/infrared/worker/infrared_worker.c
+++ b/lib/infrared/worker/infrared_worker.c
@@ -612,7 +612,7 @@ void infrared_worker_set_raw_signal(
     furi_check(timings);
     furi_check(timings_cnt > 0);
     furi_check((frequency <= INFRARED_MAX_FREQUENCY) && (frequency >= INFRARED_MIN_FREQUENCY));
-    furi_check((duty_cycle < 1.0f) && (duty_cycle > 0.0f));
+    furi_check((duty_cycle <= 1.0f) && (duty_cycle > 0.0f));
 
     size_t max_copy_num = COUNT_OF(instance->signal.raw.timings) - 1;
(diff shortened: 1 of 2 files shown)


skotopes:
To be honest `duty_cycle=1` is an invalid value that we never tested. It kinda make no sense. It means that no carrier present. Do you have real life examples for this change?

WillyJL:
I got this report from a friend, they had an IR file with all raws at duty cycle 1. I am not sure if it was captured or generated, but from what they were saying, it seemed to work correctly and do what it was intended to do on older firmware. With furi checks now, it would crash.

skotopes:
> I got this report from a friend, they had an IR file with all raws at duty cycle 1. I am not sure if it was captured or generated, but from what they were saying, it seemed to work correctly and do what it was intended to do on older firmware. With furi checks now, it would crash.

That was clearly generated file. What happens with duty cycle 1.0: there will be no carrier frequency. We really'd like to know what system actually uses that kind of signal.

WillyJL:
He told me it was set to 1 on purpose and used to test accurate IR measuring devices

skotopes (approved):

skotopes:
@Willy-JL this will work, but not because it was designed to do so.

I highly don't recommend to use it for modulation debugging. Also after some experiments I found that some timings are slightly off.

I'll merge this PR, but more research required.

PR #3571, opened 2024-04-04
Status output !TX/RX on the GDO2 CC1101 pin

quen0n:
# What's new

The essence is to display the transmission/reception status in the form of a logical level on the GDO2 pin of the CC1101 chip. This allows external devices to understand the current operating status of the radio chip
At rest, the GDO2 pin is in the floating-state. In receive/transmit mode, GDO2 switches to push-pull mode. When receiving, a logical 1 appears, when transmitting, a logical 0 appears. When the chip goes into rest mode, the GDO2 pin also returns to the floating-state
This can be used to control external signal amplifiers or to indicate the current state of the radio chip

Demo: https://disk.yandex.ru/d/kEsjUYyuDQM_gA

# Verification 

1) Go to SubGHz -> Read. А high logic level of 3.3 V (or other chip supply voltage) should appear at the GDO2 output
2) Just go to the SubGHz menu. The voltage on the GDO2 pin becomes unstable and can take on different values
3) Go to SubGHz -> Saved and send any saved signal. A logical zero (or 0 volt) will appear at the GDO2 output

# Checklist (For Reviewer)


Skorpionm (approved):

Skorpionm (dismissed):

skotopes:
Why low when transmitting? Can you please give some details on what kind of circuit expected to be controlled by gd2?

quen0n:
> Why low when transmitting? Can you please give some details on what kind of circuit expected to be controlled by gd2?

I used this pin to make a LED indication of signal transmission and reception. If you connect LEDs to pins GDO2 and GDO0, the LEDs will flash in time with the transmission or reception of a signal
A low level on GDO2 is needed so that by connecting the cathode of the LED to GDO2, and the anode to GDO0, the LED blinks in time with the pulses on GDO0. If you swap the logical levels, it turns out that the LED will always be on when there is no transmission, and go out in time with the pulses on GDO0

image

quen0n:
And to control, for example, a board on the e07-433m20s module, it makes no difference what the logical level will be when transmitting or receiving at the GDO2 output. It is enough to connect GDO2 to RXEN, and invert GDO2 to TXEN

skotopes:
To be honest we thought that you are planning to add external amplifier or companion chip(CC1190).
Also what about compatibility with other ext subghz boards?

quen0n:
> To be honest we thought that you are planning to add external amplifier or companion chip(CC1190). Also what about compatibility with other ext subghz boards?

The GDO2 pin is not used on any external radio modules that I know of. Therefore, this function will not harm anyone, but will open up new opportunities

skotopes:
> > Why low when transmitting? Can you please give some details on what kind of circuit expected to be controlled by gd2?
> 
> I used this pin to make a LED indication of signal transmission and reception. If you connect LEDs to pins GDO2 and GDO0, the LEDs will flash in time with the transmission or reception of a signal A low level on GDO2 is needed so that by connecting the cathode of the LED to GDO2, and the anode to GDO0, the LED blinks in time with the pulses on GDO0. If you swap the logical levels, it turns out that the LED will always be on when there is no transmission, and go out in time with the pulses on GDO0
> 
> image

I'd say that this is very specialized solution that you've made for your board which is not ideal for other uses cases.

Here is why:
- GD2 low on TX is not suites to control ext PA. You don't want you PA active in idle.
- GD0 line was not designed to drive LEDs(50R on flipper side + extra ext load going to affect signal fronts). Most likely it will work without any issues, but I'd do extra tests to ensure signal integrity(we have resolution of 1us for timings, +/- 0.1us per each front should be ok).

quen0n:
Everyone is free to use this signal as they wish

> GD2 low on TX is not suites to control ext PA.

There is already a module e07-433m20s with a receive and transmit amplifier. It has direct gain control inputs TXEN and RXEN. If you control the amplifier using one pin, then in any case you need an inverter - either for TXEN or for RXEN. Therefore, it is not so important what the logical level is for TX or RX.
I'm not sure that anyone in the community will use a separate transmission amplifier, for which an inverter would be superfluous.
In addition, reception amplifiers also require a high signal level to allow operation

>GD0 line was not designed to drive LEDs

You're right, there are some risks. However, the current consumption of LEDs is quite low, and the resistance of 50 ohms is low enough that there will be no problems. I can now take an oscillogram from GDO0 and show the result

quen0n:
I compared the rise time when transmitting on the GDO0 pin with and without an LED load. I didn't notice any difference. I think that there is no effect on the pulse width
Witout LED
image
With led
image

However, this PR is still about the ability to control other devices depending on the state of the transceiver. Or indicating the status of the transceiver.

It seems to me that the current implementation allows this. And how this or that device will be connected is the responsibility of the circuit engineer. LED is just a special case of use

You have a lot of experience, so I will be happy to listen to your arguments for or against

skotopes (approved):

quen0n:
Thank you

PR #3573, opened 2024-04-05
Desktop/Loader: Unload animations before loading FAPs

WillyJL:
# What's new

- Unload animations earlier, before attempting to load FAPs
- Allows for bigger FAP files with bigger animations, without interfering with each other and causing out of memory crashes
- Due to animation unload happening in lifetime of `loader_start()`, instead of app thread state callback, this causes deadlocks (`loader_start()` is happening in desktop event, it needs to load FAP, which needs to unload animation, which happens in desktop event, therefore one desktop event is waiting on a future desktop event, which can't happen)
- To solve these deadlocks, implemented `loader_start_detached_with_gui_error()`. This copies the start arguments to loader event, and returns immediately. Loader thread will process the start detached event and cleanup, therefore it is loader event waiting for desktop event to free animation, so no deadlocks
- This is a bit of a workaround, but i dont see much of a better choice. The only other way would be to remove the semaphore, and move animation unload directly into `desktop_loader_callback()`, which would also be a workaround. I am open to better ideas on how to solve this deadlock problem

# Verification 

- Can load apps fine, big animations don't interfere with big apps
- Opening favorite apps and passport work as intended without regressions
- Holding back button for shutdown does not cause deadlock

# Checklist (For Reviewer)


WillyJL:
marked as draft as usage of favorite apps and passport needs some more testing, havent had time for that.

also in case this loader start detached workaround isn't what we want to go with, im open to ideas.

WillyJL:
just spotted an issue with the latest commit, will fix asap

WillyJL:
tested behavior of favorite apps, passport, shutdown, and changing all of these in desktop settings, seems to be working fine to me

DrZlo13:
When I made #3569, I was thinking about exactly this approach. Good catch.

DrZlo13 (dismissed):
lgtm

skotopes:
2 things that seriously bothering me:
- How many threads are involved in pubsub send
- How many methods that performs similar things are available in the API

I'd do pubsub delivery only from one thread and refactor api

skotopes (approved):

skotopes:
@Willy-JL good job )

PR #3574, opened 2024-04-05
Furi: Add "out of memory" and "malloc(0)" crash messages

WillyJL:
# What's new

- Clear crash messages for "out of memory" and "malloc(0)"
- Allows for easier debugging of what went wrong

# Verification 

- Crash by out of memory or malloc(0), see new messages

# Checklist (For Reviewer)

### DIFF
--- a/furi/core/memmgr_heap.c
+++ b/furi/core/memmgr_heap.c
@@ -486,7 +486,7 @@ void* pvPortMalloc(size_t xWantedSize) {
 
     configASSERT((((size_t)pvReturn) & (size_t)portBYTE_ALIGNMENT_MASK) == 0);
 
-    furi_check(pvReturn);
+    furi_check(pvReturn, xWantedSize ? "out of memory" : "malloc(0)");
     pvReturn = memset(pvReturn, 0, to_wipe);
     return pvReturn;
 }


CookiePLMonster:
`malloc(0)` should be valid AFAIK, so maybe it's better to skip the check (and `memset`) in that case.

WillyJL:
Usually, from my experience, malloc(0) indicates that some unexpected input was given to something that wasn't made to support it, so it would expect to always get back a pointer from malloc and use it anyway. Giving back NULL I imagine would cause follow up issues, which would make it even more difficult to track down... perhaps we could do as I suggested on the discord contributor channel, make a second mallow_notrow() or similar, that doesn't check the return?

Also I see that memmgr is being reworked in a PR recently, so might be best to wait for that

CookiePLMonster:
> make a second mallow_notrow() or similar, that doesn't check the return?

That would be nice - and then `operator new()` could be mapped to `malloc`, as it most likely is already, while `operator new(std::nothrow)` - to `malloc_nothrow`. I'm not sure if this one is defined when exceptions are disabled, though.

skotopes:
@CookiePLMonster embed makes following standards too dangerous. So we intentionally make them much stricter, that allows us to crash early.

skotopes (approved):

PR #3580, opened 2024-04-10
Allow setting view_dispatcher callbacks to NULL again

xMasterX:
# What's new

- Allow setting view_dispatcher callbacks to NULL again
- Checks for callback existence is already present in the code
- Recently added furi_check removes ability to remove callback without freeing whole view_dispatcher, which breaks logic in some Apps

# Verification 

- Check that `view_dispatcher_set_custom_event_callback(view_dispatcher, NULL)` can be called in apps now again
- All working fine

# Checklist (For Reviewer)


skotopes (approved):

skotopes:
@xMasterX I've checked `flipper-xremote` source code and it's kinda puzzling, clearly not how we want people to use GUI.
I'm going to accept this PR, but I still don't see valid use case for changing callbacks on fly.

PR #3358, opened 2024-01-11
[FL-3672] CLI: cat command crash workaround

nminaylov:
# What's new

- CLI: input line length limit
- USB CDC: switch from 2x unidirectional data endpoints to 1x bidirectional

# Verification 

- Check #3206
- Check all USB-related functions

# Checklist (For Reviewer)

- [x] PR has description of feature/bug or link to Confluence/Jira task
- [x] Description contains actions to verify feature/bugfix
- [x] I've built this code, uploaded it to the device and verified feature/bugfix

PR #3622, opened 2024-05-01
USB: Fix file writing hang regression for big files

### LINKED ISSUE #3452: USB: File writing hang regression for big files
### Describe the bug.

Have 10+ user reports, including first hand, of qFlipper consistently hanging during the update process while writing resources.tar (although at random points throughout the write operation). I tested back through the commit history and narrowed it down to #3358. In particular, reverting the USB CDC parameter changes fixes the issues.

### Reproduction

1. Connect to qFlipper
2. Start an update operation (eg. Install from file)
3. At some random point through resources.tar, writes hang
4. qFlipper reports timeout, update fails

### Target

Flipper USB -> qFlipper on all operating systems

### Logs

In Flipper serial debug logs, write operations flow through normally until they suddenly stop, for no apparent reason, and a few seconds later qFlipper reports a timeout error. No other information is present in the logs.

### Anything else?

The PR in question mentions "switch from 2x unidirectional data endpoints to 1x bidirectional".
Seems like this is what is causing issues, although not clear why. Is there any more info on this change?

WillyJL:
# What's new

- Fixes #3452 as it is still an issue
- TLDR:
  - #3358 fixed crashes when using `cat` on flipper serial
  - it also included some unrelated USB CDC parameter changes
  - from what I've tested, the crash for `cat` on serial is fixed even when not including the USB CDC changes
  - this additional unrelated change was only denoted as `USB CDC: switch from 2x unidirectional data endpoints to 1x bidirectional`
  - from what I can see there are no additional details, no bugs this is attempting to fix, and no improvements this brings
  - there are however some very real issues with this tiny USB CDC changes
  - during USB file writing (like qFlipper or Web update) there is a random chance of USB lockup
  - when this happens, Flipper USB will be fully unresponsive until rebooted
  - qFlipper will say timeout after some seconds, Web will just remain stuck forever
  - the bigger the file being written over USB, the higher chance of lockup
  - it is most prominent when installing some big CFW (6MB+ resources.tar), but I have seen multiple users struggling to update OFW due to this bug
  - in other cases I told people to downgrade to older OFW to avoid this bug, and they could not even downgrade due to having newer OFW currently installed with this bug
  - CFWs have reverted this changes in same way as this PR and have had no issues since. going from CFW to new CFW, or from CFW to OFW
  - but since OFW after 0.99 has this bug, if user has OFW installed they can have issues updating to newer OFW or installing CFW

# Verification 

- Install firmware with big resources.tar multiple times and no issues

# Checklist (For Reviewer)

### DIFF
--- a/targets/f7/furi_hal/furi_hal_usb_cdc.c
+++ b/targets/f7/furi_hal/furi_hal_usb_cdc.c
@@ -7,13 +7,13 @@
 #include "usb.h"
 #include "usb_cdc.h"
 
-#define CDC0_RXD_EP 0x02
+#define CDC0_RXD_EP 0x01
 #define CDC0_TXD_EP 0x82
-#define CDC0_NTF_EP 0x81
+#define CDC0_NTF_EP 0x83
 
 #define CDC1_RXD_EP 0x04
-#define CDC1_TXD_EP 0x84
-#define CDC1_NTF_EP 0x83
+#define CDC1_TXD_EP 0x85
+#define CDC1_NTF_EP 0x86
 
 #define CDC_NTF_SZ 0x08
 


skotopes:
https://github.com/flipperdevices/flipperzero-firmware/pull/3358 actually fixes two problems: 
- dual endpoint configuration causes linux to send garbage to the device
- line buffer overflow in cli caused by garbage sent to the device

So what you doing here is breaking first one. I'd rather see proper fix that is finding root cause of single endpoint cdc crash.

WillyJL:
I see, thanks for clearing that up. It's the crucial piece of information that wasn't specified anywhere (atleast as far as I could see).

I'll try to find out more but it is really difficult to debug, there are no weird logs or crashes usually, just stops working. It almost feels like random memory corruption.

skotopes:
Correct fix that properly covers all cases in #3705

PR #3653, opened 2024-05-16
TLSF and the Temple of Corrupted Memory

DrZlo13:
# What's new

- [ Describe changes here ]

# Verification 

# Checklist (For Reviewer)


CookiePLMonster:
Regarding the 8-byte alignment, GCC indeed appears to make that assumption, with or without `-Os`.

This code:
```c
int32_t align_test_app(void* p) {
    UNUSED(p);
    FURI_LOG_I("TEST", "Hello world");
    FURI_LOG_I("TEST", "I'm align_test!");

    void* b = malloc(4);

    uintptr_t valB = (uintptr_t)b;
    return (valB & 7) == 0;
}
```

compiled into:
image

So it's a very good idea to guarantee proper alignment, at least for `malloc` and `realloc`.

DrZlo13:
I disabled all dolphin animations and ran some tests:

# tlsf

>: free
Free heap size: 141416
Total heap size: 185560
Minimum heap size: 137696
Maximum heap block: 140328
Aux pool total free: 2752
Aux pool max free block: 2644

>: free
Free heap size: 141304
Total heap size: 185560
Minimum heap size: 125768
Maximum heap block: 132512
Aux pool total free: 2752
Aux pool max free block: 2644

# heap4

>: free
Free heap size: 146200
Total heap size: 185544
Minimum heap size: 142512
Maximum heap block: 145136
Pool free: 1464
Maximum pool block: 1348

>: free
Free heap size: 146136
Total heap size: 185544
Minimum heap size: 142440
Maximum heap block: 145296
Pool free: 1464
Maximum pool block: 1348


It looks like in the current version tlsf has no meaning except for the correct implementation of realloc.

CookiePLMonster:
Suggestion - if this goes through, consider exposing `tlsf_memalign_offs` as a non-standard `aligned_offset_alloc(size_t alignment, size_t size, size_t offset)` function, much like MSVC does:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-offset-malloc

It may be handy for cases where a nested object or class field requires alignment, but the rest of the structure does not.

PR #3672, opened 2024-05-28
Furi: wrap message queue in container, prepare it for epoll. Accessor: disable expansion service on start.

skotopes:
# What's new

- Furi: wrap message queue in container, prepare it for epoll.
- Accessor: disable expansion service on start.

# Verification 

- Check everything
- And accessor app too

# Checklist (For Reviewer)

### DIFF
--- a/applications/debug/accessor/accessor_app.cpp
+++ b/applications/debug/accessor/accessor_app.cpp
@@ -34,12 +34,16 @@ void AccessorApp::run(void) {
 AccessorApp::AccessorApp()
     : text_store{0} {
     notification = static_cast<NotificationApp*>(furi_record_open(RECORD_NOTIFICATION));
+    expansion = static_cast<Expansion*>(furi_record_open(RECORD_EXPANSION));
     onewire_host = onewire_host_alloc(&gpio_ibutton);
+    expansion_disable(expansion);
     furi_hal_power_enable_otg();
 }
 
 AccessorApp::~AccessorApp() {
     furi_hal_power_disable_otg();
+    expansion_enable(expansion);
+    furi_record_close(RECORD_EXPANSION);
     furi_record_close(RECORD_NOTIFICATION);
     onewire_host_free(onewire_host);
 }
--- a/applications/debug/accessor/accessor_app.h
+++ b/applications/debug/accessor/accessor_app.h
@@ -6,6 +6,7 @@
 #include "helpers/wiegand.h"
 #include <one_wire/one_wire_host.h>
 #include <notification/notification_messages.h>
+#include <expansion/expansion.h>
 
 class AccessorApp {
 public:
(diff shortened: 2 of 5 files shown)


CookiePLMonster:
If the intention behind using a static message queue is memory savings, you could allocate the buffer directly after the queue struct, coalescing those two allocations into one. Since their lifetime is always identical, it might be a good idea to reduce fragmentation and allocation bookkeeping.

GCC extension `uint8_t* buffer[];` as the last structure field makes it easy, as this field doesn't contribute to the size of the structure, but dictates the type of a memory buffer following it and also enforces alignment. This means you can just do a single alloc of size `sizeof(FuriMessageQueue) + buffer_size`.

skotopes:
> If the intention behind using a static message queue is memory savings, you could allocate the buffer directly after the queue struct, coalescing those two allocations into one. Since their lifetime is always identical, it might be a good idea to reduce fragmentation and allocation bookkeeping.
> 
> GCC extension `uint8_t* buffer[];` as the last structure field makes it easy, as this field doesn't contribute to the size of the structure, but dictates the type of a memory buffer following it and also enforces alignment. This means you can just do a single alloc of size `sizeof(FuriMessageQueue) + buffer_size`.

Nope, that was not goal. I need ability to extended message queue with additional fields for epoll implementation. This is a first step for that.
Also multiple smaller allocations are preferred: more chance to fit them into fragmented memory.

CookiePLMonster:
> Also multiple smaller allocations are preferred: more chance to fit them into fragmented memory.

In the current case you're not winning much - the struct allocation is 12 bytes, but 4 bytes are wasted for a pointer to the other allocation. This means that if those two allocations get merged together, the bigger allocation grows by just 8 bytes - if the memory is fragmented enough for this to matter when allocating the queue, the Flipper is likely about to crash on OOM anyway.

skotopes:
> > Also multiple smaller allocations are preferred: more chance to fit them into fragmented memory.
> 
> In the current case you're not winning much - the struct allocation is 12 bytes, but 4 bytes are wasted for a pointer to the other allocation. This means that if those two allocations get merged together, the bigger allocation grows by just 8 bytes - if the memory is fragmented enough for this to matter when allocating the queue, the Flipper is likely about to crash on OOM anyway.

I've pushed couple changes that combines your proposal and gives a place for epoll.

hedger (approved):

gsurkov (approved):

CookiePLMonster:
> > > Also multiple smaller allocations are preferred: more chance to fit them into fragmented memory.
> > 
> > 
> > In the current case you're not winning much - the struct allocation is 12 bytes, but 4 bytes are wasted for a pointer to the other allocation. This means that if those two allocations get merged together, the bigger allocation grows by just 8 bytes - if the memory is fragmented enough for this to matter when allocating the queue, the Flipper is likely about to crash on OOM anyway.
> 
> I've pushed couple changes that combines your proposal and gives a place for epoll.

Looks optimal, great stuff!

PR #3673, opened 2024-05-30
[FL-3810] Felica emulation

RebornedBrain:
# What's new

- Now Flipper can emulate Felica Lite-S cards

# Verification 

- You need two Flippers 
- Read some real Felica card with one then select 'Emulate UID'
- On the second one select 'Read' and apply them to each other
- The second one will read all card data just like with real card.

# Checklist (For Reviewer)


skotopes (dismissed):

lib/nfc/nfc_mock.c
```diff
@@ -50,11 +53,31 @@ typedef struct {
     Iso14443_3aSelResp sel_resp[2];
 } Iso14443_3aColResData;
 
+typedef struct {
+    uint8_t length;
+    uint8_t polling_cmd;
+    uint16_t system_code;
+    uint8_t request_code;
+    uint8_t time_slot;
+} Felica_PollingRequest;
```
gornekich: FelicaPollingRequest

lib/nfc/nfc_mock.c
```diff
@@ -50,11 +53,31 @@ typedef struct {
...
+
+typedef struct {
+    uint8_t code;
+    FelicaIDm idm;
+    FelicaPMm pmm;
+} Felica_SensfResData;
```
gornekich: FelicaSensfResData

lib/nfc/nfc_mock.c
```diff
@@ -50,11 +53,31 @@ typedef struct {
...
+
+typedef struct {
+    uint16_t system_code;
+    Felica_SensfResData sens_res;
+} Felica_PT_Memory;
```
gornekich: FelicaPTMemory

lib/nfc/protocols/felica/felica_listener.c
```diff
@@ -49,6 +57,102 @@ const FelicaData* felica_listener_get_data(const FelicaListener* instance) {
...
+    const FelicaListenerGenericRequest* const generic_request) {
+    FURI_LOG_D(TAG, "Write cmd");
+
+    const FelicaListenerWriteRequest* request = (FelicaListenerWriteRequest*)generic_request;
+    const FelicaListenerWriteBlockData* data_ptr =
+        felica_listener_get_write_request_data_pointer(instance, generic_request);
+
+    FelicaListenerWriteCommandResponse* resp =
+        (FelicaListenerWriteCommandResponse*)malloc(sizeof(FelicaListenerWriteCommandResponse));
+    furi_check(resp);
```
gornekich: no need in furi_check(). Malloc will crash if there is no memory
Also unnecessary type cast

lib/nfc/protocols/felica/felica_listener.c
```diff
@@ -49,6 +57,102 @@ const FelicaData* felica_listener_get_data(const FelicaListener* instance) {
...
+    FelicaListener* instance,
+    const FelicaListenerGenericRequest* generic_request) {
+    const uint8_t cmd_code = generic_request->header.code;
+    switch(cmd_code) {
+    case FELICA_CMD_READ_WITHOUT_ENCRYPTION:
+        return felica_listener_command_handler_read(instance, generic_request);
+    case FELICA_CMD_WRITE_WITHOUT_ENCRYPTION:
+        return felica_listener_command_handler_write(instance, generic_request);
+    default:
+        furi_crash("FeliCa incorrect command");
```
gornekich: May be we should print log with unsupported command here? I think it's not right to crash if we don't support a command yet

lib/nfc/protocols/felica/felica_listener_i.h
```diff
@@ -2,20 +2,178 @@
...
+#define FELICA_LISTENER_READ_BLOCK_COUNT_MIN (1U)
+#define FELICA_LISTENER_WRITE_BLOCK_COUNT_MAX (2U)
+#define FELICA_LISTENER_WRITE_BLOCK_COUNT_MIN (1U)
+
 typedef enum {
     Felica_ListenerStateIdle,
     Felica_ListenerStateActivated,
 } FelicaListenerState;
 
+/// @brief Generic Felica request same for both read and write operations.
```
gornekich: Please, rework comments in doxygen style that we use everywhere

gornekich (dismissed):

lib/nfc/protocols/felica/felica_listener.c
RebornedBrain: Actually there are only 2 possible commands in Felica: Read and Write and no others, therefore I've added crash here in order to escalate situation when wrong command code was provided. In normal conditions such situation is not possible, unless user will put another command code directly (for example using proxmark felica raw command) and in such situation crash is more preferable, because it attracts user's attention.

gornekich: I see. However, if we want to emulate the card, we still shouldn't crash. Even if you send wrong command to original card, it will not brick. I think just adding error log is enough.

skotopes (approved):

PR #3685, opened 2024-06-04
RPC: Add TarExtract command, some small fixes

WillyJL:
# What's new

- RPC storage command to extract tar file to destination (needs flipperdevices/flipperzero-protobuf#50, probably need to update submodule if/when it is merged)
- Fixed extract failing if not extracting at `/ext` or `/int`, this is because first directory entry in tar is `/` and `path_concat()` would result in `/path/to/dest/` with trailing slash. This is a basic fix but might be worth changing the behavior of the path library in such edge cases?
- Added some missing storage state resets, all other functions had them so it seemed like they should be there, revert if i misunderstood :D

# Verification 

- Storage operations, backup and restore, work fine over RPC
- Tar files (like resources.tar, so USTAR format) can be extracted to chosen destination (not sure on easiest way to make one-off RPC calls to test)

# Checklist (For Reviewer)

### DIFF
--- a/applications/services/rpc/rpc_storage.c
+++ b/applications/services/rpc/rpc_storage.c
@@ -220,6 +220,8 @@ static void rpc_system_storage_list_root(const PB_Main* request, void* context) 
     RpcSession* session = rpc_storage->session;
     furi_assert(session);
 
+    rpc_system_storage_reset_state(rpc_storage, session, true);
+
     const char* hard_coded_dirs[] = {"any", "int", "ext"};
 
     PB_Main response = {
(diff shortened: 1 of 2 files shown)


gsurkov:
@Willy-JL 
> not sure on easiest way to make one-off RPC calls to test

We have some python scripts for that purpose which can be found [here](https://github.com/flipperdevices/flipperzero_protobuf_py).

In the meantime, I've added support for `TarExtractRequest` at flipperdevices/flipperzero_protobuf_py#25

applications/services/rpc/rpc_storage.c
```diff
@@ -219,6 +220,8 @@ static void rpc_system_storage_list_root(const PB_Main* request, void* context)
     RpcSession* session = rpc_storage->session;
     furi_assert(session);
 
+    rpc_system_storage_reset_state(rpc_storage, session, true);
```
gsurkov: Not needed as the same call has already been made in `rpc_system_storage_list_process()`

gsurkov (changes requested):

gsurkov (dismissed):

skotopes (approved):

PR #3690, opened 2024-06-06
CCID: Updated types to match Short APDU size

kidbomb:
# What's new

The CCID stack  currently implemented in Flipper operates in short APDU mode, which supports only 256 bytes for the data length. The buffer on furi_hal_usb_ccid.c already supports this size - however the data types on the callback functions are still using uint32_t instead of uint8_t. This PR updates the callbacks so they use uint8_t as the proper type for the data length

# Verification 

 - Start the CCID app and verify that the device was successfully detected as a CCID device with the command:

> opensc-tool --atr

The command above shoud return `3b:00`
- Check that example 1 works with:

> opensc-tool --send-apdu 01:01:00:00

The command above should return

```
Received (SW1=0x90, SW2=0x00)`
```

- Check that example 2 works with:

> opensc-tool --send-apdu 01:02:00:00

The command above should return:

```
Received (SW1=0x90, SW2=0x01):
62 63 bc
```

- Check that example 3 works with:

> opensc-tool --send-apdu 01:03:00:00:02:CA:FE

The command above should return:

```
Received (SW1=0x90, SW2=0x02):
CA FE ..
```

# Checklist (For Reviewer)

### DIFF
--- a/targets/f7/furi_hal/furi_hal_usb_ccid.c
+++ b/targets/f7/furi_hal/furi_hal_usb_ccid.c
@@ -332,7 +332,7 @@ void CALLBACK_CCID_IccPowerOn(
             if(callbacks[CCID_SLOT_INDEX] != NULL) {
                 callbacks[CCID_SLOT_INDEX]->icc_power_on_callback(
                     responseDataBlock->abData,
-                    &responseDataBlock->dwLength,
+                    (uint8_t*)&responseDataBlock->dwLength,
                     cb_ctx[CCID_SLOT_INDEX]);
                 responseDataBlock->bStatus = CCID_COMMANDSTATUS_PROCESSEDWITHOUTERROR |
                                              CCID_ICCSTATUS_PRESENTANDACTIVE;
@@ -364,10 +364,11 @@ void CALLBACK_CCID_XfrBlock(
             if(callbacks[CCID_SLOT_INDEX] != NULL) {
                 callbacks[CCID_SLOT_INDEX]->xfr_datablock_callback(
                     (const uint8_t*)receivedXfrBlock->abData,
-                    receivedXfrBlock->dwLength,
+                    (uint8_t)receivedXfrBlock->dwLength,
                     responseDataBlock->abData,
-                    &responseDataBlock->dwLength,
+                    (uint8_t*)&responseDataBlock->dwLength,
                     cb_ctx[CCID_SLOT_INDEX]);
+
                 responseDataBlock->bStatus = CCID_COMMANDSTATUS_PROCESSEDWITHOUTERROR |
                                              CCID_ICCSTATUS_PRESENTANDACTIVE;
             } else {
--- a/targets/furi_hal_include/furi_hal_usb_ccid.h
+++ b/targets/furi_hal_include/furi_hal_usb_ccid.h
@@ -17,12 +17,12 @@ typedef struct {
 } FuriHalUsbCcidConfig;
 
 typedef struct {
-    void (*icc_power_on_callback)(uint8_t* dataBlock, uint32_t* dataBlockLen, void* context);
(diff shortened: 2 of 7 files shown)


targets/f7/furi_hal/furi_hal_usb_ccid.c
```diff
@@ -332,7 +332,7 @@ void CALLBACK_CCID_IccPowerOn(
             if(callbacks[CCID_SLOT_INDEX] != NULL) {
                 callbacks[CCID_SLOT_INDEX]->icc_power_on_callback(
                     responseDataBlock->abData,
-                    &responseDataBlock->dwLength,
+                    (uint8_t*)&responseDataBlock->dwLength,
```
skotopes, line 335: Quite dangerous operation.

targets/f7/furi_hal/furi_hal_usb_ccid.c
```diff
@@ -364,10 +364,11 @@ void CALLBACK_CCID_XfrBlock(
             if(callbacks[CCID_SLOT_INDEX] != NULL) {
                 callbacks[CCID_SLOT_INDEX]->xfr_datablock_callback(
                     (const uint8_t*)receivedXfrBlock->abData,
-                    receivedXfrBlock->dwLength,
+                    (uint8_t)receivedXfrBlock->dwLength,
                     responseDataBlock->abData,
-                    &responseDataBlock->dwLength,
+                    (uint8_t*)&responseDataBlock->dwLength,
```
skotopes, line 369: and here. 

I know that you checking sizes in EP callback handling code, but still.

targets/furi_hal_include/furi_hal_usb_ccid.h
```diff
@@ -17,12 +17,12 @@ typedef struct {
 } FuriHalUsbCcidConfig;
 
 typedef struct {
-    void (*icc_power_on_callback)(uint8_t* dataBlock, uint32_t* dataBlockLen, void* context);
+    void (*icc_power_on_callback)(uint8_t* dataBlock, uint8_t* dataBlockLen, void* context);
     void (*xfr_datablock_callback)(
         const uint8_t* pcToReaderDataBlock,
-        uint32_t pcToReaderDataBlockLen,
+        uint8_t pcToReaderDataBlockLen,
```
skotopes, line 23: On arm32 smaller size types are not going save space or make code faster. Any specific reasons why you want to have uint8_t here?

skotopes (changes requested):

targets/furi_hal_include/furi_hal_usb_ccid.h
kidbomb: I can change this function back to uint32_t then, since it will return only values up to 255 and it is called only by core components

skotopes: I'm wondering why original structures were designed to have `unit32_t`? Do standard got payloads bigger than 256 bytes?

kidbomb: That's because CCID has two modes: Short APDU and Extended APDU.

In Short APDU mode, packets have size up to 0xFF. 
In Extended APDU mode, packets have size up to 0xFFFF

Althouth devices can have different modes, there is only one property in the structure, which can go up to 0xFF or 0xFFFF in either case

skotopes: I'd keep uint32_t, but final decision is up to you.

kidbomb: I'll keep it as uint32_t, since:

- the CCID buffer size  is already 255 in size
- we are already checking ii dwLength is < 255
- i will add extended apdu support at some point anyway

kidbomb:
I'll keep it as uint32_t, since:

the CCID buffer size is already 255 in size
we are already checking ii dwLength is < 255
i will add extended apdu support at some point anyway

Issue #3646, opened 2024-05-13
125kHz Card Not Reading - Images and Raw Files Attached


davewilliamstx:
### Description of the feature you're suggesting.

Have tried two different versions of this card, not reading on Flipper. I'm not sure on source of the cards, our HOA uses them.

[Unknown RFID Card.zip](https://github.com/flipperdevices/flipperzero-firmware/files/15288222/Unknown.RFID.Card.zip)

![IMG_0804 Large](https://github.com/flipperdevices/flipperzero-firmware/assets/3608077/6940529d-0708-4a1e-abd4-1bf18650bf38)
![IMG_0805 Large](https://github.com/flipperdevices/flipperzero-firmware/assets/3608077/85148bd2-d25a-4fdc-9d7e-a42580f231f4)
![IMG_0806 Large](https://github.com/flipperdevices/flipperzero-firmware/assets/3608077/2395a35a-7cdc-4875-9c46-9a0a1f295ab7)

skotopes:
Yep this one is RFID, payload ASK coded, should be totally doable.

zinongli:
The reading using [flipper-raw-rfid](https://github.com/hnesk/flipper-raw-rfid) from .ask looks similar to #3336
`100000000011010010111111111111111111111011110011011100111011000011011010010111111111111111111111`
Inverting it and parsing spacers gives wiegand 26-bit format
`preamble     ??bitlen   reserved        EPf   fffffffc   cccccccc   cccccccOP  CS1        CS2`
`0111111111 0 01011010 0 00000000 0 00000000 0 01000011 0 01000110 0 01001111 0 01001011 0 10000000 0 00000000 0 0000`
`f` deontes facility code and `c` denotes card id. In this case card id is `1010001100100111` which is equal to 41767 in decimal. The same card number as printed on OP's card. 

Also long pulses are around 330 mu sec given by Flipper Lab's pulse plotter. Close enough to RF/40 clock which would give 320 mu sec. 

I'm not sure if inverting is intentional or if it is my mistake in analyzing the signal. If it is intended to be like the second sequence, then my PR #3697 should enable support for this card as well. 

@davewilliamstx : do you mind taking a photo of the reader? It might help confirming the brand and protocol of this card.

davewilliamstx:
Sure thing, it’s pretty beat up - hope it helps.

zinongli:
[...]

Issue #2744, opened 2023-06-07
125khz RFID Protocol Request


f1yaw4y:
[125_rfid.zip](https://github.com/flipperdevices/flipperzero-firmware/files/11671518/125_rfid.zip)

https://fccid.io/NNHRKWM

Attached are raw readings of my RFID gym access card and a picture of the reader, along with the FCC ID info page. I would attach a picture of the actual card but it is full of identifying information both front and back. It is the size of a typical credit card or HID card and white with printed text / image of my face.

I'd like to request adding the card to the list of recognized protocols. I wish I could contribute but I have no idea how to analyze the file. 

Thank you for the help :)

skotopes:
PSK record looks valid. Should be possible to implement protocol on top of it.

skotopes:
Support added in https://github.com/flipperdevices/flipperzero-firmware/pull/3697 Thanks to @zinongli

PR #3697, opened 2024-06-08
[LFRFID] Added Support for Securakey Protocol

### LINKED ISSUE #2619: LF-RFID Support for RadioKey/Securakey
### Describe the enhancement you're suggesting.

I'm hoping support for SecuraKey/RadioKey can be added to the LF-RFID app. 

There was a post was made in the forum here back in September:
https://forum.flipperzero.one/t/securakey/6434

I'm able to provide raw reads of my own keys. (I have 3 of them that can be tested with)

### Anything else?

Just a huge thank you to the Flipper team for the consistent firmware updates. This is a great tool. 

Edit. Here's a RAW read of one of the tags.
[SecuraKey.zip](https://github.com/flipperdevices/flipperzero-firmware/files/11335453/SecuraKey.zip)

zinongli:
# What's new

- Added reading, emulation, and writing support for Radio Key Securakey protocol.
- Works for 26-bit and 32-bit formats. 
- Reading and emulation verified to be working. 
- I don't own a T5577 so I can't verify writing.
- This should resolve #2619, #2744, potentially #3646, and [all the requests on the forum](https://www.google.com/search?q=securakey+site%3Aforum.flipper.net)

# Verification 

- Build the firmware and flash it onto flipper zero. 
- Read a Securakey fob. Emulate it and try on a reader. Write it onto a T5577 tag. Try the T5577 on the same reader.

# Checklist (For Reviewer)


lib/lfrfid/protocols/.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,19 @@
+{
```
skotopes: Looks like this file shouldn't be a part of PR

lib/lfrfid/protocols/protocol_securakey.c
```diff
@@ -0,0 +1,301 @@
...
+void protocol_securakey_free(ProtocolSecurakey* protocol) {
+    free(protocol->debug_string);
+    free(protocol);
+};
+
+uint8_t* protocol_securakey_get_data(ProtocolSecurakey* protocol) {
+    return protocol->data;
+};
+
+//static const char* protocol_securakey_get_encoded_data_debug(ProtocolSecurakey* protocol) {
```
skotopes: Please remove all commented-out code

lib/lfrfid/protocols/protocol_securakey.c
```diff
@@ -0,0 +1,301 @@
...
+void protocol_securakey_decoder_start(ProtocolSecurakey* protocol) {
+    memset(protocol->encoded_data, 0, SECURAKEY_ENCODED_FULL_SIZE_BYTE);
+    manchester_advance(
+        protocol->decoder_manchester_state,
+        ManchesterEventReset,
+        &protocol->decoder_manchester_state,
+        NULL);
+};
+
+bool protocol_securakey_decoder_feed(ProtocolSecurakey* protocol, bool level, uint32_t duration) {
```
skotopes, line 131: Please use `./fbt format` to format sources

skotopes (changes requested):

lib/lfrfid/protocols/.vscode/c_cpp_properties.json
zinongli: My bad. Deleted.

lib/lfrfid/protocols/protocol_securakey.c
zinongli: Done.

lib/lfrfid/protocols/protocol_securakey.c
zinongli: Done. Thanks for introducing that to me! I didn't know it can do this.

zinongli:
After further research, I learned that wiegand parity bits are, as in the name, parity bits that needs to be calculated. I will reopen this PR after adding the parity calculation in the encoder.

skotopes:
@zinongli see GDB Output: https://github.com/flipperdevices/flipperzero-firmware/actions/runs/9454491882/job/26042147462?pr=3697

Your implementation is failing RFID unit tests. 
use `./fbt LIB_DEBUG=1 FIRMWARE_APP_SET=unit_tests flash_usb_full` and then run `unit_tests` in device cli

skotopes:
Un-draft when ready

zinongli:
@skotopes I'm getting the same result with my CLI as GDB. It hits `test_lfrfid_protocol_em_read_simple()` and just stops moving forward. I'm very new to flipper development. Do you have any tips on where I should be looking at to debug? My assumption is me extending the protocol dictionary makes `test_lfrfid_protocol_em_read_simple()` unhappy. But I couldn't understand why this function should care if the changes I'm making is irrelevant to it.

skotopes:
> @skotopes I'm getting the same result with my CLI as GDB. It hits `test_lfrfid_protocol_em_read_simple()` and just stops moving forward. I'm very new to flipper development. Do you have any tips on where I should be looking at to debug? My assumption is me extending the protocol dictionary makes `test_lfrfid_protocol_em_read_simple()` unhappy. But I couldn't understand why this function should care if the changes I'm making is irrelevant to it.

The crash itself happens in your code. Use `bt` after flipper crash and then check where exactly in your code it crashes.

zinongli:
@skotopes I see. Thank you. If by crash you mean unit_tests making CLI stuck, I couldn't type any commands when it is stuck. If you mean crash in like everyday use, I can't trigger any bug or crash when I flash this build and use it. Do you mind if we talk on discord? I could use some help working with this.

skotopes:
@zinongli sure drop in in voice channel and tag me in chat

skotopes:
- Key generation on flipper produces invalid card
- 5577 write is unsuccessful
- emulating freshly created key crashes proxmark

skotopes:
Un-draft when all fixed

zinongli:
> - Key generation on flipper produces invalid card
> 
> - 5577 write is unsuccessful
> 
> - emulating freshly created key crashes proxmark

I think I can guess the reason for artificially generated keys to be invalid and crash pm3. The decoded data includes one byte to represent the bit-length of encoded data, and it has to be 0x1A or 0x20. If not, the generated key will not be recognized by this decoder itself as it only treats 0x1A and 0x20 bit lengths as can-be-decoded.

I'm surprised the T5577 writing failed. I have reports from another user saying that it succeeded with cloning existing fobs. Just to clarify, is it the writing failed, or is it the written T5577 failed on a reader/pm3? And are you writing an artificially generated key or an existing key? It's probable that the bit-length issue can cause encoder to not function normally on artificial keys as it fails to find correct bit-length.

I will fix the bit length issue to not allow user manually change bit length. This fix should resolve all the problems with flipper-to-flipper, flipper-to-T5577, flipper-to-pm3, flipper-written-T5577-to-pm3 pathways with artificial keys.

However, the last two checksum bytes will still need some more reverse engineering before I can make them calculated based on facility code and card number, instead of just copying from the fobs that's being read like it does now. This problem affects the artificially-generated-key-to-official-reader pathway since only the official reader checks checksum. The primary scenario of this pathway I can think of is someone trying to create a key on flipper based on pm3 reading of an existing fob. So if we can integrate the pipeline to be more compatible with pm3 outputs, it should solve most of the use cases on this pathway. I might need your help on this since I don't own a pm3. By any chance can you send me a pm3 reading of a securakey fob? 

The bit length fix will be done today or tomorrow. Checksum might need more time.

skotopes:
Artificially generated key write failed.

zinongli:
The issues encountered with artificially generated keys should be resolved with the last commit. The bit-length issue is patched. 

I currently can't create a better implementation for checksums due to lack of knowledge on its algorithm. As of the current state, there should be no scenarios that can cause anything to crash. This checksum issue wouldn't affect any usage on genuine fobs since they would have correct checksums. The only possible less than ideal scenario would be when a user artificailly generates a key and can't open a door even though the facility code and card number are correct, due to checksum failure. This issue is not resolved in PM3's repo either. So I propose we leave that part for future improvements. If, in the future, there are discoveries of the algorithm, I will update this code and open a new PR. What do you think?

skotopes (approved):

PR #3721, opened 2024-06-18
[FL-3846] Event Loop Timers

gsurkov:
# What's new

- Add software timer functionality to `FuriEventLoop`
- Add a test blinky application
- Refactor the `Dolphin` application

# Verification 

1. Connect 6 LEDs to pins 2-7 on Flipper (use series resistors of course)
2. Flash the firmware first and then run the test application: `./fbt launch APPSRC=event_loop_blink_test`
3. Observe the blinkenlights! Each button on the Flipper can stop or start a particular LED. Press and hold `Back` to exit.
4. Speed up the Dolphin: compile with `--extra-define=DOLPHIN_DEBUG`, observe the messages in the log.

# Checklist (For Reviewer)


applications/debug/event_loop_blink_test/event_loop_blink_test.c
```diff
@@ -0,0 +1,165 @@
...
+static void view_port_input_callback(InputEvent* input_event, void* context) {
+    EventLoopBlinkTestApp* app = context;
+    furi_message_queue_put(app->input_queue, input_event, 0);
+}
+
+static bool input_queue_callback(FuriMessageQueue* queue, void* context) {
+    EventLoopBlinkTestApp* app = context;
+
+    InputEvent event;
+    FuriStatus status = furi_message_queue_get(queue, &event, FuriWaitForever);
```
skotopes: FuriWaitForever is bad. I think we should prevent people from using Forever if queue uses event loop

furi/core/event_loop.h
```diff
@@ -60,6 +60,138 @@ void furi_event_loop_run(FuriEventLoop* instance);
...
+ *
+ * @param[in,out] context pointer to a user-specific object that was provided during
+ *                        furi_event_loop_pend_callback() call
+ */
+typedef void (*FuriEventLoopPendingCallback)(void* context);
+
+/**
+ * @brief Call a function when all preceding timer commands are processed
+ *
+ * This function may be useful to call another function when the even loop has been started.
```
CookiePLMonster:
```suggestion
 * This function may be useful to call another function when the event loop has been started.
```

targets/f18/api_symbols.csv
```diff
@@ -1098,13 +1098,22 @@ Function,+,furi_event_flag_free,void,FuriEventFlag*
...
 Function,+,furi_event_flag_set,uint32_t,"FuriEventFlag*, uint32_t"
 Function,+,furi_event_flag_wait,uint32_t,"FuriEventFlag*, uint32_t, uint32_t, uint32_t"
-Function,-,furi_event_loop_alloc,FuriEventLoop*,
-Function,-,furi_event_loop_free,void,FuriEventLoop*
-Function,-,furi_event_loop_message_queue_subscribe,void,"FuriEventLoop*, FuriMessageQueue*, FuriEventLoopEvent, FuriEventLoopMessageQueueCallback, void*"
-Function,-,furi_event_loop_message_queue_unsubscribe,void,"FuriEventLoop*, FuriMessageQueue*"
-Function,-,furi_event_loop_run,void,FuriEventLoop*
-Function,-,furi_event_loop_stop,void,FuriEventLoop*
-Function,-,furi_event_loop_tick_set,void,"FuriEventLoop*, uint32_t, FuriEventLoopTickCallback, void*"
+Function,+,furi_event_loop_alloc,FuriEventLoop*,
```
skotopes, line 1108: Bump api symbols version

targets/f7/api_symbols.csv
```diff
@@ -1203,13 +1203,22 @@ Function,+,furi_event_flag_free,void,FuriEventFlag*
...
 Function,+,furi_event_flag_set,uint32_t,"FuriEventFlag*, uint32_t"
 Function,+,furi_event_flag_wait,uint32_t,"FuriEventFlag*, uint32_t, uint32_t, uint32_t"
-Function,-,furi_event_loop_alloc,FuriEventLoop*,
-Function,-,furi_event_loop_free,void,FuriEventLoop*
-Function,-,furi_event_loop_message_queue_subscribe,void,"FuriEventLoop*, FuriMessageQueue*, FuriEventLoopEvent, FuriEventLoopMessageQueueCallback, void*"
-Function,-,furi_event_loop_message_queue_unsubscribe,void,"FuriEventLoop*, FuriMessageQueue*"
-Function,-,furi_event_loop_run,void,FuriEventLoop*
-Function,-,furi_event_loop_stop,void,FuriEventLoop*
-Function,-,furi_event_loop_tick_set,void,"FuriEventLoop*, uint32_t, FuriEventLoopTickCallback, void*"
+Function,+,furi_event_loop_alloc,FuriEventLoop*,
```
skotopes, line 1213: and here

furi/core/event_loop.c
```diff
@@ -186,13 +454,19 @@ void furi_event_loop_run(FuriEventLoop* instance) {
                         }
                     }
                 }
+
+                furi_event_loop_restore_flags(instance, flags & ~FuriEventLoopFlagEvent);
```
skotopes, line 162: why?

furi/core/event_loop.c
```diff
@@ -145,30 +218,225 @@ static FuriEventLoopProcessStatus
...
 
         instance->state = FuriEventLoopStateProcessing;
+
         if(ret == pdTRUE) {
             if(flags & FuriEventLoopFlagStop) {
-                instance->state = FuriEventLoopStateIdle;
+                instance->state = FuriEventLoopStateStopped;
                 break;
+
             } else if(flags & FuriEventLoopFlagEvent) {
```
skotopes, line 133: else if -> if ?

applications/debug/event_loop_blink_test/event_loop_blink_test.c
CookiePLMonster: As someone who is trying to study the usage of the event loop - why? This call is executed when there is something on the queue, so it should always immediately get some data.

Or is this to account for the cases of type "if this `queue_get` fails, then something is very wrong and you should crash instead of blocking endlessly"?

skotopes: we want to know if event loop was correctly designed, so we can fail early, FuriWaitForever will hide it from us

furi/core/event_loop.c
gsurkov: Makes no difference, since there is a break statement in the branch above.

furi/core/event_loop.c
gsurkov: The flags are being cleared upon each call to `xTaskNotifyWaitIndexed()`, so in case there is a pending `FuriEventLoopFlagRequest` together with a `FuriEventLoopFlagEvent`, the former will not get processed, so we need to restore all flags except `FuriEventLoopFlagEvent`. 

This is done to ensure that event processing gets higer priority than requests (timers, pending callbacks etc), i.e. they will not receive control until there are no more events to process.

If this is not a desirable behaviour, I can change it to a sequential one - everything will be processed in one fell swoop.

applications/debug/event_loop_blink_test/event_loop_blink_test.c
gsurkov: > I think we should prevent people from using Forever if queue uses event loop.

True, but it is not trivial. I propose we do it in the next phase where we stabilise the primitive API.

skotopes (approved):

PR #3730, opened 2024-06-21
[FL-3847, FL-3513] Thread Signals

gsurkov:
# What's new

- Add a signal API for threads
- Integrate signals into the `Loader` and `EventLoop` subsystems
- Add `loader close` CLI subcommand
- Show the application name when `loader info` is executed (FL-3513)

# Verification 

1. Start any `EventLoop`-based application, for example: `loader open NFC`
2. Run the `loader info` command. The application name should be printed.
3. Run the `loader close` command. The application should exit and the corresponding message should be printed.
4. Start any application that does not use the `EventLoop` subsystem.
5. Repeat step 3. This time, an error message should be printed.

# Checklist (For Reviewer)


CookiePLMonster:
Can this be expanded to send arbitrary (user) signals from CLI to the application? Everything in this PR appears to be ready for that, except for `loader_cli`.

An ability to send arbitrary signals via e.g. `loader signal 101 [optional_args]` could come in handy for debugging, especially without a developer board. For example, one could use those signals to toggle debug messages on/off without the need to employ awkward button combinations, to skip to a specific part of the application, or to toggle a developer mode.

EDIT:
One more remark - when I first saw the PR, I thought those signals are going to execute in the context of the thread a signal is sent to. It might be worth mentioning in the docs that this is not the case, and the signaled thread may run concurrently with the signal function.

Realistically I don't see a way to ensure that this callback runs in the correct context, but for the end user it's easy if they just serialize signals via a queue.

applications/services/loader/loader.h
```diff
@@ -88,6 +88,26 @@ void loader_show_menu(Loader* instance);
...
+ * @param[in] instance pointer to the loader instance
+ * @param[in] signal signal value to be sent
+ * @param[in,out] arg optional argument (can be of any value, including NULL)
+ *
+ * @return true if the signal was handled by the application, false otherwise
+ */
+bool loader_signal(Loader* instance, uint32_t signal, void* arg);
+
+/**
+ * @brief Get the name of the currently running applicaton
```
CookiePLMonster:
```suggestion
 * @brief Get the name of the currently running application
```

applications/services/loader/loader_cli.c
```diff
@@ -60,6 +64,20 @@ static void loader_cli_open(FuriString* args, Loader* loader) {
...
 
+static void loader_cli_close(Loader* loader) {
+    FuriString* app_name = furi_string_alloc();
+
+    if(!loader_get_application_name(loader, app_name)) {
+        printf("No application is running\r\n");
+    } else if(!loader_signal(loader, FuriSignalExit, NULL)) {
+        printf("Application \"%s\" has to be closed manually\r\n", furi_string_get_cstr(app_name));
+    } else {
+        printf("Applicaton \"%s\" was closed\r\n", furi_string_get_cstr(app_name));
```
CookiePLMonster:
```suggestion
        printf("Application \"%s\" was closed\r\n", furi_string_get_cstr(app_name));
```

gsurkov: Thanks, I should really use a spell checker.

applications/services/gui/view_dispatcher.c
```diff
@@ -326,6 +336,44 @@ void view_dispatcher_handle_custom_event(ViewDispatcher* view_dispatcher, uint32
...
+
+    if(instance->event_loop) {
+        furi_event_loop_stop(instance->event_loop);
+        is_consumed = true;
+    }
+
+    return is_consumed;
+}
+
+bool view_dispatcher_handle_signal_event(uint32_t signal, void* arg, void* context) {
```
skotopes: Kinda breaks contract between view dispatcher and application. Should be delivered through the message queue.

skotopes: Also I don't think that view dispatcher should be responsible for signal delivery. User can override callback directly in thread api

skotopes (changes requested):

applications/services/gui/view_dispatcher.c
CookiePLMonster: However, with the current change, how do apps register for signals while also keeping the default View Dispatcher behaviour?

skotopes: @CookiePLMonster keep in mind that this PR is still WIP.

Here is what I imagine in my head:
- Thread signal subscription is going to be low level API, you can use it in applications without EventLoop
- Otherwise you use event loop signals API, which is going to arrive a little bit later

I'll make one more iteration over this PR on this weekend, and then call you for final review

gsurkov:
> An ability to send arbitrary signals via e.g. loader signal 101 [optional_args] could come in handy

done. 

> I thought those signals are going to execute in the context of the thread a signal is sent to.

No, this is just a settable callback, the recipient is responsible for queueing the request, if necessary, on their side (this highly depends on the concrete implementation of the handling mechanism).

applications/services/gui/view_dispatcher.c
CookiePLMonster: @skotopes Fair enough, the PR is not drafted so I considered it final-ish 😁

CookiePLMonster:
> done.

Great! Just one nitpick - if instead of `%p` the argument became a pointer to the first non-whitespace character after the signal number (or NULL if that character happens to be `\0`), the arguments would become arbitrary. It'd then be up to the app to treat it as a string, parse as integer etc.

You should be able to easily do it with `strtoul` (as it gives a pointer to the first character after the parsed integer) + `while (isspace(*ch)) ch++;`

skotopes (approved):

applications/services/gui/view_dispatcher.c
skotopes: Check now. There will be more polishing in future, but on concept level it's what I wanted.

CookiePLMonster: Looks good - and I assume that `furi_event_loop_signal_callback` will in the future propagate the signal down to user-registered signals via the upcoming signals API, right?

skotopes: Yep.

gsurkov:
@CookiePLMonster nice idea, we'll see how to best implement it a bit later.

CookiePLMonster:
> @CookiePLMonster nice idea, we'll see how to best implement it a bit later.

Sounds good! As I mentioned, `strtoul` (instead of `sscanf`) will do most of the job for you since it provides an output variable pointing at the character the function finished parsing at. From there it's just the matter of scanning for the first non-whitespace char, which will be either the user-supplied string or `\0`. The revised approach shouldn't be any more complex than the current `sscanf`.

Issue #3701, opened 2024-06-10
Feature Request: NFC 14443a I-Block chaining


bettse:
### Description of the feature you're suggesting.

Based on my read of iso14443_4_layer_decode_block (https://github.com/flipperdevices/flipperzero-firmware/blob/dev/lib/nfc/helpers/iso14443_4_layer.c#L49-L63) I believe the flipper is lacking I-Block chaining support.  I've done a dump of a card interaction and the first is 0x12 (0b00010010) which I think is chaining based on this: https://github.com/RfidResearchGroup/proxmark3/blob/master/armsrc/iso14443a.c#L2953 .  I'm by no means a low level 14a dev, so I might be wrong.

I'd love to have I-block chaining so I could have further interaction with this javacard.

gornekich:
Working on this right now. Should be ready this week

gornekich:
Still working on it, should be ready on next release

P-L-K:
> Still working on it, should be ready on next release

Does the release of 1.0 help?

bettse:
Someone linked me https://github.com/flipperdevices/flipperzero-firmware/pull/3918

Tested my ndef app with dev and 🎉  no iblock problem anymore!

PR #3744, opened 2024-07-02
NFC ISO14443a-4 I-Block Chaining support

bettse:
# What's new

- Support I-Block Chaining (#3701)

Doesn't actually work `bit_buffer_append_right` fails when the destination is non-empty.  No crash according to the UART logs or the UI, but the flipper freezes in whatever state it was in.  

- Works: `30758 [D][iso14443_4_layer] used=0, capacity=384, available=384, needed=175`
- Fails: `30947 [D][iso14443_4_layer] used=252, capacity=384, available=132, needed=7` (And this is the last line of the UART logs until reboot using Left+Back)

# Verification 

- Read something ISO14443a-4 that does use I-block chaining

# Checklist (For Reviewer)

### DIFF
--- a/lib/nfc/helpers/iso14443_4_layer.c
+++ b/lib/nfc/helpers/iso14443_4_layer.c
@@ -2,8 +2,12 @@
 
 #include <furi.h>
 
+#define TAG "iso14443_4_layer"
+
 #define ISO14443_4_BLOCK_PCB (1U << 1)
 #define ISO14443_4_BLOCK_PCB_I (0U)
+#define ISO14443_4_BLOCK_PCB_I_MASK (0x0FU)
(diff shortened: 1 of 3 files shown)


lib/nfc/helpers/iso14443_4_layer.c
```diff
@@ -54,11 +65,39 @@ bool iso14443_4_layer_decode_block(
...
+            if(needed > available) break;
+            FURI_LOG_D(
+                TAG,
+                "used=%zu, capacity=%zu, available=%zu, needed=%zu",
+                used,
+                capacity,
+                available,
+                needed);
+
+            bit_buffer_append_right(output_data, block_data, 1);
```
bettse, line 89: This is where it fails if output_data is non-empty (even when output_data has more than enough room for block_data)

PR #3748, opened 2024-07-02
input_srv: Put input state data on the stack of the service

CookiePLMonster:
# What's new

`input_srv` did two allocations on startup that span the entire lifetime of the service, and thus the entire session. Since this service has a 1KB stack, let's make use of that by putting this data on the stack instead.

# Verification 

Verify that input still works.

# Checklist (For Reviewer)


CookiePLMonster:
@hedger With these changes in place, should `input_i.h` go? `input_cli` does not need the `Input` structure anymore, and so the entire contents of this file could be moved to `input.c`.

hedger:
@CookiePLMonster I think so - since now we can pass pubsub pointer to CLI code.

CookiePLMonster:
> @CookiePLMonster I think so - since now we can pass pubsub pointer to CLI code.

Done.

CookiePLMonster:
I now made the `counter` start from `1` and not `0`, as I realized that lets me use post-increment rather than pre-increment while keeping the behaviour unchanged.

skotopes (approved):

PR #3757, opened 2024-07-05
LFRFID: add DEZ 8 format to EM4100 info screen

korden32:
# What's new

- DEZ 8 is a popular representation of EM4100' ID and usually printed/embossed on original keyfobs.

# Verification 

- Reading or opening saved EM4100 card should now shows DEZ 8 representation in addition to existing FC+Card.
I should also note small mess between DEZ 8 and DEZ 10 representation. Seems like due to widespread usage of legacy Wiegand-26 interface (3 bytes of data) most access systems and other hardware actually shows DEZ8 code but in a form of DEZ10. This can be easily checked with EM4100 card "00 01 00 00 01" (DEZ8=00000001; DEZ10=0016777217).

# Checklist (For Reviewer)


skotopes:
There are too many formats for card representation. We don't think that having DEZ 8 or DEZ 10 will be enough. That's why we are not going to accept this PR.

PR #3781, opened 2024-07-10
Tar archive: fix double free

DrZlo13:
# What's new

- no more double free

# Verification 

- idk, еру current allocator doesn't seem to care

# Checklist (For Reviewer)

### DIFF
--- a/lib/toolbox/tar/tar_archive.c
+++ b/lib/toolbox/tar/tar_archive.c
@@ -91,7 +91,6 @@ static int mtar_heatshrink_file_close(void* stream) {
             compress_stream_decoder_free(hs_stream->decoder);
         }
         storage_file_close(hs_stream->stream);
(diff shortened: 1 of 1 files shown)


hedger (approved):

CookiePLMonster:
Does `storage_file_close` free memory too? If not, shouldn't `storage_file_free` be the one that is kept, considering the docs say freeing the storage also closes it?

WillyJL:
It's freed when freeing TarArchive struct @CookiePLMonster. The function being modified in this PR is called inside mtar library when doing mtar_close(), so TarArchive free does mtar_close() then also frees the file again

CookiePLMonster:
> It's freed when freeing TarArchive struct @CookiePLMonster

Thanks for the clarification, that's what I get for checking the diff only and not the full file =)

PR #3783, opened 2024-07-12
Improve bit_buffer.h docs

### LINKED ISSUE #3608: bit_buffer Documentation improvement
### Describe the enhancement you're suggesting.

Two things:
- I think the `bit_buffer_get_parity` documentation was copied from `bit_buffer_get_data` and not changed: https://github.com/flipperdevices/flipperzero-firmware/blob/dev/lib/toolbox/bit_buffer.h#L224-L229
- Could examples be added to the header doc comment for `bit_buffer_write_bytes_with_parity` and `bit_buffer_copy_bytes_with_parity`?  The system I am working with has parity integrated as every the 9th bit in the bit stream, (`DDDDDDDD PDDDDDDD DPDDDDDDD DP...`) which has the parity starting at the most significant bit and gradually moving down, whereas I believe (https://github.com/flipperdevices/flipperzero-firmware/blob/dev/lib/toolbox/bit_buffer.c#L113-L114) the bit_buffer methods put the 0th parity bit at the 0th (least significant) position in the byte to start with.  I recognize that either is valid, and it was just after testing the bit_buffer version I realized it didn't work for me.  A documentation example could have saved me a little time.

### Anything else?

_No response_

Astrrra:
# What's new

- Fixes #3608 

# Verification 

- Read through the new comments, check that they are correct and match reality

# Checklist (For Reviewer)


lib/toolbox/bit_buffer.h
```diff
@@ -98,6 +98,8 @@ void bit_buffer_copy_bits(BitBuffer* buf, const uint8_t* data, size_t size_bits)
  * @param [in,out] buf pointer to a BitBuffer instance to copy into
  * @param [in] data pointer to the byte array to be copied
  * @param [in] size_bitss size of the data to be copied, in bits
+ * @note Parity bits are placed starting with the least significant bit of each byte and moving up.
```
bettse: I think this is the "most" significant bit (https://en.wikipedia.org/wiki/Bit_numbering)
(I think the convention is based on non-binary arabic numbering where the left side contains the most significant place.)

lib/toolbox/bit_buffer.h
```diff
@@ -10,134 +16,146 @@ extern "C" {
...
- * @param [in] size_bitss size of the data to be copied, in bits
+ * @warning       The destination capacity must be no less than the source data
+ *                size.
+ *
+ * @param[in,out] buf        pointer to a BitBuffer instance to copy into
+ * @param[in]     data       pointer to the byte array to be copied
+ * @param[in]     size_bits  size of the data to be copied, in bits
+ * @note          Parity bits are placed starting with the least significant bit
+ *                of each byte and moving up.
+ * @note          Example: PDDDDDDD DPDDDDDD DDPD...
```
gornekich: The parity bit follows the 8 data bits and is packed into an uint8_t array.
DDDDDDDDP     DDDDDDDDP     DDDDDDDDP ... -> DDDDDDDD     PDDDDDDD     DPDDDDDD     DDP ....
MSB -> LSB

gornekich: Please fix the description

lib/toolbox/bit_buffer.h
```diff
@@ -10,134 +16,146 @@ extern "C" {
...
+ *             destination capacity must be no less than the source data size
+ *             plus parity.
+ *
+ * @param[in]  buf           pointer to a BitBuffer instance to write from
+ * @param[out] dest          pointer to the destination memory location
+ * @param[in]  size_bytes    maximum destination data size, in bytes
+ * @param[out] bits_written  actual number of bits written, in bits
+ * @note       Parity bits are placed starting with the least significant bit of
+ *             each byte and moving up.
+ * @note       Example: PDDDDDDD DPDDDDDD DDPD...
```
gornekich: Same here

gornekich (changes requested):

lib/toolbox/bit_buffer.h
gornekich: As mentioned above, this is MSB

skotopes (approved):

gornekich (approved):

lib/toolbox/bit_buffer.h
skotopes: Fixed, nice catch

PR #3785, opened 2024-07-14
CCID: Add CCIDWorker

kidbomb:
This new thread allows for Flipper to receive CCID messages bigger than CCID_EPSIZE by receiving multiple packaged in a row and adding them to the buffer

# What's new

- A new thread was added (CcidWorker)
- CCID now works as a producer-consumer: ccid_tx_ep_callback acts as a producer, adding items into the buffer, while the worker thread removed them once complete
- Allows for CCID data payloads up to 0xF3 to be sent

# Verification 

 - Start CCID Test application
 - Run the ccid_client.py script under applications/debug/ccid_test/client
 - Select the 'Generic USB Smart Card Reader'
 - All tests should pass

# Checklist (For Reviewer)

### DIFF
--- a/targets/f7/furi_hal/furi_hal_usb_ccid.c
+++ b/targets/f7/furi_hal/furi_hal_usb_ccid.c
@@ -435,15 +435,15 @@ void furi_hal_ccid_send_packet(uint8_t* data, uint8_t len) {
 }
 
 void furi_hal_ccid_send_response(uint8_t* data, uint32_t len) {
-    uint32_t dataToSendLen = len;
-    uint32_t dataIndex = 0;
-    while(dataToSendLen >= CCID_EPSIZE) {
-        furi_hal_ccid_send_packet(&data[dataIndex], CCID_EPSIZE);
-        dataToSendLen = dataToSendLen - CCID_EPSIZE;
-        dataIndex = dataIndex + CCID_EPSIZE;
+    uint32_t data_to_send = len;
+    uint32_t data_index = 0;
+    while(data_to_send >= CCID_EPSIZE) {
+        furi_hal_ccid_send_packet(&data[data_index], CCID_EPSIZE);
+        data_to_send = data_to_send - CCID_EPSIZE;
+        data_index = data_index + CCID_EPSIZE;
     }
 
-    furi_hal_ccid_send_packet(&data[dataIndex], dataToSendLen);
+    furi_hal_ccid_send_packet(&data[data_index], data_to_send);
 }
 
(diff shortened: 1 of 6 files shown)


targets/f7/furi_hal/furi_hal_usb_ccid.c
```diff
@@ -186,6 +199,12 @@ static bool connected = false;
 static bool smartcard_inserted = true;
 static CcidCallbacks* callbacks[CCID_TOTAL_SLOTS] = {NULL};
 static void* cb_ctx[CCID_TOTAL_SLOTS];
+static FuriThread* ccidThread;
```
skotopes: I think it's time to cleanup this part:
- encapsulate everything into FuriHalUsbCcid object
- allocate buffers only when you need them
- naming convention

skotopes (changes requested):

targets/f7/furi_hal/furi_hal_usb_ccid.c
kidbomb: Will do - my main concern with buffer is that we would end up with a fragmented heap memory if we perform too many sequential allocations/de-allocations (specially given the fact we are dealing with an embedded device).

skotopes: Check furi_hal_subghz_start_async_tx, it also allocates buffer but only when subghz susbsystem is used.

kidbomb: That aligns with what I had in mind - thanks

skotopes:
There is a use after free in your code.

I've made couple changes to highlight problem and show way to defend your code from it:  https://github.com/flipperdevices/flipperzero-firmware/pull/3785/commits/7033607f6ec9d7b6bed4df2c5dee20541cddacb5 

Open and then close app to crash flipper.

Also please fix naming: file furi_hal_usb_ccid.h must contain APIs that starts with furi_hal_usb_ccid, things like furi_hal_ccid_ccid_remove_smartcard are clearly violating this rule. Naming things in that way makes debugging much easier.

kidbomb:
Not sure what this build error is about. My local build seems to be running fine. Any clues?

skotopes:
> Not sure what this build error is about. My local build seems to be running fine. Any clues?

No problem, I'll fix it by my self.

skotopes (approved):

skotopes:
I'll merge current version as is. One more recommendation: it's much safer to perform all operations with endpoints from one thread.

PR #3822, opened 2024-08-04
MIFARE Classic Key Recovery Improvements

noproto:
# What's new

- MIFARE Classic Accelerated dictionary attack: dictionary attacks reduced to several seconds - checks ~3500 keys per second
- MIFARE Classic Nested attack support: collects nested nonces to be cracked by MFKey, no longer requiring users to downgrade to FW 0.93.0
- MIFARE Classic Static encrypted backdoor support: collects static encrypted nonces to be cracked by MFKey using NXP/Fudan backdoor, allowing key recovery of all non-hardened MIFARE Classic tags on-device

# Verification

This PR currently contains the minimum necessary code to achieve the intended functionality. It will be superseded with performance improvements.

- Accelerated dictionary attack: Benchmark against standard dictionary attack
  - Note: This PR adds nonce collection methods which trade some of the time reclaimed by this improvement
- Nested (weak PRNG): Using any weak PRNG tag and MFKey (v3.0 available for testing [here](https://github.com/flipperdevices/flipperzero-good-faps/pull/243))
- Hardnested (hard PRNG): Verify nonces are stored at /ext/nfc/.nested.log (can be cracked using [HardnestedRecovery](https://github.com/noproto/HardnestedRecovery#usage))
- Static encrypted ([backdoor](https://eprint.iacr.org/2024/1275)): Using static encrypted tag and MFKey

# Checklist (For Reviewer)


noproto:
This PR also resolves `bit_buffer_copy_bytes_with_parity` improperly storing parity bits: https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/8dd3daf6257a1047a31d5ef30b3efb1d24620524#diff-1d0dbb15ed26364f785d68ba753267b5326c95629d29641ef53276ac32336f7e

noproto:
The accelerated dictionary attack is mostly working. I'm tracking down a minor bug in it and making sure the UI reflects the state of the attack.

The fork takes an average of 10 seconds to run dictionary attacks in my tests (1 second of the average is backdoor detection - separate from the dictionary attack). OFW 0.105.0 takes an average of 3 minutes 10 seconds. On real tags the number of unknown keys and the offset in the dictionary are all different, here are five benchmarks on random MIFARE Classic tags:

```
OFW 0.105.0: 4 min 44 seconds (found 19/32 keys, 16 sectors read)
nestednonces 26845cb: 10 seconds (found 19/32 keys, 16 sectors read)
---
OFW 0.105.0: 5 min 8 seconds (found 18/32 keys, 2 sectors read) 
nestednonces 26845cb: 25 seconds (found 18/32 keys, 2 sectors read)
---
OFW 0.105.0: 22 seconds (found 32/32 keys, 16 sectors read) 
nestednonces 26845cb: 3 seconds (found 32/32 keys, 16 sectors read)
---
OFW 0.105.0: 35 seconds (found 32/32 keys, 16 sectors read) 
nestednonces 26845cb: 10 seconds (found 32/32 keys, 16 sectors read)
---
OFW 0.105.0: 5 min 3 seconds (found 18/32 keys, 16 sectors read) 
nestednonces 26845cb: 10 seconds (found 18/32 keys, 16 sectors read)
```

noproto:
One more benchmark from @equipter on a tag with all random dictionary keys: 

```
OFW 0.105.0: 8 min 2 seconds (found 32/32 keys, 16 sectors read) 
nestednonces 26845cb: 17 seconds (found 32/32 keys, 16 sectors read)
```

WillyJL:
gonna add my own benchmark:

mfc 1k 4b uid tag with all keys in dict:
```
CFW (same NFC as OFW 0.105.0), 4117 keys in dict:   10:00.00, all keys and sectors
nestednonces 26845cb, 1277 keys in dict:            00:18.61, all keys and sectors
nestednonces 26845cb, 4117 keys in dict:            00:22.51, all keys and sectors
```
mfc 1k 7b uid tag with only 1 known key in dict:
```
CFW (same NFC as OFW 0.105.0), 4117 keys in dict:   1 hour +, 1 key and sector
nestednonces 26845cb, 1277 keys in dict:            00:50.06, 1 key and sector
nestednonces 26845cb, 4117 keys in dict:            01:36.46, 1 key and sector
```

truly incredible work @noproto :rocket:

noproto:
Still on track to be completed within 4 days. PR will be ready for review then.

noproto:
I’d consider this PR to be beta quality now. A review would be helpful in guiding the next steps and prioritizing fixes.

Some FAQ:

~~**How do I crack the nonces?**: I am working on [MFKey 3.0](https://github.com/flipperdevices/flipperzero-good-faps/tree/dev/mfkey) and a fork of [FlipperNestedRecovery](https://github.com/AloneLiberty/FlipperNestedRecovery). These will be the primary methods of key recovery when I am finished.~~ Solved: https://github.com/flipperdevices/flipperzero-good-faps/pull/243 https://github.com/noproto/HardnestedRecovery

~~**It gets stuck on collecting nonces.**: This is a calibration issue and will be fixed. The process isn't actually stuck — it's taking longer than expected due to difficulties finding a candidate range.~~ Solved: https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/901bdf96921c4b2e9bc02e12c424d475a05bc9f2

~~**It crashes after a few reads.**: There may be a memory leak. Any help in identifying this issue would be appreciated.~~ Solved: https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/13411da449ec9632b12b1db48ee4dcceeb42c063

~~**There's no progress bar.**: This is a UI issue and will be fixed. In the meantime, you can track progress using the header, which shows the current phase and how many keys have been recovered.~~ Solved: https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/c1cdd491a6877bbf50a06eb955c1329e99d85848 You can also interrupt nonce collection, if you want to only run the accelerated dictionary attack/backdoor read. I'd like to add a preference to toggle nonce collection on/off (default: on).

~~**Hardnested is slow.**: I'll be optimizing this over the next few days. The slowdown is related to how Hardnested nonces are collected, but I've identified several ways to possibly improve it. I will run some tests to find the fastest/most reliable approach.~~ Solved: https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/18f8cfbef07ad8fd25f8e06e633176604ea82da1

targets/f7/api_symbols.csv
```diff
@@ -2040,6 +2041,7 @@ Function,-,initstate,char*,"unsigned, char*, size_t"
 Function,+,input_get_key_name,const char*,InputKey
 Function,+,input_get_type_name,const char*,InputType
 Function,-,iprintf,int,"const char*, ..."
+Function,+,is_weak_prng_nonce,_Bool,uint32_t
```
skotopes: We need to prefix all symbols in crypto1 package if symbols are exported to api symbol table

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
```diff
@@ -5,6 +5,8 @@
 
 #define TAG "NfcMfClassicDictAttack"
 
+// TODO: Update progress bar with nested attacks
```
skotopes: TODOs will fail CI/CD lint stage, normally we don't leave them without appropriate issue in task tracker.

lib/nfc/protocols/mf_classic/mf_classic_poller.c
```diff
@@ -680,24 +811,49 @@ NfcCommand mf_classic_poller_handler_key_reuse_start(MfClassicPoller* instance)
...
-            instance->mfc_event_data.key_attack_data.current_sector =
-                dict_attack_ctx->reuse_key_sector;
-            command = instance->callback(instance->general_event, instance->context);
+            dict_attack_ctx->reuse_key_sector++;
+            if(dict_attack_ctx->reuse_key_sector == instance->sectors_total) {
+                instance->mfc_event.type = MfClassicPollerEventTypeKeyAttackStop;
+                command = instance->callback(instance->general_event, instance->context);
+                // Nested entrypoint
+                // TODO: Ensure nested attack isn't run if tag is fully read
+                if(dict_attack_ctx->nested_phase == MfClassicNestedPhaseNone ||
```
skotopes: image

lib/nfc/protocols/mf_classic/mf_classic_poller_i.h
```diff
@@ -3,13 +3,39 @@
 #include "mf_classic_poller.h"
 #include <lib/nfc/protocols/iso14443_3a/iso14443_3a_poller_i.h>
 #include <bit_lib/bit_lib.h>
+#include "nfc/helpers/iso14443_crc.h"
```
skotopes: `<>` - for global includes(outside of this file folder)
`""` - for local includes(in this file folder)

lib/toolbox/bit_buffer.c
```diff
@@ -113,7 +113,7 @@ void bit_buffer_copy_bytes_with_parity(BitBuffer* buf, const uint8_t* data, size
             uint8_t bit =
                 FURI_BIT(data[bits_processed / BITS_IN_BYTE + 1], bits_processed % BITS_IN_BYTE);
 
-            if(bits_processed % BITS_IN_BYTE) {
+            if((bits_processed % BITS_IN_BYTE) == 0) {
```
skotopes, line 116: unit tests maybe?

skotopes (changes requested):

skotopes:
@noproto that is some very cool thing that you've made )

skotopes:
We'll include this one in next release after @gornekich review and QA testing. 
Also we may need to free some RAM to make it more stable, I'll think about it.

lib/nfc/protocols/mf_classic/mf_classic_poller.c
```diff
@@ -505,6 +532,126 @@ NfcCommand mf_classic_poller_handler_request_read_sector_blocks(MfClassicPoller*
...
+    dict_attack_ctx->current_key = mf_classic_backdoor_keys[next_key_index].key;
+
+    // Attempt backdoor authentication
+    MfClassicError error = mf_classic_poller_auth(
+        instance, 0, &dict_attack_ctx->current_key, MfClassicKeyTypeA, NULL, true);
+    if((next_key_index == 0) &&
+       (error == MfClassicErrorProtocol || error == MfClassicErrorTimeout)) {
+        FURI_LOG_E(TAG, "No backdoor identified");
+        dict_attack_ctx->backdoor = MfClassicBackdoorNone;
+        instance->state = MfClassicPollerStateRequestKey;
```
gornekich: @noproto Could you check this part?

Now I can see that you check only first backdoor key (MfClassicBackdoorAuth3), and if backdoor auth is failed we don't check other 2 known backdoor keys. Also the `instance->mode_ctx.dict_attack_ctx` is always 0-initialized at this point, so I think the `current_key_index` will be always `mf_classic_backdoor_keys_count - 1`

gornekich (commented):
Hi @noproto
Thank you for working on that! I started reviewing the code, I will come back with the review soon

lib/nfc/protocols/mf_classic/mf_classic_poller.c
noproto: It depends on how backdoor authentication fails. If the tag does not respond with a nonce to the backdoor authentication command, we skip detection (as it is an unrecognized command by the tag) and save time. If it does, we continue unless we've failed to authenticate with all known backdoor keys. current_keys_index is initialized to 2 initially, so that the "next" key is 0 (since `current_key` is initially not a backdoor key) and starts the loop through the backdoor key array.

Verified with v1, v3, and a tag with no backdoor present:

v1:
```
59810309 [E][MfClassicPoller] Trying backdoor v3
59810423 [E][MfClassicPoller] Trying backdoor v1
59810428 [E][MfClassicPoller] Backdoor identified: v1
```

v3:
```
20070 [E][MfClassicPoller] Trying backdoor v3
20075 [E][MfClassicPoller] Backdoor identified: v3
```

No backdoor:
```
15543 [E][MfClassicPoller] Trying backdoor v3
15547 [E][MfClassicPoller] No backdoor identified
```

noproto: `MfClassicErrorProtocol` is the error code when a nonce is not sent by the tag. Let me know if I can test anything else for this.

noproto:
~Here is what I'll resolve before moving this PR back out of draft, will keep this list updated:~

* ~~Fix crash when skipping tag, then re-reading it~~
* ~~Make Hardnested nonce collection faster~~ (Faster now but we can still buffer log writes for another 20%-30% speedup)
* Resolve all TODOs in code (15 remaining)
* ~~Fix calibration issue, and eliminate outliers~~
* ~~Add UI progress bar~~
* Resolve all reviews (3 remaining)

Planning to open this PR on Sept 24, after final feature implemented (~~NFC app to validate static encrypted cracked keys without nested attacks~~).

noproto:
@skotopes @gornekich Ready for verification of the functionality and QA testing. MFKey 3.0 is available at https://github.com/flipperdevices/flipperzero-good-faps/pull/243

While testing please note the file browser state consumes several kilobytes of RAM that would be useful for MFKey performance-wise (quick access apps and `ufbt launch` from the desktop often demonstrates the full speed behavior). The memory may be free if its launched by the NFC app as mentioned on the Miro 1.1 roadmap/https://github.com/flipperdevices/flipperzero-firmware/pull/3874#issuecomment-2335174117

noproto:
NFC Magic will need the `MfClassicPollerModeDictAttack` identifier in scenes/nfc_magic_scene_mf_classic_dict_attack.c L45 changed to `MfClassicPollerModeDictAttackStandard` as well as a version bump: https://github.com/flipperdevices/flipperzero-good-faps/blob/dev/nfc_magic/scenes/nfc_magic_scene_mf_classic_dict_attack.c#L45 (thanks scorpionul22 and RM!).

noproto:
The complete (draft) process is documented here: https://github.com/noproto/flipper-community-wiki/blob/master/flip-wiki/docs/mifareclassic.md

We'll need the MFKey flow (launched by the NFC app) as listed on the 1.1 roadmap for the documentation to be accurate.

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
```diff
@@ -1,11 +1,16 @@
...
 
+#include <bit_lib/bit_lib.h>
 #include <dolphin/dolphin.h>
 #include <lib/nfc/protocols/mf_classic/mf_classic_poller.h>
 
 #define TAG "NfcMfClassicDictAttack"
 
+// TODO: Fix lag when leaving the dictionary attack view after Hardnested
+// TODO: Re-enters backdoor detection between user and system dictionary if no backdoor is found
+
```
RebornedBrain, line 10: What about these TODOs? Are they still actual or not?

noproto: They are actual, but not for this PR necessarily. I have them documenting the planned performance and code style improvements. Ideally with a follow-up PR, so users have a working process before I've had the chance to fully optimize it.

noproto: More context: to the best of my knowledge the PR is stable and testing is appreciated. The cost of the performance TODOs will add up to milliseconds for the vast majority of attacks.

For the two TODOs referenced here, I've observed under certain conditions Hardnested taking a few seconds to return from the dictionary attack. The backdoor re-entry cost is somewhere in the order of milliseconds.

lib/nfc/protocols/mf_classic/mf_classic_poller.c
```diff
@@ -829,6 +1033,1081 @@ NfcCommand mf_classic_poller_handler_key_reuse_read_sector(MfClassicPoller* inst
...
+               stream, MF_CLASSIC_NESTED_LOGS_FILE_PATH, FSAM_WRITE, FSOM_OPEN_APPEND))
+            break;
+
+        bool params_write_success = true;
+        for(size_t i = 0; i < nonce_pair_count; i++) {
+            MfClassicNestedNonce* nonce = &dict_attack_ctx->nested_nonce.nonces[i];
+            // TODO: Avoid repeating logic here
+            uint8_t nonce_sector = nonce->key_idx / (weak_prng ? 4 : 2);
+            MfClassicKeyType nonce_key_type =
+                (nonce->key_idx % (weak_prng ? 4 : 2) < (weak_prng ? 2 : 1)) ? MfClassicKeyTypeA :
```
RebornedBrain, line 1726: Maybe this should be replaced? According to comments

noproto: @RebornedBrain I was asking if I could submit those as a separate PR to further refactor the code and improve performance (to be merged in 1.2), as I don't want to miss the merge window for 1.1 to give users a working process. To the best of my knowledge it's stable.

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
```diff
@@ -117,19 +129,72 @@ static void nfc_scene_mf_classic_dict_attack_update_view(NfcApp* instance) {
...
 }
 
 static void nfc_scene_mf_classic_dict_attack_prepare_view(NfcApp* instance) {
     uint32_t state =
         scene_manager_get_scene_state(instance->scene_manager, NfcSceneMfClassicDictAttack);
+    if(state == DictAttackStateCUIDDictInProgress) {
+        do {
+            size_t cuid_len = 0;
+            const uint8_t* cuid = nfc_device_get_uid(instance->nfc_device, &cuid_len);
+            FuriString* cuid_dict_path = furi_string_alloc_printf(
```
RebornedBrain: Looks like `furi_string_free(cuid_dict_path) ` is missing somewhere below

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
```diff
@@ -117,19 +129,72 @@ static void nfc_scene_mf_classic_dict_attack_update_view(NfcApp* instance) {
...
+                "%s/mf_classic_dict_%08lx.nfc",
```
RebornedBrain: Sorry for dumb question, but what is the purpose of this `mf_classic_dict_%08lx.nfc` file? Is it some kind of card specific dictionary, where only keys related to the card are stored?

lib/nfc/protocols/mf_classic/mf_classic_poller.c
RebornedBrain: @noproto ok, no problem

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
noproto: Good catch. Thank you, fixed!

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
noproto: Right! Static encrypted nested attacks produce anywhere between 10k-50k key candidates per key. We need to ensure those keys are only used for that **specific** tag or we will get slower dictionary attacks for all future tags.

Also we must not use the accelerated dictionary attack since the properties it uses to search for keys faster are satisfied by all of the static encrypted key candidates.

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
```diff
@@ -117,19 +129,75 @@ static void nfc_scene_mf_classic_dict_attack_update_view(NfcApp* instance) {
...
+                EXT_PATH("nfc/assets"),
+                (uint32_t)bit_lib_bytes_to_num_be(cuid + (cuid_len - 4), 4));
+
+            if(!keys_dict_check_presence(furi_string_get_cstr(cuid_dict_path))) {
+                state = DictAttackStateUserDictInProgress;
```
RebornedBrain, line 152: `furi_string_free(cuid_dict_path);` also required here, because if flow goes here then break, it will not reach the free added below.

lib/nfc/protocols/mf_classic/mf_classic_poller.c
```diff
@@ -6,8 +6,25 @@
...
 #define MF_CLASSIC_MAX_BUFF_SIZE (64)
 
+// Ordered by frequency, labeled chronologically
+const MfClassicBackdoorKeyPair mf_classic_backdoor_keys[] = {
+    {{{0xa3, 0x96, 0xef, 0xa4, 0xe2, 0x4f}}, MfClassicBackdoorAuth3}, // Fudan (static encrypted)
+    {{{0xa3, 0x16, 0x67, 0xa8, 0xce, 0xc1}}, MfClassicBackdoorAuth1}, // Fudan, Infineon, NXP
+    {{{0x51, 0x8b, 0x33, 0x54, 0xe7, 0x60}}, MfClassicBackdoorAuth2}, // Fudan
+};
+const size_t mf_classic_backdoor_keys_count =
+    sizeof(mf_classic_backdoor_keys) / sizeof(mf_classic_backdoor_keys[0]);
```
RebornedBrain: You can use COUNT_OF macro here, like this: `const size_t mf_classic_backdoor_keys_count = COUNT_OF(mf_classic_backdoor_keys);`

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
```diff
@@ -29,7 +34,9 @@ NfcCommand nfc_dict_attack_worker_callback(NfcGenericEvent event, void* context)
     } else if(mfc_event->type == MfClassicPollerEventTypeRequestMode) {
         const MfClassicData* mfc_data =
             nfc_device_get_data(instance->nfc_device, NfcProtocolMfClassic);
-        mfc_event->data->poller_mode.mode = MfClassicPollerModeDictAttack;
+        mfc_event->data->poller_mode.mode = (instance->nfc_dict_context.enhanced_dict) ?
```
RebornedBrain, line 36: I see one place where this `enhanced_dict` is set to `true` on line 175, which means that we will start from `MfClassicPollerModeDictAttackEnhanced`. And how `MfClassicPollerModeDictAttackStandard` will be performed now? I'm asking because the only place where `enhanced_dict` is set to `false` is on line 385, in `_on_exit` callback.

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
noproto: Relocated the logic to keep a single call to `furi_string_free` and minimize flash use.

noproto: https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/3976f128dc86be161cb0bb62900e1259aa3a34e1

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
noproto: This flag is to indicate whether nested attacks should be attempted. Nested attacks are disabled during the CUID dictionary attack to force the default behavior (`MfClassicPollerModeDictAttackStandard`), as all of the keys in the CUID dictionary match the properties that the enhanced attack searches for. After `enhanced_dict` is set to `true`, it should no longer be set to `false` except upon exit as the CUID dictionary is tested first - the only time it is needed. The variable is necessary to inform the poller how it should behave during initial phase of the dictionary attack. Standard attacks are still attempted in `MfClassicPollerModeDictAttackEnhanced`, up until the point the first key is found.

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
```diff
@@ -222,7 +304,27 @@ bool nfc_scene_mf_classic_dict_attack_on_event(void* context, SceneManagerEvent
         } else if(event.event == NfcCustomEventDictAttackSkip) {
             const MfClassicData* mfc_data = nfc_poller_get_data(instance->poller);
             nfc_device_set_data(instance->nfc_device, NfcProtocolMfClassic, mfc_data);
-            if(state == DictAttackStateUserDictInProgress) {
+            bool ran_nested_dict = instance->nfc_dict_context.nested_phase !=
```
RebornedBrain, line 306: In my opinion, this one is not needed here because we have the same on line 264 which we can move out from if to line 262.

RebornedBrain: And by the way, maybe it would be better to change type of nested_phase  to appropriate enum instead of `uint8_t`? I mean this one https://github.com/flipperdevices/flipperzero-firmware/pull/3822/files#r1795343085

applications/main/nfc/nfc_app_i.h
```diff
@@ -93,6 +97,12 @@ typedef struct {
     bool is_key_attack;
     uint8_t key_attack_current_sector;
     bool is_card_present;
+    uint8_t nested_phase;
+    uint8_t prng_type;
+    uint8_t backdoor;
```
RebornedBrain: These three variables hold values from enum types, so it would be nice to change their types to appropriate enums

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
noproto: > In my opinion, this one is not needed here because we have the same on line 264 which we can move out from if to line 262.

This exists to reduce computation so we don't do the comparison on every custom event.

> And by the way, maybe it would be better to change type of nested_phase to appropriate enum instead of `uint8_t`? I mean this one https://github.com/flipperdevices/flipperzero-firmware/pull/3822/files#r1795343085

Will follow up on your other comment: https://github.com/flipperdevices/flipperzero-firmware/pull/3822#discussion_r1795343085

applications/main/nfc/nfc_app_i.h
noproto: @RebornedBrain Fixed: https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/a7c0819034fdbe591d64d16b95ba1f88cff660eb

lib/nfc/protocols/mf_classic/mf_classic_poller_i.c
```diff
@@ -89,29 +96,34 @@ MfClassicError mf_classic_poller_get_nt(
...
+MfClassicError mf_classic_poller_auth_common(
     MfClassicPoller* instance,
     uint8_t block_num,
     MfClassicKey* key,
     MfClassicKeyType key_type,
     MfClassicAuthContext* data,
-    bool is_nested) {
+    bool is_nested,
+    bool backdoor_auth,
+    bool early_ret) {
```
RebornedBrain, line 126: From my point of view it's not very good to put bool vars into function params. This means that this function doesn't respond to single responsibility principle. Also reading and understanding of such functions requires more effort, in my opinion.
Despite the diff, which says that this problem was here even before (`is_nested`), we must keep our api as much clean as possible without such things, that's why I spent some time and tried to remove those booleans, it required some efforts, but looks like that it works fine. 
Here is the patch with things which I've done: [nestednonces_api_patch.patch](https://github.com/user-attachments/files/17346670/nestednonces_api_patch.patch)
Could you please apply it and check whether everything still works as expected?
This is the most important blocker for merging now.

RebornedBrain: @gornekich it would be great if you will take a look, after PR will be updated and approve such changes

lib/nfc/protocols/mf_classic/mf_classic_poller.c
```diff
@@ -505,6 +557,128 @@ NfcCommand mf_classic_poller_handler_request_read_sector_blocks(MfClassicPoller*
     return command;
 }
 
+NfcCommand mf_classic_poller_handler_analyze_backdoor(MfClassicPoller* instance) {
+    NfcCommand command = NfcCommandReset;
```
RebornedBrain, line 561: Looks like this logic can be simplified and moved to mf_classic api or even to a separate module, smth like mf_classic_backdoor.c

In that module all logic of testing backdoor can be done, and as a public function there can be something like: `mf_classic_test_backdoor(MfClassicBackdoor backdoor_type);` or even better `MfClassicBackdoor backdoor = mf_classic_find_backdoor();` which will return us value from enum and do all the magic under the hood

This can be placed there: https://github.com/flipperdevices/flipperzero-firmware/pull/3822/files#r1797349885

lib/nfc/protocols/mf_classic/mf_classic_poller.c
```diff
@@ -6,8 +6,24 @@
...
+// TODO: Buffer writes for Hardnested, set state to Log when finished and sum property matches
+// TODO: Store target key in CUID dictionary
+// TODO: Fix rare nested_target_key 64 bug
+// TODO: Dead code for malloc returning NULL?
+// TODO: Auth1 static encrypted exists (rare)
+
 #define MF_CLASSIC_MAX_BUFF_SIZE (64)
 
+// Ordered by frequency, labeled chronologically
+const MfClassicBackdoorKeyPair mf_classic_backdoor_keys[] = {
```
RebornedBrain, line 18: This can be moved to a separate module with backdoor logic

lib/nfc/protocols/mf_classic/mf_classic_poller_i.c
noproto: Skimming the patch, dict_attack.h duplicates the include on line 6 and backdoor detection uses `FURI_LOG_E` (indicating an error) instead of the currently defined log levels. My understanding is that API version 85 in api_symbols.csv is a development artifact. I'm closely reviewing how it handles nested authentication, then I'll test it on each type of card.

noproto: The changes in the patch to nfc_scene_mf_classic_dict_attack.c increase total computation performed. https://github.com/flipperdevices/flipperzero-firmware/pull/3822#discussion_r1796151862

mf_classic_poller_send_frame_callback in nfc_test.c needs to have the call updated to match the new prototype of `mf_classic_poller_auth`.

noproto: @RebornedBrain As mentioned in my earlier comment, I was closely reviewing how the changes in the patch affected nested authentication. The changes affect a subtle but critical part of the code.

The device must perform *almost exactly* the same operations in order for the calibration code to have the same runtime as the nested nonce collection. This is a very sensitive operation: the LFSR on the tag (assuming its not static or static encrypted) shifts every ~9.4395 µs. The function call overhead of calling two functions versus one (`mf_classic_poller_auth_nested`+`mf_classic_poller_get_nt_and_update_auth_context` versus just `mf_classic_poller_get_nt_and_update_auth_context`) introduces a timing discrepancy that cannot be detected by the device. During nonce collection, you'll collect nonces but they'll either be offset by an unknown amount (I am not able to get any clocksource on the device to calculate the difference with sufficient precision) or unsolvable.

To verify this issue exists, I used a non-static weak card which I was able to successfully solve with the original PR. I flashed the changes in the patch, it collected multiple invalid nonce pairs during collection (as far as I could tell this affected a majority of collected nonce pairs). Only by flashing back the original changes am I able to find the key immediately.

Further, because the nonce is not located in the correct position, I noted only one third of the possible nonce pairs were collected across the retry attempts (resulting in longer collection times and fewer logged pairs).

I do not believe we should merge this patch if it will disrupt collection for one category of cards: cards with regular weak PRNGs.

noproto:
My understanding of where we're at:

1. *Improved dictionary attack/nested nonce collection*. This PR #3822 which is functionally working. The API implementation is being discussed in [this review thread](https://github.com/flipperdevices/flipperzero-firmware/pull/3822#discussion_r1797344520). I will open a new PR for additional code-related improvements and performance optimizations. 
2. *MFKey included for key recovery*. #3943 was opened separately but intended to be merged alongside #3822 so they support each other.
3. MFKey flow: I'd write this if I could. I don't know how to unload the currently running app and launch a new app from the active application (NFC app). I'm thinking this should be a menu item of "Recover keys" or "Crack keys" that is listed after partially reading a MIFARE Classic card, and launches the MFKey application included in #3943.

lib/nfc/protocols/mf_classic/mf_classic_poller_i.c
RebornedBrain: > Skimming the patch, dict_attack.h duplicates the include on line 6 and backdoor detection uses `FURI_LOG_E` (indicating an error) instead of the currently defined log levels. My understanding is that API version 85 in api_symbols.csv is a development artifact. I'm closely reviewing how it handles nested authentication, then I'll test it on each type of card.

Oh, sorry for that, I did patch before pulling your latest changes, that's why there were duplications, and Log level I increased for tests only, and I agree that it shouldn't be changed

I will investigate your feedback on patch more precisely today.

skotopes:
@noproto sorry for delay, @RebornedBrain and QA being testing this PR against our card library. They found bunch of cards that cannot be read with this PR. They'll bring details later.

noproto:
That is unfortunate. I tested it on 110 MIFARE Classic cards here. Very interested to find out the issue so I can remedy it.

noproto:
@RebornedBrain Could I have an update on this PR? Which cards no longer read specifically?

For example - if it is MIFARE Plus, is it SL1 or SL2? I do not possess a MIFARE Plus card in my collection (all of my cards scan), but I can find members of the community who have certain cards for further testing and to assist us with locating any issue.

Please let me know if there's anything I can do. I've paused development while I wait on your feedback (both on the cards which no longer read and on this thread: https://github.com/flipperdevices/flipperzero-firmware/pull/3822#pullrequestreview-2363439806 ).

skotopes:
@noproto hold on, I'll bring report and test cards here in next couple hours

skotopes:
So, @RebornedBrain @doomwastaken did some tests. Here is what they found:

# Test 1: no plugins, stuck at particular key

Tests were performed without plugins (remove folder Plugins at "SD Card/apps_data/nfc")
This card stuck at one particular key. In logs I see that process still goes, but
key remains the same, but before keys were changing. See details on the screenshot.

[test_1.zip](https://github.com/user-attachments/files/17400706/test_1.zip)

# Test 2:  MFC 4k, crash

This MFC 4k (not magic) card where first sector was protected with key A "DEADBEAFFFFF"
which is not present in any dictionaries. Result - crash "Wrong sector num", repeats
every time. This card has backdoor v2.

[test_2.zip](https://github.com/user-attachments/files/17400720/test_2.zip)

# Test 3: MFC 1k, can't find key 

This MFC 1k (magic) card where first sector was also protected with key A "DEADBEAFFFFF".
Result Flipper cannot find this key. This card has no any backdoor.

[test_3.zip](https://github.com/user-attachments/files/17400730/test_3.zip)

# Test 4: Test with plugins

Flipper reads this card totally when there is no plugins (it passes nested attack and etc).
When plugins are present Flipper unable to read it and in logs I see some auth errors.

[test_4.zip](https://github.com/user-attachments/files/17400745/test_4.zip)

Test 5: On this card Flipper stuck

On this card Flipper stuck

[test_5.zip](https://github.com/user-attachments/files/17400758/test_5.zip)

@noproto let us know if you need any additional help/info/etc...

noproto:
Awesome, on it and I'll follow up shortly with fixes and feedback. Thank you for testing the PR!

noproto:
All issues resolved that I could reproduce, please re-test this PR @skotopes @RebornedBrain @doomwastaken:

> So, @RebornedBrain @doomwastaken did some tests. Here is what they found:
> # Test 1: no plugins, stuck at particular key
> 
> Tests were performed without plugins (remove folder Plugins at "SD Card/apps_data/nfc") This card stuck at one particular key. In logs I see that process still goes, but key remains the same, but before keys were changing. See details on the screenshot.
> 
> [test_1.zip](https://github.com/user-attachments/files/17400706/test_1.zip)
> # Test 2: MFC 4k, crash
> 
> This MFC 4k (not magic) card where first sector was protected with key A "DEADBEAFFFFF" which is not present in any dictionaries. Result - crash "Wrong sector num", repeats every time. This card has backdoor v2.
> 
> [test_2.zip](https://github.com/user-attachments/files/17400720/test_2.zip)

## Steps taken
1. Noted power of 2 in both sector and keys found. Possible issue with datatype. Decided to not remove plugins for initial test.
2. Using NFC Magic, cloned Plantan_white.nfc to a Gen4 UMC magic card
3. Read card with NFC app. Crashes on sectors read 32/40, keys found 43/80, 33/40 on progress bar, "[CRASH][NfcWorker] furi_check failed" after "Found key candidate"
4. Dumped call stack:
```
__furi_crash_implementation@0x08012280 (/home/ubuntu/Flipper/nestednonces/flipperzero-firmware/furi/core/check.c:170)
mf_classic_get_device_name@0x08036452 (/home/ubuntu/Flipper/nestednonces/flipperzero-firmware/lib/nfc/protocols/mf_classic/mf_classic.c:340)
mf_classic_get_first_block_num_of_sector@0x08036d6a (/home/ubuntu/Flipper/nestednonces/flipperzero-firmware/lib/nfc/protocols/mf_classic/mf_classic.c:565)
mf_classic_poller_handler_key_reuse_auth_key_a@0x080420fe (/home/ubuntu/Flipper/nestednonces/flipperzero-firmware/lib/nfc/protocols/mf_classic/mf_classic_poller.c:916)
(..)
```
5. Identified issue: reuse_key_sector assigned to 43, when tag only has 40 sectors (0-39). This is because of a bad assumption (sector number = block number / 4, is not true for MFC 4K)
6. Fixed in https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/4be9e793421828516579b152bac716b961dd55f7
7. Re-ran dictionary attack, reproduced test 2 crash at end of nonce collection: "[CRASH][NfcWorker] Wrong sector num"
8. Dumped call stack:
```
__furi_crash_implementation@0x08012280 (/home/ubuntu/Flipper/nestednonces/flipperzero-firmware/furi/core/check.c:170)
mf_classic_get_sector_trailer_num_by_sector@0x0803666c (/home/ubuntu/Flipper/nestednonces/flipperzero-firmware/lib/nfc/protocols/mf_classic/mf_classic.c:403)
mf_classic_get_sector_trailer_num_by_sector@0x0803666c (/home/ubuntu/Flipper/nestednonces/flipperzero-firmware/lib/nfc/protocols/mf_classic/mf_classic.c:395)
mf_classic_poller_handler_nested_collect_nt_enc@0x080427fa (/home/ubuntu/Flipper/nestednonces/flipperzero-firmware/lib/nfc/protocols/mf_classic/mf_classic_poller.c:1362)
(..)
```
9. Identified issue: off by one in nested_target_key, doesn't crash on MFC 1K since no effect besides reduced performance (same issue as "TODO: Fix rare nested_target_key 64 bug")
10. Fixed in https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/4be9e793421828516579b152bac716b961dd55f7
11. Test: Read tag, recovered 29 keys with MFKey (78/80 keys), read tag, recovered 2 keys with MFKey, 80/80 keys 40/40 sectors read. Working.
12. Began re-testing original issues. Removed plugins folder, unable to reproduce. Removed user dictionary, unable to reproduce.

## Conclusion
Two issues: faulty state machine logic, target sector on 4K cards. Fixed in https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/4be9e793421828516579b152bac716b961dd55f7

> # Test 3: MFC 1k, can't find key
> 
> This MFC 1k (magic) card where first sector was also protected with key A "DEADBEAFFFFF". Result Flipper cannot find this key. This card has no any backdoor.
> 
> [test_3.zip](https://github.com/user-attachments/files/17400730/test_3.zip)

## Steps taken
1. Verified Prox.nfc has sector 1 key A DEADBEAFFFFF
2. Using NFC Magic, cloned Prox.nfc to a Gen1a magic card
3. Read card with NFC app. 31/32 keys. 1 nonce collected in .nested.log
4. Cracked nonce in MFKey.
5. Returned to NFC app and read card. 32/32, 16/16 sectors read
6. Flashed second Flipper device with `nestednonces` fork
7. Same result

## Conclusion
Possible cache issue? Verify no keys are cached for this tag.
If issue persists, debug logs would be useful.

> # Test 4: Test with plugins
> 
> Flipper reads this card totally when there is no plugins (it passes nested attack and etc). When plugins are present Flipper unable to read it and in logs I see some auth errors.
> 
> [test_4.zip](https://github.com/user-attachments/files/17400745/test_4.zip)

## Steps taken
1. Ensured NFC plugins folder present on Flipper device
2. Using NFC Magic, cloned Small_troyka.nfc to a Gen4 UMC magic card
3. Read card with NFC app. 32/32 keys.
4. Flashed second Flipper device with `nestednonces` fork
5. Reproduced auth errors. Narrowed issue to interaction between `nestednonces` and troika_parser.fal NFC plugin.
6. Identified issue: inconsistent assignment of known key and known key type/sector. This is because of a bad assumption (key found in dictionary attack was assumed to be first key)
7. Fixed in https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/897817a829583f705d0ece5734c2a3fdd235304a and https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/db26c85b231cee324a6e3520e5eb199b52991190
8. Re-ran dictionary attack on all card types, works. No longer able to reproduce original issue.

## Conclusion
Inconsistent assignment of known key and known key type/sector led to repeated failed authentication attempts. Keys were provided by specific NFC plugins instead of the dictionary attack. Fixed in https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/897817a829583f705d0ece5734c2a3fdd235304a and https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/db26c85b231cee324a6e3520e5eb199b52991190

> # Test 5: On this card Flipper stuck
> 
> On this card Flipper stuck
> 
> [test_5.zip](https://github.com/user-attachments/files/17400758/test_5.zip)

## Steps taken
1. Ensured NFC plugins folder present on Flipper device
2. Using NFC Magic, cloned Disappeared_corridor.nfc to a Gen4 UMC magic card
3. Read card with NFC app. 70/80 keys.
4. Flashed second Flipper device with `nestednonces` fork
5. Same result. Troyka card recognized
6. Removed plugins folder.
7. Read card with NFC app. 70/80 keys.
8. Added a user dictionary file to the device with 1 irrelevant key
9. Read card with NFC app. 70/80 keys. Unable to reproduce original issue.

## Conclusion
Likely related to the other (now fixed) issues. Please re-test.
If issue persists, debug logs would be useful.

noproto:
Since https://github.com/flipperdevices/flipperzero-firmware/pull/3961#issuecomment-2423383808 will delay 1.1, is it still possible to squeeze this PR in too if the issues identified in QA are resolved by https://github.com/flipperdevices/flipperzero-firmware/pull/3822#issuecomment-2421454416 ?

These changes would allow me to share an early, significantly easier process with the users as well as limit the scope of each PR versus rolling up more changes into this single PR making it unwieldy to do QA. Additionally, it would be useful to collect feedback from users so we can identify any rare or uncommon issues (thanks to the exceptional QA already done, every issue that was reported to me - even prior to the review - has been resolved).

RE: how users could use the nonces collected by this PR, PR 243 to good-faps would make a complete process available: https://github.com/flipperdevices/flipperzero-good-faps/pull/243

Wanted to bring this up for your consideration. I understand if it's not possible. For context, [this is what is required for reading cards today](https://gist.github.com/noproto/63f5dea3f77cae4393a4aa90fc8ef427#reading-the-card). It becomes a two step process with this PR (3822) plus PR 243 to good-faps.

RebornedBrain:
@noproto, Hi, I've retested those cases and here are the results:

**Test 1: no plugins, stuck at particular key** - Fixed, now works fine. ✅ 
**Test 2: MFC 4k, crash** - Fixed, now Flipper reads card and collect nonces. ✅ 
**Test 3: MFC 1k, can't find key** - Fixed, ✅  but new issue appeared, more info below 
**Test 4** - Fixed, in both cases (no plugins/with plugins) Flipper reads card ✅ 
**Test 5: On this card Flipper stuck** - Fixed. ✅ 
Nicely done 👍 

Now about new issue which I've noticed. Actually, maybe it was even previously, but we didn't see it. When I tested test 2 and 3, sometimes Flipper tries to perform attack again, even if I already have cracked nonces with the help of MFKey and key `DEADBEAFFFFF` is already present in user_dict.
I tested a lot on two Flippers which I have, in order to narrow roots of possible problem, and came to a conclusion that it mostly depends on key count present in `mf_classic_dict_user.nfc`. I attach my dictionary with 89 keys in it, with this dictionary one Flipper fails to read one sector more often (1 to 3 fails in 10 attempts), and another one fails rearly (1 fail in 40 attempts), but still. On the other hand when I deleted user dictionary completely, and then added only one key `DEADBEAFFFFF` then issue dissapeared, which proofs that dictionary size can cause some artefacts. 

Here are steps to reproduce:
1. Import `mf_classic_dict_user.nfc` to nfc->assets
2. Check that key `DEADBEAFFFFF` is already present in users' keys
3. Try to read card from test 2 or test 3.
After this step some steps may vary because I don't have the exact sequence.
4. If card read fine, press retry (sometimes 1 or 2 retries might be enough, sometimes even 20 will give no effect).
5. If card read fine, exit NFC app, then enter again and retry step 4.
6. If card read fine, reboot Flipper completely and repeat steps from 4.
Steps 4-6 can be combined in different ways before issue appears.
7. If card read fails to read 1 sector, Flipper collect nonces to `.nested.logs`, then you can delete this file via qFlipper and press retry, this can cause issue to repeat again, but not every time.

**Expected result:** Flipper reads card totally each time without any attempts of collecting nonces, because it already has known key in user dictionary.

**Result:** Flipper sometimes doesn't apply known key during read process, instead it collects nonces.

[dictionary_with_issue.zip](https://github.com/user-attachments/files/17494156/dictionary_with_issue.zip)

noproto:
Hi @RebornedBrain, good catch. Fixed in https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/6dbb46a81d1de7aa0bc55775461ef747a3ab6964 I believe. Try again.

skotopes:
PVS Report:
image

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
skotopes: Can we change TODO to something else, otherwise it will trigger our TODO checker?

applications/main/nfc/scenes/nfc_scene_mf_classic_dict_attack.c
skotopes: @RebornedBrain don't forget to resolve all conversations

skotopes (changes requested):

skotopes:
I guess we are ready to merge this PR, couple small things left to cleanup:
- TODOs(add `FL-3926` after TODO, so CI/CD will stop complaining): https://github.com/flipperdevices/flipperzero-firmware/actions/runs/11526331987/job/32197423477?pr=3822
- PVS Checks

noproto:
@skotopes :

Remaining TODOs marked with FL-3926.

I'll open the refactoring PR when 3822 is merged, and I'll keep improving it as needed (clearing all TODOs, unit tests for parity, backdoor detection moved outside the poller).

If its helpful, I'll add [FL-3926] to the title of the next PR. RE: PVS:

* mf_classic_poller.c:1292: Redundant assignment removed in https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/eb1aabbec4819b02c430ad62013723063fe941ca
* mf_classic_poller.c:1293: Redundant assignment removed in https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/eb1aabbec4819b02c430ad62013723063fe941ca
* mf_classic_poller.c:1707: Format specifier corrected in https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/c24007755a192ce4cd925e5c65a83a5b00813244
* mf_classic_poller.c:1938: PVS optimization included in https://github.com/flipperdevices/flipperzero-firmware/pull/3822/commits/907019c136d57ef727e75291869a94e746db3af1

All good warnings by PVS this time.

[FL-3926]: https://flipperzero.atlassian.net/browse/FL-3926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

skotopes (approved):

skotopes:
@noproto awesome work )

psifertex:
Thanks everyone involved for the work landing this! Looking forward to trying it out.

mishamyte:
Thanks for your work @noproto.

And an opened question. Do you think, guys, nonces recovery could be implemented on a mobile app side conceptually?
Just as we have for MFKey32, to speed up recovery

noproto:
> Thanks for your work @noproto.
> 
> And an opened question. Do you think, guys, nonces recovery could be implemented on a mobile app side conceptually? Just as we have for MFKey32, to speed up recovery

I appreciate your reply @mishamyte. As far as offloading goes, I am only interested in doing this for attacks which are not possible to run on the Zero (Hardnested).

Since the Flipper Zero is a first generation device, the on-device attacks today are the worst they will ever be. Optimization of the key recovery process on the device has brought multiple recovery methods from taking months to minutes. I will continue to optimize and develop code for the process defined in this PR, but as new hardware revisions are imminent I expect that even without changes it will soon run in seconds rather than minutes (which would immediately deprecate any offloading effort done except for Hardnested).

mishamyte:
> > Thanks for your work @noproto.
> > 
> > And an opened question. Do you think, guys, nonces recovery could be implemented on a mobile app side conceptually? Just as we have for MFKey32, to speed up recovery
> 
> I appreciate your reply @mishamyte. As far as offloading goes, I am only interested in doing this for attacks which are not possible to run on the Zero (Hardnested).
> 
> Since the Flipper Zero is a first generation device, the on-device attacks today are the worst they will ever be. Optimization of the key recovery process on the device has brought multiple recovery methods from taking months to minutes. I will continue to optimize and develop code for the process defined in this PR, but as new hardware revisions are imminent I expect that even without changes it will soon run in seconds rather than minutes (which would immediately deprecate any offloading effort done except for Hardnested).

Thank you!
Personally I don't think my proposal should replace existing functionality.
IMO for me it's just an alternative.
Calculating on Flipper is good for small amount of keys, but when you need to calculate full diversified tag, it takes an eternity (that's feedback from guys, who tested).

So personally I would prefer to have both options. As we had it previously - with Python script + old fap

RaYRAy773 (commented):
Cool

davidsobey1:
Awesome work @noproto , do you think this might now work on legacy oyster card, which I think are based on MIFARE Classic?

noproto:
> Awesome work @noproto , do you think this might now work on legacy oyster card, which I think are based on MIFARE Classic?

I don't see why not. Update to the latest firmware and try it out, let me know if you experience any issues. Here is a guide to using the features: https://flipper.wiki/mifareclassic/

PR #3823, opened 2024-08-05
[FL-3880] Fix cumulative error in infrared signals

gsurkov:
# What's new

- Transmitted infrared signals no longer exhibit a cumulative error due to rounding.
- Fix the wrong length of the leading carrier pulse.

# Verification 

1. Compile and flash this branch
2. Attach a logic analyser to the PA7 pin
3. Begin the capture and run `./fbt launch APPSRC=infrared_test`
4. Look in the debug log output, note the predicted signal duration
5. Measure the captured signal duration from the first rising edge to the last falling edge
6. The signal duration should be within predicted limits.

# Checklist (For Reviewer)


CookiePLMonster:
If you want to preserve the remainder of the division, isn't https://en.cppreference.com/w/c/numeric/math/div or `div` + `rem` a better way to go here? Might potentially also let you keep calculating on integers, without the need to cast to floats.

gsurkov:
The code was already dealing in floats (`cycle_duration` is a float), and the remainder has to be a float because it is always less than 1. I could probably calculate it in terms of microseconds instead of carrier periods, but I can't see any benefits of doing so without profiling.

CookiePLMonster:
> The code was already dealing in floats (`cycle_duration` is a float), and the remainder has to be a float because it is always less than 1. I could probably calculate it in terms of microseconds instead of carrier periods, but I can't see any benefits of doing so without profiling.

Ah, I see. In this case you can likely use `modf` to get the integral part and the fractional part of the division, and store the fractional part as a remainder. Should be similar to your current solution, but without rounding and the awkward subtraction used to calculate the remainder.

gsurkov:
Proper rounding (up and down) is preferable in this case because it spreads extra pulses more evenly across time. Suppose we get an intended duration of 3.9 pulses, which will turn into 3 this time and 1 extra pulse next time if we use `modf`, and 4 and 0 with `round`, which will be more representative of the intended signal. Same goes for small fractional parts, too.

skotopes (approved):

PR #3830, opened 2024-08-08
[FL-3889] 5V on GPIO control for ext. modules

### LINKED ISSUE #3537: API request: Automatically detecting the voltage at which a external module operates.

Is it possible to electronically implement (adding a resistor to some pin) or using the "Expansion Module Protocol," for F0 to activate 5V automatically? This would involve sending a signal to utilize `furi_hal_power_enable_otg`.

With this capability, each external module could communicate to F0 the voltage at which it operates, thereby avoiding some voltage drops that occur at 3.3V, it would eliminate the need to route through `GPIO` / `5V on GPIO  ON`

### Anything else?

_No response_

gsurkov:
# What's new

- Add support for OTG (5V) on GPIO:
    - control and get status via RPC.
    - control via Expansion protocol.
- Update the `expansion_test` app to test the OTG feature.
- Fixes #3537

Requires the [proto files](https://github.com/flipperdevices/flipperzero-protobuf/pull/51)  to be merged first.

# Verification 

1. Flash the firmware and disconnect the debug interface, in case it was connected (it will interfere with the next steps).
2. Read the instructions for the [expansion_test](https://github.com/flipperdevices/flipperzero-firmware/blob/dev/applications/debug/expansion_test/expansion_test.c) application and set up the test device accordingly.
3. Connect an LED (with an appropriate series resistor) to pins 1 (5V) and 8 (GND).
4. Build the `expansion_test` app and upload it to Flipper by running `./fbt launch APPSRC=expansion_test`. The app should run normally (see what it means in the instructions).
5. Disconnect the USB cable from Flipper. The LED should turn off.
6. Navigate to `Apps -> Debug -> Expansion Module Test` and run the app again. The LED should blink 2 times with ON and OFF times = 1 second.
7. Ensure that there are no regressions in the operation of prior expansion modules (e.g. VGM).

# Checklist (For Reviewer)


EstebanFuentealba:
It works perfectly for me, thnks @gsurkov . One detail that took me some time to figure out—and might be helpful for others—is that it can't be used within an RPC session, for example:

```c
/* expansion_enable_otg() OK */
if(!expansion_start_rpc()) continue;
/* expansion_enable_otg() won't work here */
if(!expansion_stop_rpc()) continue;
/* expansion_enable_otg() OK */
```


https://github.com/user-attachments/assets/e7a2e1b7-86e8-42c8-8c18-1ba1d56497bf

gsurkov:
@EstebanFuentealba this behaviour is not intentional, but an oversight on my part.
If using the `ExpansionFrameControlCommandEnableOtg` command is desirable in RPC mode, I can add it. Otherwise, I will do as you suggested: explicitly state that in RPC mode, the RPC messages take over as the control method for OTG.

EstebanFuentealba:
The requirement was to implement support for modules that require 5V to function, eliminating the need to manually enable 5V through Menu/GPIO/Enable 5V. The current implementation fulfills this requirement perfectly. However, it is important to document that this solution should not be used within an RPC session.

skotopes (dismissed):

skotopes (approved):

PR #3839, opened 2024-08-13
[FL-3884] Proper integer parsing

portasynthinca3:
# What's new
- Adds the following string to integer conversion API functions that verify input validity including overflow checks (header `lib/toolbox/strint.h`):
  ```c
  StrintParseError strint_to_uint64(const char* str, char** end, uint64_t* out, uint8_t base);
  StrintParseError strint_to_int64(const char* str, char** end, int64_t* out, uint8_t base);
  StrintParseError strint_to_uint32(const char* str, char** end, uint32_t* out, uint8_t base);
  StrintParseError strint_to_int32(const char* str, char** end, int32_t* out, uint8_t base);
  StrintParseError strint_to_uint16(const char* str, char** end, uint16_t* out, uint8_t base);
  StrintParseError strint_to_int16(const char* str, char** end, int16_t* out, uint8_t base);
  ```
- Adds unit tests for these new functions
- Changes existing code to use these new functions instead of `sscanf` and `strtol` and friends wherever warranted and possible

# Verification 
Run unit tests

# Remarks
  - "strint" is not a typo, it's short for "string & integer"
  - not really sure if the API version in `api_symbols.csv` is correct

# Checklist (For Reviewer)


hedger (changes requested):
Please sync all targets' API versions.
./fbt TARGET_HW=18

portasynthinca3:
Not sure how to fix the PVS-Studio diagnostic. I see two possible causes:
  - PVS-Studio thinks that `strint_to_int32` might access the provided `out` outside the first element. This just does not happen.
  - PVS-Studio thinks that `int` and `int32_t` have different sizes. This is not the case with our platform.

lib/toolbox/strint.c
```diff
@@ -0,0 +1,121 @@
...
+StrintParseError strint_to_uint64_internal(
+    const char* str,
+    char** end,
+    uint64_t* abs_out,
+    bool* sign_out,
+    uint8_t base,
+    uint64_t max_abs_negative,
+    uint64_t max_positive) {
+    // skip whitespace
+    while(((*str >= 9) && (*str <= 13)) || *str == 32) {
```
hedger: Better use char literals:
```c
    while(((*str >= '\x9') && (*str <= '\xd')) || *str == '\x20') {
```

lib/toolbox/strint.c
```diff
@@ -0,0 +1,121 @@
...
+ * @param [in] max_abs_negative Largest permissible absolute part of result if
+ * the sign is negative
+ * @param [in] max_positive Largest permissible absolute part of result if the
+ * sign is positive
+ */
+StrintParseError strint_to_uint64_internal(
+    const char* str,
+    char** end,
+    uint64_t* abs_out,
+    bool* sign_out,
```
hedger: `val_is_negative` or smth similar would be more intuitive - don't have to look up the meaning in docs when calling

lib/toolbox/strint.c
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    if(abs_out) *abs_out = result;
+    if(sign_out) *sign_out = negative;
+    if(end) *end = (char*)str;
+    return StrintParseNoError;
+}
+
+#define STRINT_MONO(name, type, neg, pos)                                              \
```
hedger: Please add more context for names, like `ret_type`, `neg_abs_limit`, `pos_limit`, or write a comment

hedger (dismissed):

CookiePLMonster:
I'll submit a proper review tomorrow, but I'll comment it now - would it not make sense to also replace `atoi`. `atol` and `atoll`?

applications/main/infrared/infrared_cli.c
```diff
@@ -189,9 +189,11 @@ static bool infrared_cli_parse_raw(const char* str, InfraredSignal* signal) {
         }
 
         char timing_str[INFRARED_CLI_BUF_SIZE];
-        if(sscanf(str, "%9s", timing_str) != 1) {
+        if(strlen(str) > 9) {
```
CookiePLMonster:
```suggestion
        if(strlen(str) >= INFRARED_CLI_BUF_SIZE) {
```

Less magic constants.

applications/main/infrared/infrared_cli.c
```diff
@@ -189,9 +189,11 @@ static bool infrared_cli_parse_raw(const char* str, InfraredSignal* signal) {
...
             break;
         }
+        strncpy(timing_str, str, sizeof(timing_str));
+        timing_str[sizeof(timing_str) - 1] = 0;
```
CookiePLMonster: Any particular reason this can't be just a `strlcpy`? I doubt you need the "zero pad the remaining buffer" behaviour of `strncpy` here.

applications/main/infrared/infrared_cli.c
```diff
@@ -189,9 +189,11 @@ static bool infrared_cli_parse_raw(const char* str, InfraredSignal* signal) {
...
 
         str += strlen(timing_str);
         uint32_t timing = atoi(timing_str);
```
CookiePLMonster: This can probably become `strint_to_uint32` too?

lib/toolbox/strint.c
```diff
@@ -0,0 +1,121 @@
...
+        if(*str == '-') negative = true;
+        str++;
+    }
+    if(*str == '+' || *str == '-') return StrintParseSignError;
+    if(max_abs_negative == 0 && negative) return StrintParseSignError;
+
+    // infer base
+    // not assigning directly to `base' to permit prefixes with explicit bases
+    uint8_t inferred_base = 0;
+    if(memcmp(str, "0x", 2) == 0 || memcmp(str, "0X", 2) == 0) {
```
CookiePLMonster: I wouldn't trust those `memcmp` not to read past the string bounds if `str` at this time is `\0`. `strncmp` should be more suitable here.

CookiePLMonster (commented):
A few comments.

portasynthinca3:
Thank you, I'll address these comments.

lib/toolbox/strint.c
CookiePLMonster: Or even better, `strncasecmp`, turns those into a single case-insensitive check each.

CookiePLMonster:
> Thank you, I'll address these comments.

Looks nice now! Do you think it also makes sense to replace all uses of `atoi`?

portasynthinca3:
> Looks nice now! Do you think it also makes sense to replace all uses of `atoi`?

Yep, I think so too. I'll do that tomorrow.

applications/main/infrared/infrared_cli.c
```diff
@@ -177,8 +178,14 @@ static bool infrared_cli_parse_raw(const char* str, InfraredSignal* signal) {
     }
 
     uint32_t* timings = malloc(sizeof(uint32_t) * MAX_TIMINGS_AMOUNT);
-    uint32_t frequency = atoi(frequency_str);
-    float duty_cycle = (float)atoi(duty_cycle_str) / 100;
+    uint32_t frequency;
+    uint32_t duty_cycle_u32;
+    if(strint_to_uint32(frequency_str, NULL, &frequency, 10) != StrintParseNoError ||
+       strint_to_uint32(duty_cycle_str, NULL, &duty_cycle_u32, 10) != StrintParseNoError) {
+        free(timings);
```
CookiePLMonster: IMHO just move `malloc` under those checks, it'll save you having to free that allocation, and technically any code size reduction like that is nice (because there'll be no need to call `free`).

applications/main/infrared/infrared_cli.c
```diff
@@ -188,14 +195,11 @@ static bool infrared_cli_parse_raw(const char* str, InfraredSignal* signal) {
             ++str;
         }
 
-        char timing_str[INFRARED_CLI_BUF_SIZE];
-        if(sscanf(str, "%9s", timing_str) != 1) {
+        uint32_t timing;
+        if(strint_to_uint32(str, (char**)&str, &timing, 10) != StrintParseNoError) {
```
hedger: Better use a local discardable variable than cast const away

CookiePLMonster (commented):
One more nitpicky, but (technically) code-size-reducing comment.

applications/main/infrared/infrared_cli.c
CookiePLMonster: Seems fine in this context to be fair, it's just a limitation of C because you can't express "pointer to a pointer to char **or** const char as we don't care about the constness". `strtol` from the standard seems to face the same issue.

applications/main/infrared/infrared_cli.c
portasynthinca3: yeah performance is also a concern, I'll fix it

CookiePLMonster: Realistically you'll "never" reach this error condition, as in it's probably rare enough that for all intents and purposes it's dead code as far as a regular user is concerned - so it probably doesn't make any difference for performance either way

skotopes (dismissed):

applications/main/infrared/infrared_cli.c
skotopes: @CookiePLMonster here is rabbit hole for you: https://c-faq.com/ansi/constmismatch.html

skotopes (approved):

CookiePLMonster:
@skotopes Does it make sense to strip away `atoi`, `strtol` and friends from the firmware/API now that this is available?

skotopes:
@CookiePLMonster yes, but we are not going to do it because external libraries/code may depend on it.

CookiePLMonster:
Yeah that makes sense, if third parties use it you'll not save any flash space anyway, so might as well keep exposing it...

PR #3858, opened 2024-08-28
Exposed `view_dispatcher_get_event_loop`

CookiePLMonster:
# What's new

`view_dispatcher_get_event_loop` is available in the public View Dispatcher header, but it was marked as inaccessible for FAPs. This was most likely done because of the initial event loop implementation not supporting adding/removing additional entries whilst the event loop was running, but now it is supported so it should be OK to expose.

# Verification 

Compiles?

# Checklist (For Reviewer)


hedger (dismissed):

CookiePLMonster:
Additionally, is there any point to keeping a deprecated `view_dispatcher_enable_queue`? Maybe it should just be removed at this point to force people to upgrade?

skotopes:
We are currently thinking about how interface to event loop should look like. 

At this point we have couple options like: passing event loop to alloc_ex or leaving current API. It will become clear which one (or maybe both) will stay after we port GUI to JS.

CookiePLMonster:
> We are currently thinking about how interface to event loop should look like.
> 
> At this point we have couple options like: passing event loop to alloc_ex or leaving current API. It will become clear which one (or maybe both) will stay after we port GUI to JS.

Sounds good! If you want to defer this PR to until after this decision is made, I can either mark this PR as draft or close it.

skotopes (approved):

skotopes:
Long story short: we will have both APIs available: borrowing event loop from view dispatcher or injecting your own loop into view dispatcher

CookiePLMonster:
> Long story short: we will have both APIs available: borrowing event loop from view dispatcher or injecting your own loop into view dispatcher

After I tried using the event loop in my app, I think it's indeed a good idea. Just be mindful of the thread signal thing when "adopting" the event loop to the view dispatcher.

PR #3866, opened 2024-09-02
Replace all calls to strncpy with strlcpy, use strdup more, expose strlcat

CookiePLMonster:
# What's new

This PR improves the consistency of string copy calls, removing all uses of `strncpy` and `strcat`, at least from the code that isn't in submodules. `strlcpy` is preferable over `strncpy`, because it doesn't needlessly zero the remaining destination buffer, and it ensures null-termination.

Because `strlcat` is already used in mjs, I exposed it in the API. Please check if this is fine.

To make buffer sizes consistent, I modified the following constants to include the space for the null terminator. Please check if this is fine:
* `IBUTTON_KEY_NAME_SIZE`
* `INFRARED_MAX_BUTTON_NAME_LENGTH`
* `INFRARED_MAX_REMOTE_NAME_LENGTH`
* `MAX_TEXT_INPUT_LEN`

# Verification 

* Run a full suite of tests.

# Checklist (For Reviewer)


applications/debug/rpc_debug_app/rpc_debug_app.c
```diff
@@ -60,7 +60,10 @@ static void rpc_debug_app_rpc_command_callback(const RpcAppSystemEvent* event, v
         furi_assert(event->data.type == RpcAppSystemEventDataTypeBytes);
 
         rpc_debug_app_format_hex(
-            event->data.bytes.ptr, event->data.bytes.size, app->text_store, TEXT_STORE_SIZE);
+            event->data.bytes.ptr,
+            event->data.bytes.size,
+            app->text_store,
+            sizeof(app->text_store));
```
skotopes: There is one thing that I don't like about sizeof: compiler will not inform you if you change char[] with char*.

skotopes (changes requested):

skotopes:
Un-draft when ready

applications/debug/rpc_debug_app/rpc_debug_app.c
CookiePLMonster: Fair point. Should I change all the occurrences back? Or does `COUNT_OF` account for it? AFAIK it doesn't.

skotopes: yep, roll em back please

skotopes (approved):

Issue #3847, opened 2024-08-19
LFRFID: EM4100 detected as GProxII


korden32:
### Describe the bug.

Looks like #3796 broke detection for EM4100 tags starting with 3D.

### Reproduction

- get EM4100 tag starting with 3D (e.g. create EM4100 with ID 3D00000000 and write to a T5577 or emulate on another flipper/pm3)
- read it
- during ASK search it'll be detected as GProxII, not EM4100

skotopes:
fixed in #3869

PR #3868, opened 2024-09-03
LFRFID: Guard GProxII Wiegand Check Against False Positive and Correct 36-bit Parsing

zinongli:
# What's new

- Add Wiegand checks to prevent false positive readings because this protocol have all zeros as parity bits. 
- Also fix 36-bit format wrong parsing. See [here](https://github.com/flipperdevices/flipperzero-firmware/pull/3868#issuecomment-2325846048) for detail.
- Wiegand code snippet copied from [indala](https://github.com/flipperdevices/flipperzero-firmware/blob/dev/lib/lfrfid/protocols/protocol_indala26.c#L272-L287)
- See also: #3847 #3869

# Verification 

- Try it with these two dumps before and after this PR. Both are supposed to be EM4100:
[test_dumps.zip](https://github.com/user-attachments/files/16844683/test_dumps.zip)

# Checklist (For Reviewer)


zinongli:
Update and explanation for my commit b3831c5: The previous version parsed 36-bit format incorrectly. The original author seems to have interpreted 36-bit as meaning the "profile" being 36 bits long. But [here](https://github.com/RfidResearchGroup/proxmark3/blob/master/client/src/cmdlfguard.c#L477-L478) and [here](https://github.com/RfidResearchGroup/proxmark3/blob/master/client/src/cmdlfguard.c#L165-L173) we can see it acutally meant it's a 36-bit wiegand. This can also be referenced from [card info's record of Chubb 36-bit](https://web.archive.org/web/20200304005131/http://cardinfo.barkweb.com.au/index.php). 

To verify, use PM3 and write a T5577 with this command: `lf gproxii clone --xor 236 --fmt 36 --fc 8 --cn 22744`
If you are on the version before this PR, it will read as FC: 128, which is wrong. And it actually fails wiegand but because the original code didn't check for wiegand it wouldn't be tested. But with my latest commit FC will read correctly and wiegand check will pass.

zinongli:
Additional changes finished. Waiting for #3869 to be merged first to undraft.

skotopes (approved):

PR #3869, opened 2024-09-03
Fix detection of GProx II cards and false detection of other cards

Astrrra:
# What's new

- GProx II parser no longer works in the opposite way from how it should due to the inverted parity check

# Verification 

- See #3847

# Checklist (For Reviewer)


zinongli:
Hi! Coincidentally I just opened a PR #3868 for this issue. I think your fix is probably where the real bug has been. I didn't question the parity check and thought going for the wiegand would fix it. I still think wiegand check is needed though, as the parity check would pass for tags that has many zeros in raw. 

What do you think? Should we combine the two PR?

Astrrra:
@zinongli hi, was just about to comment on your PR as well :)

Yes, having the Wiegand check (and the 36-bit parsing fixes) would be a very good idea, however, I think it would be a good idea to move the check into a separate function instead of keeping it inside decoding function.

I think the easiest way to solve this all would be to first merge my PR, and then merge yours afterwards.

zinongli:
Totally agree. Mine has the wrong parity check. I'll mark mine as draft before yours being merged. And yes, I'll make that check into a function. 

Side note, we probably should have something like PM3's wiegand.c in /lib so we don't have to reinvent wheels.

skotopes (approved):

Issue #3343, opened 2024-01-05
15693 listener can get get into a failed in some noisy conditions and remain there until restart


nvx:
### Describe the bug.

While tracking down some bugs I suspect were introduced with the NFC refactor I discovered an interesting condition that can occur in the 15693 emulation code.

In `signal_reader_callback` in `iso15693_parser.c` when a SOF is recieved it then sets the expected demodulation mode to either 1or4 or 1or256 depending on what the SOF looked like. After this mode has been set remaining calls to the callback look for data encoded in that modulation only. The issue is that readers will generally only use one of the two modulations exclusively (for example picopass only supports the higher speed 1of4 mode) and there is no timeout that resets the state back to `Iso15693ParserStateParseSoF`, so under real world conditions sometimes a bit of noise can trigger the wrong mode at which point emulation will stop working until it is manually restarted as we never encounter the appropriate end of frame.

I'd expect that the demodulator would upon encountering invalid modulation reset the state back to `Iso15693ParserStateParseSoF` so that it can start from the beginning again.

### Reproduction

1. Attach debugger
2. Run picopass and emulate a credential
3. Read the credential a bunch of times from varying distances (note at some point you may run into flipperdevices/flipperzero-good-faps#105 as well)
4. When emulation stops working (everything looks fine on the FZ side of things, but the reader will not be able to see anything) add a breakpoint on the `signal_reader_callback` function
5. Observe that `instance->state` is `Iso15693ParserStateParseFrame` and `instance->mode` is `Iso15693ParserMode1OutOf256` which isn't used by picopass and that in all subsequent calls to this function the state and mode remains unchanged

PR #3885, opened 2024-09-08
Add API to enforce ISO15693 mode

aaronjamt:
# What's new

- Adds a new API to the NFC parser to allow applications to enforce a specific parsing mode. Some apps, such as Picopass, will always use one of the modes, so disabling autodetection can help to prevent issues resulting from incorrectly autodetecting the state (i.e. due to noise). It also includes a way to re-enable autodetection, should that be useful in the future.

# Verification 

- Build a modified version of the Picopass app which calls the new `nfc_iso15693_force_1outof4` method and emulate a card. Scanning the card on the reader no longer causes the Flipper to crash when the reader has LF enabled.

# Checklist (For Reviewer)


nvx:
A bit of background, picopass only ever uses 1of4 modulation not 1of256. As readers will never use 1of256 if noise ever makes it appear like a 1of256 SOF was sent it will hang there indefinitely as the reader will never send a 1of256 EOF causing https://github.com/flipperdevices/flipperzero-firmware/issues/3343

It is probably worth also having a timeout (we know the maximum frame size, if we don't see the rest of the frame in this time we should reset the emulation state ready for the next frame) and fix what looks like a buffer overrun in the 1of256 code (https://github.com/flipperdevices/flipperzero-good-faps/issues/105), but even with those two things fixed I still think it makes sense allowing emulation code to opt out of 1of256 when it's known it should never happen as accidentally interpreting something as the 1of256 SOF would still likely interrupt emulation until the timeout occurs which would almost certainly cause the reader to time out.

A quick look at the patch seems fine to me, but I'm not an expert on the HAL design.

bettse:
This should be an easier one to review since it is just adding a new API and not changing how anything functions

lib/signal_reader/parsers/iso15693/iso15693_parser.c
```diff
@@ -26,6 +26,7 @@ typedef enum {
 struct Iso15693Parser {
     Iso15693ParserState state;
     Iso15693ParserMode mode;
+    bool detect_mode;
```
RebornedBrain, line 29: Can this detect_mode be moved to the Iso15693ParserMode enum?

aaronjamt: Sure, so you're proposing to add two new modes like `Iso15693ParserModeForced1OutOf4` and `Iso15693ParserModeForced1OutOf256` in place of the boolean?

RebornedBrain: Yes, but there is a more important thing, which I've missed previously. Putting protocol specific functions into nfc.c / nfc.h layer it's not very good idea, because nfc layer was designed to be a generic protocol free. I had a quick conversation with original nfc stack designer and he agreed on that.
Also, we came up to the idea that we need to investigate this problem deeper and try to fix this in iso15 parser. Currently I'm investigating this issue and will be in touch with you on the results.

skotopes:
Fixed in https://github.com/flipperdevices/flipperzero-firmware/pull/3988

nvx:
> Fixed in #3988

While this fixed the crash bug in the picopass code, there are certain cards based on 15693 (eg, picopass) that does does not support the 1of256 mode so technically without this functionality it would always be possible to detect emulation vs a real card if the reader tested if the card responds to 1of256 mode.

PR #3896, opened 2024-09-13
[FL-3891] Folder rename fails

portasynthinca3:
# What's new
  - Split the `storage_common_equivalent_path` API function with a bool parameter at the end into two: `storage_common_equivalent_path` and `storage_common_is_subdir`
  - Fix the subdirectory checking functionality that setting the aforementioned parameter to `true` intends to achieve
  - By doing that, fix the bug where a folder rename would fail in some cases (e.g. `/ext/test` -> `/ext/test-test`)
  - A test for the aforementioned bug
  - Refactored storage tests a bit

# Verification 
  - Run unit tests

Or (in either qFlipper or the CLI):
  - Create a `/ext/test` folder
  - Try to rename the folder into `/ext/test-test`
  - Vierfy that the operation completes successfully
  
# Remarks
  - I decided to split the function into two because previously it was not immediately clear that the semantics of the function change drastically if the last argument is set to `true`. Among other things, the order of the string parameters suddenly starts to matter, which is not clear from their names (`path1` and `path2`). It still doesn't matter in the equivalence relation, but it does matter in the subdirectory relation.
  - That said, the implementation of both of these API functions is quite similar, so I decided not to add another message type in order to reduce code size.

# Checklist (For Reviewer)

### DIFF
--- a/applications/debug/unit_tests/tests/storage/storage_test.c
+++ b/applications/debug/unit_tests/tests/storage/storage_test.c
@@ -6,8 +6,8 @@
 // This is a hack to access internal storage functions and definitions
 #include <storage/storage_i.h>
 
-#define UINTTEST_RESOURCES_PATH(path) EXT_PATH("unit_tests/" path)
-#define UNIT_TESTS_PATH(path)         EXT_PATH(".tmp/unit_tests/" path)
+#define UNIT_TESTS_RESOURCES_PATH(path) EXT_PATH("unit_tests/" path)
+#define UNIT_TESTS_PATH(path)           EXT_PATH(".tmp/unit_tests/" path)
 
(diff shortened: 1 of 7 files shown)


skotopes:
Unit tests please

portasynthinca3:
> Unit tests please

I already added unit tests for the folder rename functionality. Or are you referring to the new API?

skotopes:
> > Unit tests please
> 
> I already added unit tests for the folder rename functionality. Or are you referring to the new API?

yep

portasynthinca3:
> yep

Done!

skotopes (dismissed):

skotopes (approved):

PR #3906, opened 2024-09-17
目覚め時計

skotopes:
# What's new

- FuriHal: RTC Alarm API
- Settings: Clock Alarm settings
- Alarm app

# Verification 

- Set alarm and wait for it to trigger
- Set alarm and poweroff flipper, wait for flipper to power on and alarm to trigger

# Checklist (For Reviewer)


targets/f7/furi_hal/furi_hal_rtc.c
```diff
@@ -1,3 +1,4 @@
+#include "furi_hal_interrupt.h"
```
CookiePLMonster:
```suggestion
#include <furi_hal_interrupt.h>
```

So it's consistent with the other includes below.

targets/f7/furi_hal/furi_hal_rtc.h
```diff
@@ -98,6 +98,14 @@ void furi_hal_rtc_deinit_early(void);
 /** Initialize RTC subsystem */
 void furi_hal_rtc_init(void);
 
+/** Prepare system for shutdown
+ *
+ * This function must be called before system sent to transport mode(power off).
+ * FlipperZero implementation configures and enables ALARM output on pin PC13
+ * (Back button). This allows system to woke-up charger from transport mode.
```
CookiePLMonster:
```suggestion
 * (Back button). This allows the system to wake-up charger from transport mode.
```

targets/f7/furi_hal/furi_hal_rtc.h
```diff
@@ -247,6 +255,37 @@ void furi_hal_rtc_set_datetime(DateTime* datetime);
...
+typedef void (*FuriHalRtcAlarmCallback)(void* context);
+
+/** Set alarm callback
+ *
+ * Use it to subscribe to alarm trigger event. Setting alarm callback is
+ * independent from setting alarm.
+ *
+ * @warning    Normally this callback will be delivered from ISR, however we may
+ *             deliver it while you calling this function. This happens when
+ *             alarm already triggered, but there were no ISR set.
```
CookiePLMonster:
```suggestion
 * @warning    Normally this callback will be delivered from the ISR, however we may
 *             deliver it while this function is called. This happens when
 *             the alarm has already triggered, but there was no ISR set.
```

targets/f7/furi_hal/furi_hal_rtc.c
```diff
@@ -395,6 +445,79 @@ void furi_hal_rtc_get_datetime(DateTime* datetime) {
...
+    FURI_CRITICAL_EXIT();
+}
+
+bool furi_hal_rtc_get_alarm(DateTime* datetime) {
+    furi_check(datetime);
+
+    memset(datetime, 0, sizeof(DateTime));
+
+    datetime->hour = __LL_RTC_CONVERT_BCD2BIN(LL_RTC_ALMA_GetHour(RTC));
+    datetime->minute = __LL_RTC_CONVERT_BCD2BIN(LL_RTC_ALMA_GetMinute(RTC));
```
CookiePLMonster, line 482: Seconds are set in the setter, but they are not read here.

CookiePLMonster (commented):
The basic alarm function works well, incl. waking up the Flipper when it was shut down.

A few notes/ideas, knowing full well that this PR is WIP so I'm aware that some of those may already be planned:
* A "bell" icon, like the Bluetooth/SD card icons are placed currently
* Perhaps allow applications to use ALMB while ALMA is "reserved" for the inbuilt Clock app. Make it possible for the app to choose if it wants to get the default "wake up!" screen and wake the Flipper up, or if it wants to control that by itself
* Allow applications to subscribe to the alarm ISR at least for ALMB, right now it seems to be dominated by the Clock app
* If ALMA and ALMB get exposed to apps, reset them when resetting Flipper settings, and maybe also offer a "Reset alarms" option in the Clock app, if some app leaves alarms configured in a weird state
* One shot timers? Both in furi rtc, and in the Clock app

skotopes:
@CookiePLMonster

Regarding your previous message: despite of the fact that there are 2 alarm slots, having both of them working in a same way is impossible. Alarm output can only be configured to work with one source.

Also since we don't have way to start app on alarm, I'll hide this API from public use.

PS: I'm open to new creative ideas on how to make this thing usable, but super busy and don't have time to polish this PR further

CookiePLMonster:
> Alarm output can only be configured to work with one source.

Indeed, tricky. I think this could be solved by not treating the alarm callback as an ISR (from the API standpoint), but rather `furi_rtc_alarm` could be managing this ISR by itself. This way, the alarm ISR callback becomes part of the firmware, and it can further call to user-specified callbacks for ALMA and ALMB. Of course same restrictions as `furi_hal_isr` apply, so the apps need to un-register this callback before exiting etc.

applications/settings/application.fam
```diff
@@ -5,6 +5,8 @@ App(
     provides=[
         "passport",
         "system_settings",
+        "clock_settings",
+        "clock_settings_start",
```
hedger: Move to clock_settings' `requires=`/`provides=`

applications/settings/clock_settings/application.fam
```diff
@@ -0,0 +1,16 @@
+App(
+    appid="clock_settings",
+    name="Clock & Alarm",
+    apptype=FlipperAppType.SETTINGS,
+    entry_point="clock_settings",
+    requires=["gui"],
+    stack_size=1 * 1024,
```
hedger, line 8: here

gsurkov (dismissed):

hedger (approved):

PR #3918, opened 2024-09-23
NFC: iso14443_4a improvements. Canvas: extended icon draw.

RebornedBrain:
# What's new

- Added one new function for drawing mirrored xbm bitmaps.
- Now 4a listener throws Halt and FieldOff events to the upper levels
- iso14443_4a poller logic is now enhanced and supports more blocks according to standard

# Verification 

- See https://github.com/flipperdevices/flipperzero-firmware/pull/3744

# Checklist (For Reviewer)


CookiePLMonster:
1. Why not simply expose the `IconRotation rotation` argument in `draw_xbm`?
2. It's upside down - is it correct to call it mirrored then? Entering the PR, I expected a draw that would be mirrored horizontally, not rotated 180deg.

RebornedBrain:
> 1. Why not simply expose the `IconRotation rotation` argument in `draw_xbm`?

It was done in order not to break api backward compatibility.


> 2. It's upside down - is it correct to call it mirrored then? Entering the PR, I expected a draw that would be mirrored horizontally, not rotated 180deg.

If you take a look into `canvas_draw_u8g2_bitmap` which is called under the hood, there will be such code: 
image
In the switch there is a call to `canvas_draw_u8g2_bitmap_int` which has bool parameter mirror, and this parameter is true in case of `IconRotation180`. Here is the reason why I called this function mirror.

RebornedBrain:
Also this PR solves the same problem described [here ](https://github.com/flipperdevices/flipperzero-firmware/pull/3744) 
In addition it provides possiblity to send also S-block and R-block

CookiePLMonster:
> It was done in order not to break api backward compatibility.

You can make `IconRotation0` a default parameter and break ABI, but not sources.

applications/services/gui/canvas.c
```diff
@@ -517,6 +517,20 @@ void canvas_draw_xbm(
     canvas_draw_u8g2_bitmap(&canvas->fb, x, y, width, height, bitmap, IconRotation0);
 }
 
+void canvas_draw_xbm_custom(
```
skotopes: canvas_draw_xbm_ex please

targets/f7/api_symbols.csv
```diff
@@ -1,5 +1,5 @@
 entry,status,name,type,params
-Version,+,74.0,,
+Version,+,77.0,,
```
skotopes: f18 too please

skotopes (changes requested):

applications/services/gui/canvas.c
CookiePLMonster: Also it's a nice practice to have the base function call the `_ex`, so refactor `canvas_draw_xbm` to do nothing except call `canvas_draw_xbm_ex` with `IconRotation0`.

skotopes: yep

skotopes (approved):

PR #3942, opened 2024-10-13
Furi: A Lot of Fixes

skotopes:
# What's new

- BT Service: cleanup code
- Dialog: correct release order in file browser, fix various crashes and bus faults
- Rpc: rollback to pre #3881 state, fix deadlock on gui rpc shutdown
- Kernel: fix inverted behavior in furi_kernel_is_running, make it usable
- Log: properly take mutex when kernel is not running, prevent crash if kernel is not fully running
- Thread: rework tread control block scrubbing procedure, ensure that we don't do stupid things in idle task, add new priority for init task, prevent various deadlocks
- Timer: add control queue flush method, force flush on stop, prevent use after free if timers are incorrectly used
- Furi: system init task now performs thread scrubbing, delay now ensures that you are not using it from the idle task
- BleGlue: add some extra checks
- FreeRTOSConfig: fix bunch of issues that were preventing configuration from being properly applied and cleanup dead things

# Verification 

- Full set of integration tests
- No more lockup on BLE RCP streaming and profile change at the same time
- No more rare bus faults on app start

# Checklist (For Reviewer)

### DIFF
--- a/targets/f7/inc/FreeRTOSConfig.h
+++ b/targets/f7/inc/FreeRTOSConfig.h
@@ -84,6 +84,7 @@ to exclude the API function. */
 #define INCLUDE_xTaskGetCurrentTaskHandle   1
 #define INCLUDE_xTaskGetSchedulerState      1
 #define INCLUDE_xTimerPendFunctionCall      1
+#define INCLUDE_xTaskGetIdleTaskHandle      1
 
 /* Workaround for various notification issues:
  * - First one used by system primitives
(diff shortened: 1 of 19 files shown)


targets/f7/inc/FreeRTOSConfig.h
```diff
@@ -84,6 +84,7 @@ to exclude the API function. */
 #define INCLUDE_xTaskGetCurrentTaskHandle   1
 #define INCLUDE_xTaskGetSchedulerState      1
 #define INCLUDE_xTimerPendFunctionCall      1
+#define INCLUDE_xTaskGetIdleTaskHandle      1
```
CookiePLMonster, line 87: Can likely be set to `0`.

gsurkov (dismissed):

DrZlo13 (approved):

PR #3841, opened 2024-08-14
[FL-3893] JS modules

portasynthinca3:
# What's new
  - JS modules: `event_loop`, `gpio`,`gui` and `storage`
    - JS `gui` submodules: `gui/loading`, `gui/submenu`, `gui/empty_screen`, `gui/text_input`, `gui/text_box`, `gui/dialog`
  - Documentation for these modules
  - TS type declarations for these modules
  - JS example scripts for these modules (except `storage`)
  - Some tests for some of the modules
  - Removed legacy JS modules: `submenu` (superseded by `gui/submenu`), `textbox` (superseded by `gui/text_box`), `dialog` (superseded by `gui/dialog`)
  - `TextInputModel` field: `minimum_length`
  - SDK API functions: `text_input_set_minimum_length` , `furi_event_loop_is_subscribed`
  - Static inline convenience function: `furi_event_loop_maybe_unsubscribe`

# Verification 
  - Run unit tests
  - For modules that require access to the outside world (`gpio`, `gui`):
    - Build and flash the firmware
    - Run `gpio` and `gui` JS example scripts
    - Verify that the examples work as intended

# Remarks
## Performance
The GUI example is quite heavy, requiring about 32 KiB of heap space with all the GUI modules loaded. The script starts up slowly compared to other scripts (~500ms), but runs just fine.

## Readability
The later the date of creation of a module, the more its implementation looks like it's written in a DSL. Without all these macros the code would just become 80% noise, and thus a huge pain in the butt to write and maintain.

# Checklist (For Reviewer)

- [x] PR has description of feature/bug or link to Confluence/Jira task
- [x] Description contains actions to verify feature/bugfix
- [x] I've built this code, uploaded it to the device and verified feature/bugfix

PR #3961, opened 2024-10-17
JS: Backport³ and more additions & fixes

WillyJL:
# What's new

- backport (here) of backported missing things ([there](https://github.com/Next-Flip/Momentum-Firmware/pull/270)) of great backport & rework ([over yonder](https://github.com/flipperdevices/flipperzero-firmware/pull/3841)) xD
- i separated into individual commits that make sense on their own, might be easier to look there instead of whole diff
- original authors/contributors credited with co-author where relevant
- badusb:
  - numpad keys support
  - keyboard layout support
  - altPrint() and altPrintln() support
  - quit() function to unlock usb profile (original intent was to alternate badusb profile and mass storage profile, we have this in momentum and other cfw as usbdisk module, but outside scope of this pr, can backport this module later on if you're interested)
- serial
  - readAny() to avoid starving read loops with many small reads or hinder usability with few large reads
  - end() to release serial handle allowing new initialization in same script, maybe for gpio usage or different serial port or settings
  - expansion service is automatically disabled to avoid conflict
- storage
  - added simple example script we had previously, ported to new remade storage module
- gui:
  - text_input:
    - fixed null ptr crash when user did not pass max len prop. logic of props is from what i can see, always configure view for minimum usable state, props should configure things on top as optional
    - defaultText and defaultTextClear props
  - byte_input
  - file_picker
    - there is no file picker based on View object, so this is only gui/* module that is not based on view_factory
    - instead it uses usual dialogs app file picker in sync manner
  - viewDispatcher.currentView to allow users to have extra logic in navigation callbacks for example
  - updated gui example script with showcase of all additions above
- globals:
  - toString() works correctly on negative numbers, toString(-42) used to give "0"
  - parseInt()
  - toUpperCase() and toLowerCase()
  - example for string functions
  - __filepath and __dirpath to allow relative path contextualization, for example for load() or storage.open()
- example interactive.js repl script, allows running js in limited capacity without editing files
- many additions to typedefs that were missing or incomplete, still not everything is documented i believe but getting closer

### Needs feedback:

- toString(), toUpperCase(), toLowerCase() could all be moved to respective classes i think, for strings there is getprop_builtin_string() in mjs source, can add there, for number there is no such function but may be possible to add another if branch in getprop_builtin() for number type
- __dirpath and __filepath are set when launching script in global scope, but means that when executing scripts with load(), these have values of original script, since scope lookup is recursive (see mjs_find_scope() and its usages). i think better behavior would be load() gives child script its own correct __dirpath and __filepath but there are some issues:
  - load() accepts a custom scope parameter. with custom scope, assignment in child script is confined to custom scope, lookup is recursive to parent scopes. with no custom scope, code is executed in current scope, so everything is shared like code was inlined
  - for custom scope, we can get custom scope object passed by user and inject __filepath and __dirpath into it, easy
  - for no custom scope, i dont see any good solution: one way is saving current __dirpath and __filepath values from scope, changing them, running load(), then resetting to previous values. might work fine in most cases but if child script runs load() of its own and uses callbacks or things like this, child2 might run a function defined in child1 and then its screwed up. this is a rare edge case of course, if no function callback is passed from child1 to child2 then no code from child1 could run while child2, and second instance of mjs_load() would restore child1's __dirpath and __filepath correctly

# Verification 

- unit tests
- example scripts

# Checklist (For Reviewer)


applications/system/js_app/js_thread.c
```diff
@@ -197,13 +198,95 @@ static void js_require(struct mjs* mjs) {
...
+    double num = mjs_get_double(mjs, mjs_arg(mjs, 0));
     char tmp_str[] = "-2147483648";
     itoa(num, tmp_str, base);
     mjs_val_t ret = mjs_mk_string(mjs, tmp_str, ~0, true);
     mjs_return(mjs, ret);
 }
 
+static void js_parse_int(struct mjs* mjs) {
+    mjs_val_t arg = mjs_arg(mjs, 0);
+    if(!mjs_is_string(arg)) {
```
portasynthinca3: I recommend that you use the new JS ASS (Argument Schema for Scripts) macros, they will reduce these 11 lines of argument validation code down to just 1. Look at the new modules that I wrote, they heavily utilize ASS.

Also, use one of the new `strint_to_*` functions. They already implement this, you just need a JS wrapper around one.

applications/system/js_app/js_thread.c
```diff
@@ -197,13 +198,95 @@ static void js_require(struct mjs* mjs) {
...
+
+static void js_to_upper_case(struct mjs* mjs) {
+    mjs_val_t arg0 = mjs_arg(mjs, 0);
+
+    size_t str_len;
+    const char* str = NULL;
+    if(mjs_is_string(arg0)) {
+        str = mjs_get_string(mjs, &arg0, &str_len);
+    }
+    if(!str) {
```
portasynthinca3: Same here, use ASS

applications/system/js_app/js_thread.c
```diff
@@ -197,13 +198,95 @@ static void js_require(struct mjs* mjs) {
...
+    free(upperStr);
+    mjs_return(mjs, resultStr);
+}
+
+static void js_to_lower_case(struct mjs* mjs) {
+    mjs_val_t arg0 = mjs_arg(mjs, 0);
+
+    size_t str_len;
+    const char* str = NULL;
+    if(mjs_is_string(arg0)) {
```
portasynthinca3: Same here, use ASS

applications/system/js_app/examples/apps/Scripts/gui.js
```diff
@@ -49,15 +61,28 @@ eventLoop.subscribe(views.demos.chosen, function (_sub, index, gui, eventLoop, v
     } else if (index === 2) {
         gui.viewDispatcher.switchTo(views.keyboard);
     } else if (index === 3) {
-        gui.viewDispatcher.switchTo(views.longText);
+        gui.viewDispatcher.switchTo(views.bytekb);
     } else if (index === 4) {
+        gui.viewDispatcher.switchTo(views.longText);
+    } else if (index === 5) {
+        let path = filePicker.pickFile("/ext", "*");
```
portasynthinca3, line 66: I'm curious whether it's at all possible to integrate the file picker with the rest of the GUI system as an ordinary view and make its API asynchronous. Gonna look into it.

applications/system/js_app/types/gui/index.d.ts
```diff
@@ -3,11 +3,28 @@ import type { Contract } from "../event_loop";
...
+     * Assign value to property by name
+     * @param property Name of the property
+     * @param value Value to assign
+     */
     set<P extends keyof Props>(property: P, value: Props[P]): void;
+    /**
+     * Check if property is available
+     * @param name Name of the property
+     */
+    hasProperty(name: string): boolean;
```
portasynthinca3: I think it's possible to compute the return value of this function at compile time with something to the tune of:

```ts
hasProperty<K extends string>(name: K): K extends keyof Props ? true : false;
```

Not sure why this API method would exist at all, given that TypeScript will check things :thinking:

applications/system/js_app/types/global.d.ts
```diff
@@ -18,15 +18,53 @@ declare function print(...args: any[]): void;
...
+ * @brief Converts a string to a number
+ * @param text The string to convert to a number
+ */
+declare function parseInt(text: string): number;
+
+/**
+ * @brief Transforms a string to upper case
+ * @param text The string to transforms to upper case
+ */
+declare function toUpperCase(text: string): string;
```
portasynthinca3: Like you said, these probably belong in the String class.

portasynthinca3 (changes requested):

applications/system/js_app/js_thread.c
```diff
@@ -231,11 +314,32 @@ static int32_t js_thread(void* arg) {
     struct mjs* mjs = mjs_create(worker);
     worker->modules = js_modules_create(mjs, worker->resolver);
     mjs_val_t global = mjs_get_global(mjs);
+    if(worker->path) {
```
portasynthinca3, line 259: This is kinda nitpicky, but wouldn't it be better to name these `__dirname` and `__filename` for consistency with Node?

WillyJL: It's funny, this was actually the intention when @freehuntx suggested adding this, then @jamisonderek contributed this to momentum with `path` naming and not being js developers we never noticed, and when it was brought up it was too much hassle for what it's worth to change it. Now that we're breaking all scripts with gui and storage and gpio, might aswell change it, true.

applications/system/js_app/types/global.d.ts
WillyJL: Any thoughts on moving toString()? That's how it is in ecmascript too and would be good for consistency I think

applications/system/js_app/js_thread.c
WillyJL: Saw this one coming lol, we've had these for a long time and never bothered to update it, I was debating fixing it before submitting the pr but decided to atleast submit it first to get some opinions

applications/system/js_app/types/gui/index.d.ts
WillyJL: Typescript is useful to the developer making the script, but maybe they want to make their scripts backwards or forwards compatible, or work across firmwares. Without this, someone adding a custom prop locks scripts to their firmware. With this, the script can check and work on all firmwares.

WillyJL: For example, the keyboard we have in momentum supports characters that would break file paths, so they're disabled by default but can be enabled with an api call. I want to add a prop for this, but now any script using this will not work on other firmwares. I'd rather have a way to check availability than make a walled garden around me.

We could go onto "why not add this keyboard functionality to ofw?", and that is a valid point, but besides the objective of this function. This is just an example, there's bound to be more cases in the future, maybe a prop is added in ofw dev but isn't in ofw stable yet, no need to wait to update your script.

applications/system/js_app/types/global.d.ts
WillyJL: moved string things

applications/system/js_app/js_thread.c
WillyJL: while moving to string class, code had to be moved to /lib/mjs, while ASS is part of js_app specifically, so i could not use it.
i made similar code to what was already in mjs string source code, so it looks off compared to normal flipper C but if you look at the rest of the file it fits in. let me know if this is a problem

applications/system/js_app/js_thread.c
WillyJL: i could not find a way to specify an optional argument and give it a default value using ASS, so for now i used an if check and used 2 ASS statements. looking at what it expands to, this looks wasteful (generates lots of duplicated code), but not sure what the best thing to do is

applications/system/js_app/types/gui/index.d.ts
portasynthinca3: I'm thinking that there needs to be a more generic mechanism for detecting discrepancies between the JS API that a script expects, and the JS API that the firmware provides. This is one of the problems that I'm currently trying to solve while making a full-fledged JS SDK.

I'm thinking we should introduce a JS SDK major-minor versioning system and check whether the two are compatible on script startup. Kinda like how native apps do this already. Something along the lines of:
```ts
if (amICompatible("flipperdevices", 1, 0)) { // specify SDK vendor, major and minor versions
    // use new features!
} else {
    // depending on the use case, either make do with older features,
    // or error out.
}
```
Or, just maybe:
```ts
// food for thought: do we need to vendor features? they usually get ported between firmwares
if (apiHasFeature("flipperdevices", "feature-name")) {
  // use feature!
} else {
  // don't use feature :(
}
```
I'm thinking the former is going to be a bigger pain in the butt for JS developers, and the latter will increase the firmware image size compared to the former (which is a major consideration).

WillyJL: Being a cfw maintainer for almost 2 years, I can say I have never used the api versioning of firmware in a productive way. I get the intent of it, but due to the status quo of app catalog and multiple cfw existing, it just means we must all keep the same api version so apps can run. It's more of a way to tell which ofw version the cfw supports, and further than that it's anybody's guess, run a cfw app somewhere else and it either works or gets a missing imports error.

How you describe it here adding a way to check the "vendor" aswell as version, would make this slightly better: on the backend, cfw could keep track of which other vendor's versions it fully supports, like flipperdevices 1.0 and otherfw 1.1.

But again, as a cfw maintainer I can already tell this will get old and annoying *very* quickly. As I mentioned in the pr description, for module functions one can just check if they are set: `if (module.function) module.function(stuff)`, without any versioning. I feel like this is more intuitive for the script developer, and far easier to maintain. This hasProperty() would be a counterpart for it, since there is no list of props exposed to js at runtime.

skotopes: @Willy-JL talking about api versioning, I think we can find compromise:
- API versioning in JS got different nature, not like api versioning in API symbols, it should be much easier to maintain it
- We'll have checkSdk() but as an optional canary
- If you call it we'll check version and ask user what to do if version doesn't match
- If your script doesn't contain it, then no checks is done and you'll crash later
- I'd like to see all of us cooperating on JS API: keep it consistent across all forks, simplifying portability between different firmware flavors. It make sense to expose additional API to query which firmware flavor we are running at

WillyJL: alright that sounds like it could work, ill remove hasProperty() and hold off on implementing this extra prop i wanted until #3963 is consolidated

WillyJL:
@portasynthinca3 @skotopes with ofw 1.1 close to release, and the significant changes to battery driver, i want to release momentum 008 as close after 1.1 as possible, and this for me includes atleast agreeing on user-facing changes of this PR: we of course already had these things that this PR adds, and dont want users to fix scripts on this release, and then change the same things again in next release.

hasProperty() is gone, __dirname and __filename names are fixed, toUpperCase() and toLowerCase() are moved to string class. this is most of review comments.

although i did not get any feedback on moving toString() to number class, i got it working, so please let me know if you agree with this or not.

behavior of __dirname and __filename in load() can be fixed later on.

for filepicker, i can label in our changelog that its changed but may change again depending on [this](https://github.com/flipperdevices/flipperzero-firmware/pull/3961#discussion_r1805579153).

does this sound fine?

skotopes:
@Willy-JL I hear you. We can postpone release till this series of PRs will be ready to merge.
Also we are open to discussion if we any changes needed on our side. I've explained to @portasynthinca3 where we aiming and she we'll be discussing changes with you guys so we all be able to schedule development

WillyJL:
i just wanted some feedback on whether these user-facing changes would be ok, no need to rush this pr or delay release, i did not mean to disrupt plans :smile: 

but besides this, from what i understand this is basically ready, only missing feedback on whether moving of toString() to number class is ok, and whether there is a way to move file picker to new gui system

lets see what porta thinks :D

portasynthinca3:
To be honest, did not mean to delay the release either.

portasynthinca3:
@Willy-JL, here's our statement on the whole JS API / SDK / versioning mess.

## Goals

### In short
  - We want to maintain a base JS API that is consistent across all firmware flavors.
  - We want JS app developers to be able to leverage additional functionality found in custom firmware editions.
  - We want users not to think about cross-FW compatibility of the scripts that they find online too much.

### The user's perspective
  - A JS app that works on OFW should also work on a CFW.
  - A JS app that works on a CFW and leverages its extended features may or may not work on OFW.
  
### The JS developer's perspective
  - A JS app that does not leverage any additional CFW functionality is guaranteed to work across all firmware editions.
  - A JS app that _requires_ additional CFW functionality will not work on OFW.
  - One can write JS apps that _can make use of_ but do not _require_ additional CFW functionality.

### The CFW maintainer's perspective
  - CFWs are free to introduce additional JS APIs.
  - CFWs are discouraged from breaking the base JS API found in OFW.
  - CFWs are encouraged to backport any of their extended features into OFW, assuming they are compatible with OFW's vision.

## How do we achieve that?
Like I already said, I propose that we introduce a major-minor semantic versioning system like the one that we already have for native apps. Since we have adopted TypeScript, I propose that we consult the following standard to determine whether a change is a major or a minor one: https://www.semver-ts.org/.

Version compatibility could be checked by the developer using a combination of the following functions:
  - `checkSdkCompatibility(expectedMajor, expectedMinor)`: checks if the JS API version is compatible with the one that the script expects (i.e. `expectedMajor == actualMajor && expectedMinor >= actualMinor`), and if it doesn't, warns the user and asks whether they want to continue executing the script anyways. This is the simplest of the three checks, and it will be automatically inserted in the beginning of the output file unless the developer explicitly asks our JS "toolchain" (SDK) not to do so. By asserting compatibility by default, we guarantee that the user receives a clear warning if their firmware is incompatible with the script instead of cryptic error messages like `calling non-callable`.
  - `isSdkCompatible(expectedMajor, expectedMinor)`: performs the same check, but just returns the result as a boolean.
  - `sdkCompatibilityStatus(expectedMajor, expectedMinor)`: performs the same check, but returns a more detailed description (`"compatible"`, `"firmwareTooOld"`, `"firmwareTooNew"`)
  
To check for additional features found in CFWs, I propose that we introduce the following symbols:
  - `checkSdkVendor(expectedVendor)`: checks if the JS API vendor matches the one that the script expects, and if it doesn't, warns the user and asks whether they want to continue executing the script anyways. This is the simplest of the two checks for cases where an app _requires_ a feature found in a CFW. This check will not be automatically inserted by our SDK (unlike `checkSdkCompatibility`).
  - `sdkVendor`: a string containing the SDK vendor (e.g. `"flipperdevices"` or `"momentum"`). A JS app that is able to work without additional features on a best-effort basis may query this global constant to find out whether it can use those features.
  
Let us know what you think.

WillyJL:
This sounds like a great plan! There's only one thing i see as suboptimal:

Many features added by cfw are shared between other cfw pretty quickly, checkSdkVendor() would allow the developer to ask for the feature set of a certain firmware, and a different firmware could detect this and return a positive result even if it's not the firmware that the script asked for, rather if it supports the features present in said firmware. For example, unleashed includes most, but not all, extra js features that there are in momentum. A script requesting unleashed vendor features would therefore work on momentum, which is a superset of their features, but not viceversa.
That's good to have of course, but the alternative of letting the user check based on name might mean they do a simple `if (sdkVendor === "unleashed")` to use a certain unleashed feature, which even though its present in momentum too, would return false and not be allowed to use such features.
Perhaps something similar to what there is for versioning could work: checkSdkVendor() is a blocking call and will warn the user and possibly quit the script, while isSdkVendorCompatible() could perform the same check and return a boolean.

portasynthinca3:
That's a valid point. However, I'm afraid that CFWs falsely presenting themselves as other CFWs, although done in good faith, will quickly devolve into a [mess similar to what `User-Agent` is today](https://www.reddit.com/r/webdev/comments/2cd46k/eli5_why_are_user_agent_strings_so_broken_and/).

I propose that we declare OFW the baseline for the feature set, and allow CFWs to introduce and share _named_ features. 

Suppose that you as the maintainer of Momentum decide to add a regex validator prop to the text input view. You shall, under this social contract, introduce a named feature (e.g. `"text-input-regex"`) that developers may check compatibility with using `doesSdkSupport("text-input-regex")`. They may also use `checkSdkFeatures(["text-input-regex", "something-else"])` to display a warning screen to the user if the API does not support the required features.

Now, to keep firmware image sizes under control, we need a framework for removing these strings from the code. In other words, we need a mechanism for recognizing a feature as a baseline feature after it has been merged into OFW. I propose that we recognize a backported feature as a baseline feature one SDK major version after it has been merged. Consider the following chain of events as an example:
  1. Both OFW and Momentum are at JS SDK version 1.0.
  2. Momentum introduces a feature called `"text-input-regex"` and bumps its JS SDK version to 1.1.
  3. The feature is backported into OFW, bumping its JS SDK version to 1.1 as well.
  4. Both OFW and Momentum continue to report `true` for `doesSdkSupport("text-input-regex")` until their JS SDK versions are bumped up to 2.0
  5. Starting with JS SDK 2.0, `doesSdkSupport("text-input-regex")` starts reporting `false` and `checkSdkFeatures(["text-input-regex"])` starts warning the user that a feature is unsupported. I assert that the negative effect from these incorrect results is alleviated thanks to SDK version checks, which will fail if the script expects v1.x but is actually running on 2.x.
  
This leaves the question of vendoring unanswered. Assuming that the system for user-agent-line-creep prevention that I outlined is in place, it would make no sense to keep the `sdkVendor` and `checkSdkVendor` symbols. If a script wants to query this data for purely informational purposes, it would do so using a new `require("flipper").firmwareEdition` API, again, for purely informative purposes, not to be relied on for querying the existence or absence of features.

portasynthinca3:
@Willy-JL, @xMasterX and I came to a conclusion on this outside of GitHub. I'll be implementing this shortly.

applications/system/js_app/js_thread.c
```diff
@@ -194,14 +196,19 @@ static void js_require(struct mjs* mjs) {
...
-    itoa(num, tmp_str, base);
-    mjs_val_t ret = mjs_mk_string(mjs, tmp_str, ~0, true);
-    mjs_return(mjs, ret);
+static void js_parse_int(struct mjs* mjs) {
+    const char* str;
+    int32_t base = 10;
+    if(mjs_nargs(mjs) == 1) {
+        JS_FETCH_ARGS_OR_RETURN(mjs, JS_EXACTLY, JS_ARG_STR(&str));
+    } else {
+        JS_FETCH_ARGS_OR_RETURN(mjs, JS_EXACTLY, JS_ARG_STR(&str), JS_ARG_INT32(&base));
```
portasynthinca3: This might explode due to the way mJS represents strings and ASS handles argument deconstruction. mJS terms (`mjs_val_t`s) are 8 bytes long, and if the string is sufficiently small (<= 5 bytes), it's stored in the term itself instead of on the heap. `mjs_get_string` requires a pointer to the term for this precise reason: if the string is <= 5 bytes long, it mJS needs to know where the term is located to return a pointer to the string. `JS_FETCH_ARGS_OR_RETURN` expands to (among other things) an array of terms with the arguments. Since the scope of the array is limited to this `else` block, the pointer that ASS assigns to `str` will no longer be valid as soon as this scope is left. This is not an issue when the macro is placed at the top of the function.

applications/system/js_app/modules/js_gui/js_gui.h
```diff
@@ -9,12 +9,15 @@ typedef enum {
...
     JsViewPropTypeNumber,
     JsViewPropTypeArr,
+    JsViewPropTypeTypedArr,
+    JsViewPropTypeBool,
 } JsViewPropType;
 
 typedef union {
     const char* string;
     int32_t number;
     mjs_val_t array;
```
portasynthinca3: More of a nitpick, but maybe it's a good idea to rename this field to `term` now that not just arrays could be stored here (typed arrays are not exactly arrays :smiling_face_with_tear:)
```suggestion
    mjs_val_t term;
```

applications/system/js_app/types/global.d.ts
```diff
@@ -12,21 +12,41 @@ declare function delay(ms: number): void;
...
+ * @brief Path to the current script file
+ */
+declare const __filename: string;
 
 /**
  * @brief Reads a JS value from a file
  * 
  * Reads a file at the specified path, interprets it as a JS value and returns
- * said value.
+ * the last value pushed on the stack.
```
portasynthinca3: I don't think the user needs to know the implementation details of the JS interpreter. Is there another way to convey this idea?

portasynthinca3 (changes requested):
Additional comments, in order of decreasing importance:
  - I don't like how with the max length thing the fields have to be arranged in a specific way for the thing to work. Field ordering is absolutely not something that the ECMAScript specification guarantees. This is especially important with the new build pipeline underway.
  - Not an issue with this particular PR, just a thing to note in regards to #3963: `.d.ts` files would have to be moved, and the new declarations would have to get `@version` doc comments.
  - I don't like that the file browser is not a view. But that's an issue for later, I don't think that this PR needs to change this.

LGTM otherwise

applications/system/js_app/js_thread.c
WillyJL: should be still on stack even though its not in scope anymore, but yeah better to fix this. ill keep the top one and handle the second argument separately

applications/system/js_app/types/global.d.ts
WillyJL: i dont think its too far fetched. i myself, for interactive.js, was previously writing something like "{run: function() return [USER CODE HERE];}" to file and load()ing it, then doing .run() on it. and this had many problems (for that usecase) like weird scoping, not being able to define your own functions easily, and formatting being a bit weird too. only after i had dived very deep into how load() works, i found that it is not only to be used like the load.js and load_api.js example. and since there is no other example really, there is no indication to the user of what you can really do with load(). at the end of the day, load() works exactly like an exec() or eval(), but from file, and some users may benefit from such a usecase. to me, the existing comment paired with the provided example just sounds like load() can only evaluate one value. but instead, you can evalutate a bunch of statements liek any other file, and then "return" something at the end.

but i do agree that there might be a better way to explain this. maybe "Reads a file at the specified path and runs it as JS, returning the last evaluated value"?

WillyJL:
> * I don't like how with the max length thing the fields have to be arranged in a specific way for the thing to work. Field ordering is absolutely not something that the ECMAScript specification guarantees. This is especially important with the new build pipeline underway.

mJS does, properties are parsed from code top to bottom and stored in a linked list inserting at the head, so at the end of object parsing the properties are in exact inverse order from what one would expect. inversing the processing order of props in makeWith() also doesn't seem smart, would be so time inefficient due to it being a singly linked list from what i can see. but yes, not ideal to rely on this behavior, and also it is counterintuitive to the user.

i see 2 possible solutions:
- changing the examples to show view.set("defaultText", "text"); after the first makeWith() call
- changing behavior of defaultText and defaultData assign to resize the buffer to atleast the size specified by the default provided. for example, default is 4 bytes for byteinput, default data is 10 long so enlarge to 10, then we also have length prop provided, so enlarge or shrink as requested. this brings up another problem however: what if the user provides a defaultdata longer than the length they specify? if length is assigned first, then it will be resized later by default data. if defaultdata is assigned first, then it will be shrunk later by length. and we cannot solve this by enforcing the size specified by length, because thats the original issue: if defaultdata is assigned before than user provided length, then length is still default of 4 and default data is cutoff despire the user specifying otherwise.

so really there is no solution. this is a chicken and egg problem. best compromise to me soudns liek resizing to atelast size specified by default data, and if user also provides a length that is incompatible with this, it is their issue to deal with order of processing



> * Not an issue with this particular PR, just a thing to note in regards to [[FL-3918] Full-fledged JS SDK + npm packages #3963](https://github.com/flipperdevices/flipperzero-firmware/pull/3963): `.d.ts` files would have to be moved, and the new declarations would have to get `@version` doc comments.

if the intention is to merge that pr before this one then ill make those changes as required



[FL-3918]: https://flipperzero.atlassian.net/browse/FL-3918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

applications/system/js_app/types/global.d.ts
portasynthinca3: Yeah, I have to confess, I have not dived into `load`. Thanks for the explanation! Yeah, your new suggestion sounds good.

WillyJL: just remembered about caching. might be good to also specify on a second doc line that "Loaded JS files are cached and not re-evaluated."

portasynthinca3:
@Willy-JL 

> > Field ordering is absolutely not something that the ECMAScript specification guarantees
>
> mJS does

We prefer to rely on specifications and not implementation details :) Like I said, gonna become very important with our new tools that are going to be processing JS code before mJS even sees it.

With regards to the solution, I think we can just keep the buffer the maximum size out of all sizes that we know of (user-requested max length, user-provided default text, etc.) and explicitly throw an error if the requested max length is smaller than the length of the provided default text.

> if the intention is to merge that pr before this one then ill make those changes as required

We'll see how it goes :) I don't know which one will get merged first, that depends on Aku.

WillyJL:
ah well, that sounds good enough. ill change it again to track default data length separately and error if length is specified less than that

WillyJL:
one thing im not sure of: string returned by mjs to c is not guaranteed to be null terminated. props pass just a const char* with no length value. this sounds like a problem waiting to happen to me. like will my strlen() for default text overflow? idk probably? lol

WillyJL:
@portasynthinca3 i dont have my flipper with me right now for another few hours, logic should be ok but i cannot test right now.

only thing that remains is my concern with strings passed as props possibly not being nul terminated... will this be an issue?

portasynthinca3:
> one thing im not sure of: string returned by mjs to c is not guaranteed to be null terminated. props pass just a const char* with no length value. this sounds like a problem waiting to happen to me. like will my strlen() for default text overflow? idk probably? lol

@Willy-JL This is a valid concern based on the doc comment. However, after reviewing the actual implementation of `mjs_get_string` and some parts of the mJS' memory manager, I can't see a single code path where a string wouldn't get null terminated:
  - `MJS_TAG_STRING_I` and `MJS_TAG_STRING_5`: embedded in the lower 5 bytes of term. The tag is stored in the upper 2 bytes, leaving a byte in the middle which will be 0 b/c terms are zero-initialized, at least in `mjs_mk_string`.
  - `MJS_TAG_STRING_O`: GC'ed owned strings, which are always stored with the terminator at runtime. The parser does store them without the terminator in the same giant buffer with all the other strings, but we never use raw internally referenced strings from the parser. The bytecode interpreter does, but it converts them into nice null-terminated strings with `mjs_mk_string` for us.
  - `MJS_TAG_STRING_F`: foreign strings, not managed by mJS at all. The main use case is to avoid wasting RAM for strings that you already have in `.rodata`. One could sneak a non-null-terminated string into `mjs_mk_string(/*...*/, false)`, but I'd just call that irresponsible, honestly. Either way, nothing in our codebase does that; not us, not mJS itself.
  - `MJS_TAG_STRING_C`: seems to be used internally during garbage collection, never seen by us.
  - `MJS_TAG_STRING_D`: all uses are commented out, the symbols in the commented out code are no longer there, I can't tell.

Perhaps we're dealing with outdated documentation here. I personally have never gotten a string from mJS that isn't null terminated, and **I'm convinced that my analysis shows that this can never happen**. We might want to ask the mJS team about this, but their repo seems dead.

P.S.: Holy cow, is mJS code ever hard to read ;_;

WillyJL:
Sounds great, thanks for the very detailed break down!
Then we should be good on this pr? I can test the changes I added in about an hour when I have my flipper nearby.

portasynthinca3 (dismissed):
LGTM, assuming the functionality works as intended.

WillyJL:
just tested on device and seems to work great!

applications/system/js_app/examples/apps/Scripts/gui.js
```diff
@@ -5,8 +5,11 @@ let loadingView = require("gui/loading");
 let submenuView = require("gui/submenu");
 let emptyView = require("gui/empty_screen");
 let textInputView = require("gui/text_input");
+let byteInputView = require("gui/byte_input");
 let textBoxView = require("gui/text_box");
 let dialogView = require("gui/dialog");
+let filePicker = require("gui/file_picker");
```
skotopes, line 11: fails here:

```
71536 [E][PluginManager] Failed to load /ext/apps_data/js_app/plugins/js_gui__file_picker.fal
71541 [E][JS modules] Module gui/file_picker load error. It may depend on other modules that are not yet loaded.
71547 [E][JS] Exec error: "gui/file_picker" module load fail
71550 [E][JS] Stack trace:
        at /ext/apps/Scripts/gui.js:11

```

applications/system/js_app/modules/js_gui/file_picker.c
```diff
@@ -0,0 +1,47 @@
...
+#include <assets_icons.h>
+
+static void js_gui_file_picker_pick_file(struct mjs* mjs) {
+    const char *base_path, *extension;
+    JS_FETCH_ARGS_OR_RETURN(mjs, JS_EXACTLY, JS_ARG_STR(&base_path), JS_ARG_STR(&extension));
+
+    DialogsApp* dialogs = furi_record_open(RECORD_DIALOGS);
+    const DialogsFileBrowserOptions browser_options = {
+        .extension = extension,
+        .icon = &I_file_10px,
```
skotopes, line 12: can not resolve this symbol

skotopes (changes requested):

applications/system/js_app/modules/js_gui/file_picker.c
WillyJL: Right... I'm used to having icons exposed to api but that's not in ofw. I'll do what @xMasterX did, paste icon data as code into the plugin file

WillyJL: Unless, does fap_libs=["assets"] work for plugins?

skotopes: you can try )

WillyJL:
@skotopes fixed, works fine now (and i learned my lesson to actually flash what i code, not that but ported to another fork xD)

also i mirrored the JS SDK changes porta made but for the additions of this PR, hope it looks fine

skotopes:
;-) 
I have more questions to @portasynthinca3, since she promised to review this PR

skotopes (approved):

skotopes:
Good job everyone!
Really awesome to see JS becoming more and more usable

PR #3968, opened 2024-10-22
Input_SRV: migration to event_loop

Skorpionm:
# What's new

- Input_SRV: migration to event_loop

# Verification 

- Test GUI navigation using physical buttons on Flipper

# Checklist (For Reviewer)


applications/services/input/input.c
```diff
@@ -75,95 +72,185 @@ const char* input_get_type_name(InputType type) {
...
+    }
+
+    furi_event_loop_subscribe_semaphore(
+        instance->event_loop,
+        instance->input_semaphore,
+        FuriEventLoopEventIn,
+        input_semaphore_callback,
+        instance);
+
+    instance->key_sequence = malloc(sizeof(InputSrvKeySequence) * input_pins_count);
```
CookiePLMonster, line 241: All those allocations could easily land on the stack, VLAs come in handy here (copying my remark from the previous PR).

Skorpionm: no, if you put everything on the stack and turn on debugging, then the 1 KB stack allocated to the process is sometimes not enough and everything crashes with special effects

CookiePLMonster: Those structures are tiny - so wouldn't this mean that even without them on the stack, 1KB is pushing the limit very close?

Skorpionm: all structures on the stack will take about 220 bytes.

provided that there are only 6 buttons, when the number of buttons changes, the size of the structure will increase and if the structure is placed on the stack, the implementation ceases to be universal.

Skorpionm: at the expense of 1kb on the stack it is enough even with debugging enabled, without it about 500 bytes are used per process

CookiePLMonster: In this case the keys could stay on the heap (although maybe put them on the service heap?) while `instance` goes to the stack? It does not have to be all or nothing.

Skorpionm: Well then what's the point of chopping it into pieces?

CookiePLMonster: Currently it's chopped into pieces already, and occupies 3 heap allocations - so by moving you save one allocation, which is better than nothing.

skotopes: Keep in mind that services are allocated in special memory region and it's quite small. Increasing stack size may bring another set of problems

CookiePLMonster: Sure, in this case the key arrays could stay on the user heap, but the inputsrv is so small there's no reason not to put it on the stack. Before this PR it was just a bunch of local variables.

applications/services/input/input.c
```diff
@@ -75,95 +72,187 @@ const char* input_get_type_name(InputType type) {
...
-int32_t input_srv(void* p) {
-    UNUSED(p);
-
-    const FuriThreadId thread_id = furi_thread_get_current_id();
-    FuriPubSub* event_pubsub = furi_pubsub_alloc();
-    uint32_t counter = 1;
-    furi_record_create(RECORD_INPUT_EVENTS, event_pubsub);
+static void input_isr_key(void* context) {
+    InputSrv* instance = context;
+    furi_semaphore_release(instance->input_semaphore);
```
gsurkov, line 81: Interrupts may be missed in some edge cases

gsurkov (changes requested):

applications/services/input/input.c
Skorpionm: fix

PR #3923, opened 2024-09-27
Add NDEF plugin

bettse:
# What's new

- Bringing NDEF plugin from fork upstream

# Verification 

- Install
- Read NTAG21x or NTAGI2C... tag with NDEF content.  See URL displayed in UI

Screenshot-20240927-155124
Photo of reading Minecraft Earth figure using build of this branch

# Checklist (For Reviewer)

- [ ] PR has description of feature/bug or link to Confluence/Jira task
- [ ] Description contains actions to verify feature/bugfix
- [ ] I've built this code, uploaded it to the device and verified feature/bugfix

PR #3973, opened 2024-10-26
NFC: Add NDEF Parser for MFUL, MFC and SLIX

WillyJL:
# What's new

- NFC parser for NDEF structured data (backport from CFW, followup from #3923, and with many additions and improvements)
- Supports:
  - Mifare Ultralight (including NTAG2xx and NTAG I2C)
  - Mifare Classic (1k tested, mini should work the same, 4k not tested as i couldn't find examples but should work and MAD2 is parsed too)
  - SLIX (all types should be supported)
  - Generic ISO15693 could be supported too, but i could not find any specification or examples, so only SLIX is currently supported but can easily be adapted to any other ISO15693
- Parses:
  - URI (including Phone and Mail links) records
  - Text records
  - BT MAC address records
  - Contacts
  - WiFi credential records
  - Empty records
  - SmartPoster records
  - Unknown records are dumped without parsing
- Due to structure of supported_cards plugins, this needs to produce multiple plugins for the different card types
- Also means only code needed for this card is loaded when parsing, so it is a little less RAM heavy than having single plugin

Screenshot-20241026-021026
Screenshot-20241026-021153
Screenshot-20241026-021317
Screenshot-20241026-021118

# Verification 

- Read different types of supported NFC tags that contain NDEF records of different types
- In particular, would be good to check mifare mini and mifare classic 4k as i could not find examples to test, all other card types and record i have tested myself

# Checklist (For Reviewer)


skotopes (dismissed):

skotopes (approved):

skotopes:
@Willy-JL this one is really good: we've been talking about ndef support for a very long time ) Really cool that you've made it first

PR #3987, opened 2024-11-03
Update MFC key dictionary

jblanked:
# What's new

- Implemented MFC keys from https://github.com/ikarus23/MifareClassicTool/blob/master/Mifare%20Classic%20Tool/app/src/main/assets/key-files/extended-std.keys

# Verification 

# Checklist (For Reviewer)


equipter:
a lot of these are not validated as static in any way. "some random card" being the origin for one of these is scary enough. 

just because they're from MCT doesn't make them reliable enough to blindly merge imo.

noproto:
We need verified static keys, these are not and will only offset the improvements we just made in MFC. PM3 has many verified static keys and it's a good reference dictionary to use.

WillyJL:
> We need verified static keys, these are not and will only offset the improvements we just made in MFC. PM3 has many verified static keys and it's a good reference dictionary to use.

last time i checked a couple months ago, pm3 dictionary is a superset of mct dictionary. that is to say, everything from mct is in pm3 dict, plus pm3 one has more.

EDIT: nevermind, there are 19 total keys that are in mct dict and not in pm3. still, considering that pm3 has roughly 2000 keys 19 is so little that IMO this argument of "we should trust pm3 instead of mct" might not mean much.

onovy:
See my proposal in #4013

PR #3992, opened 2024-11-07
Loader: Fix BusFault in handling of OOM

WillyJL:
# What's new

- during OOM propagation, ELF file has small chance of trying to free relocation data that is not allocated:
- `elf_load_section_data()` does not allocate `fast_rel->data` if not enough ram
- but `elf_preload_section()` allocated `fast_rel` itself, so both would try to be freed when only 1 exists
- in particular, this was observable in upython app

# Verification 

- try to run upython app
- either it opens, or it gives out of memory message, not crash

# Checklist (For Reviewer)


WillyJL:
there is something else to consider... basically this was doing `aligned_free(NULL)`... maybe this situation should be checked? then rather than BusFault you will get something like `free(0)` message, similar to `malloc(0)`

skotopes (dismissed):

skotopes:
Fixing aligned_alloc make more sense than adding cratches to loader

skotopes (approved):

PR #3999, opened 2024-11-12
[JS] Make sdk.js serialport selection more universal.

csaboka21:
# What's new

- On windows when using NPM for creating JS project and starting with 'npm start' it will throw this error:

```
file:///C:/Users/hazic/Documents/temp/NodeTest/my-flip-app/node_modules/@flipperdevices/fz-sdk/sdk.js:95
    let portPath = flippers[0].path;
                               ^

TypeError: Cannot read properties of undefined (reading 'path')
    at Object.upload (file:///C:/Users/hazic/Documents/temp/NodeTest/my-flip-app/node_modules/@flipperdevices/fz-sdk/sdk.js:95:32)
    at async file:///C:/Users/hazic/Documents/temp/NodeTest/my-flip-app/node_modules/@flipperdevices/fz-sdk/sdk.js:175:5
```

It's clear that it can't find the flipper. The code searching for ports which 'serialNumber' starts with "flip_", but in windows it the port looking like this:

```
{
    path: 'COM4',
    manufacturer: 'Microsoft',
    serialNumber: '7&359A65E1&0&0000',
    pnpId: 'USB\\VID_0483&PID_5740&MI_00\\7&359A65E1&0&0000',
    locationId: '0001.0000.0000.007.000.000.000.000.000',
    friendlyName: 'Soros USB-eszköz (COM4)',
    vendorId: '0483',
    productId: '5740'
  }
```

It shows that the serial number is not starting with 'flip_'.
Another problem with the code is that checking for `if (!flippers)` is useless because when we create the array, a line above it should check for its length instead.

Also, when multiple flippers are connected and found when calling the prompts and assigning it to the port, it is simply not initialized yet. It will be initialized a few lines later.

I think it would be much better if the user has to define the COM port all the time they want to run on the device.

# Verification 

- Tried on my Windows PC.

# Checklist (For Reviewer)

### DIFF
--- a/applications/system/js_app/packages/fz-sdk/sdk.js
+++ b/applications/system/js_app/packages/fz-sdk/sdk.js
@@ -85,22 +85,14 @@ async function build(config) {
 
 async function upload(config) {
     const appFile = fs.readFileSync(config.input, "utf8");
-    const flippers = (await SerialPort.list()).filter(x => x.serialNumber?.startsWith("flip_"));
+    let ports = await SerialPort.list();
 
-    if (!flippers) {
-        console.error("No Flippers found");
-        process.exit(1);
-    }
-
-    let portPath = flippers[0].path;
-    if (flippers.length > 1) {
-        port = (await prompts([{
-            type: "select",
-            name: "port",
-            message: "Select Flipper to run the app on",
-            choices: flippers.map(x => ({ title: x.serialNumber.replace("flip_", ""), value: x.path })),
-        }])).port;
-    }
+    let portPath = (await prompts([{
+        type: "select",
+        name: "port",
+        message: "Select Flipper to run the app on",
+        choices: ports.map(x => ({ title: x.path, value: x.path })),
+    }])).port;
 
     console.log(`Connecting to Flipper at ${portPath}`);
     let port = new SerialPort({ path: portPath, baudRate: 230400 });


skotopes:
@portasynthinca3 review please

portasynthinca3:
@csaboka21 We have investigated this issue and found out that it's caused by the incorrect handling of Flipper serial numbers by Windows in the case of some driver combinations. We believe that the fix that you have provided will lead to lots of garbage serial ports in the selection dialog on most platforms. We will close this PR and provide our own fix, which is to fall back to filtering by VID:PID (which is a generic STM VCP VID:PID pair) if no Flippers are found after filtering by serial number. Nevertheless, thank you for pointing out this issue and providing your own fix!

portasynthinca3:
@csaboka21 Could you please update your Flipper application to use `fz-sdk` version 0.1.3 and confirm that the problem is no longer there?

csaboka21:
@portasynthinca3 I have tested it now and it's working perfectly! Thank you very much for the fast fix!

csaboka21:
@portasynthinca3 I found another small issue with it. When you have multiple flipper and the code find all of it will run to error here: (in line 108-118)

```
let portPath = flippers[0].path;
    if (flippers.length > 1) {
        port = (await prompts([{
            type: "select",
            name: "port",
            message: "Select Flipper to run the app on",
            choices: flippers.map(x => ({ title: x.serialNumber.replace("flip_", ""), value: x.path })),
        }])).port;
    }
    console.log('Connecting to Flipper at ${portPath}');
    let port = new SerialPort({ path: portPath, baudRate: 230400 });
```


So if multiple flipper found it will go in the if but there the port is not initialized yet. (It will be 10 row later)
So the `let port = new SerialPort({ path: portPath, baudRate: 230400 });` should be before the if statement?
Or better in the if statement we only change the `portPath` variable.

PR #4030, opened 2024-12-10
FBT: Don't lint JS packages

WillyJL:
# What's new

- support for excluding folders in linter with `!` prefix
- exclude js packages dir as they have js naming convention (using dashes) so fbt would always complain

# Verification 

- `./fbt format` does not complain of `applications/system/js_app/packages/fz-sdk` and `applications/system/js_app/packages/create-fz-app`

# Checklist (For Reviewer)

### DIFF
--- a/firmware.scons
+++ b/firmware.scons
@@ -29,6 +29,8 @@ env = ENV.Clone(
     TARGETS_ROOT=Dir("#/targets"),
     LINT_SOURCES=[
         Dir("applications"),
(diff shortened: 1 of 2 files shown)


skotopes (approved):

skotopes:
To be honest `!` symbol is quite problematic in terms of how shell handles it.
It will work here, since args passed directly to subprocess without shell involvement.

PR #4048, opened 2024-12-30
nfc: Fix sector overrun in MFC nested dictionary attack

GMMan:
# What's new

- Fixes sector overrun that occurs when running nested dictionary attack on MIFARE Plus ~~X~~ 2k variants in SL1.

# Verification 

1. Provision a MIFARE Plus ~~X~~ 2k variants in SL1
2. Write a Vingcard credential to the card
3. Read the card as MIFARE Classic

# Checklist (For Reviewer)

I'm not sure what the actual intended flow is supposed to be when the dictionary attack runs off the end of the card. I assume it would loop back and continue to check the first part of the dictionary, but it could also just move on to nonce collection, although that would involve some refactoring to free up the loaded dictionaries. Probably something for @noproto to clarify.

Edit: I've discovered that the test card in question is actually a MIFARE Plus EV1. Flipper's detection of this is not quite working, and because the card gave the same historical bytes as MIFARE Plus X, it was incorrectly assumed to be that chip.

### DIFF
--- a/lib/nfc/protocols/mf_classic/mf_classic_poller.c
+++ b/lib/nfc/protocols/mf_classic/mf_classic_poller.c
@@ -1897,21 +1897,22 @@ NfcCommand mf_classic_poller_handler_nested_controller(MfClassicPoller* instance
             dict_attack_ctx->nested_phase = MfClassicNestedPhaseDictAttack;
         }
     }
-    if(dict_attack_ctx->reuse_key_sector == instance->sectors_total) {
-        // Reset target sector to first sector whose key has not been found
-        for(dict_attack_ctx->reuse_key_sector = 0;
-            dict_attack_ctx->reuse_key_sector < instance->sectors_total &&
-            mf_classic_nested_is_target_key_found(instance, true);
-            dict_attack_ctx->reuse_key_sector++)
-            ;
-        // Reset to sane value just in case we happen to have all of the keys
-        if(dict_attack_ctx->reuse_key_sector == instance->sectors_total) {
-            dict_attack_ctx->reuse_key_sector = 0;
-        }
-    }
     if((dict_attack_ctx->nested_phase == MfClassicNestedPhaseDictAttack ||
         dict_attack_ctx->nested_phase == MfClassicNestedPhaseDictAttackResume) &&
        (dict_attack_ctx->nested_target_key < dict_target_key_max)) {
+        if(dict_attack_ctx->reuse_key_sector == instance->sectors_total) {
+            // Reset target sector to first sector whose key has not been found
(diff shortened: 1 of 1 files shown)


noproto:
Possible regression with this PR or 3822 for Hardnested. Waiting on a card I ordered 2 weeks ago to isolate a crash.

edit: 2K cards aren't recognized, which seems like the root cause for any sector overrun? 
https://github.com/flipperdevices/flipperzero-firmware/blob/5fb9558dbb1b48dca1d48e6bffad9f7bf9d9f7a0/lib/nfc/protocols/mf_classic/mf_classic.h#L51

GMMan:
> edit: 2K cards aren't recognized, which seems like the root cause for any sector overrun?
> 
> https://github.com/flipperdevices/flipperzero-firmware/blob/5fb9558dbb1b48dca1d48e6bffad9f7bf9d9f7a0/lib/nfc/protocols/mf_classic/mf_classic.h#L51

Potentially addressed in #4053

noproto:
This should be the correct fix for this issue (rather than fixing a byproduct of the crash): https://github.com/noproto/xero-firmware/commit/56fe7b09ae896ffdeb76522a56ab41b8a31dcc55

Tested and would welcome feedback.

RebornedBrain:
Hi @GMMan , thanks for your contribution, and sorry for such a long waiting. Could you please take a look at @noproto proposal about this fix and test, whether it works in your case?

GMMan:
> Hi @GMMan , thanks for your contribution, and sorry for such a long waiting. Could you please take a look at @noproto proposal about this fix and test, whether it works in your case?

That code does not crash, but it also causes some keys to not be detected. I've previously mentioned to @noproto that he can open another PR with the new code and I can close this one whenever that PR is made.

RebornedBrain:
@GMMan nice, thank you! @noproto, could you please open a pr with your fix?

hedger (approved):

noproto:
@hedger this is the fix, the PR is out of date: https://github.com/flipperdevices/flipperzero-firmware/pull/4048#issuecomment-2570877840

Edit: Opened https://github.com/flipperdevices/flipperzero-firmware/pull/4288

noproto:
Because this PR was superseded by #4288 and merged in https://github.com/flipperdevices/flipperzero-firmware/commit/f4c92dcd76d9c29e184d245cb493d0aa86809c40 I recommend we close #4048 (no longer required).

noproto:
Thanks @GMMan ! If you find any regressions ping me and I'll investigate. I have a source for purchasing MIFARE Plus cards now.

PR #4050, opened 2024-12-30
nfc: Enable MFUL sync poller to be provided with passwords

GMMan:
# What's new

- Consumers of the MIFARE Ultralight card read sync poller can specify a password to authenticate with

# Verification 

- When reading a MFUL card with the sync poller, pass in a `MfUltralightPollerAuthContext` with `skip_auth` set to `false` and the required password/3DES key. The card should automatically be unlocked during reading.

# Checklist (For Reviewer)

Sidenote: why not make a separate struct just for passing passwords? Because for some reason the same struct is being used to supply passwords for MFUL elsewhere, even when most of its members aren't used. It's kind of weird.


lib/nfc/protocols/mf_ultralight/mf_ultralight_poller_sync.c
```diff
@@ -23,6 +23,7 @@ typedef struct {
     FuriThreadId thread_id;
     MfUltralightError error;
     MfUltralightPollerContextData data;
+    const MfUltralightPollerAuthContext* auth_context;
```
RebornedBrain, line 26: `mf_ultralight_poller_sync_read_card` is used in unit_tests only and why to add new parameter here, if this function is then called with `NULL` populated in this parameter?

GMMan: Although there are no system usages of `mf_ultralight_poller_sync_read_card` currently, it is exposed as an API available for user apps and NFC plugins.

skotopes (dismissed):

skotopes (approved):

PR #4053, opened 2024-12-31
nfc: Attempt to detect MFP 2k in MFC mode

### LINKED ISSUE #4012: Flipper Zero Mifare Plus 2K support (SL1 ONLY)

I know flipper has supported Mifare Classic 1K systems for awhile. This is mostly due to the insecure CRYPTO1 algorithm being used to secure the card. However, it seems that Mifare Plus systems are now slowly being used in place of Classic 1k and these are supposed to be encrypted with 128bit AES encryption. However, there is a catch. SL1 Mifare Plus systems are still using the insecure Crypto1 standard and currently Flipper zeros can read/ decrypt these SL1 cards in full by telling the Flipper they are actually Classic 1K (you may need to grab the keys using "detect reader"). The only problem with this is when doing some research online, I learned that certain SL1 cards make use of 2 additional partitions (16 and 17) and it doesn't seem as though the flipper can read these yet. Is it possible to add the extras in the future for experimentation purposes?

### Anything else?

_No response_

GMMan:
# What's new

- Read extra accessible sectors on a MIFARE Plus 2k compared to MIFARE Classic 1k

# Verification 

- Read a MIFARE Plus 2k in SL1 as a MIFARE Classic card. It should be detected as a MIFARE Plus 2k.

# Checklist (For Reviewer)

Not sure whether MIFARE Plus 2k has 17 or 18 sectors. On my sample MIFARE Plus EV1 2k, sector 17 is not accessible, however that may be an effect of some sectors set to SL3. Someone with other samples will have to determine how many sectors there actually are.

Fixes #4012

### DIFF
--- a/lib/nfc/protocols/mf_classic/mf_classic.c
+++ b/lib/nfc/protocols/mf_classic/mf_classic.c
@@ -38,6 +38,14 @@ static const MfClassicFeatures mf_classic_features[MfClassicTypeNum] = {
             .full_name = "Mifare Classic 4K",
             .type_name = "4K",
         },
+    [MfClassicTypePlus2k] =
+        {
+            // TODO: need to validate whether 17 or 18 sectors are accessible
+            .sectors_total = 17,
+            .blocks_total = 544,
+            .full_name = "Mifare Plus 2K",
+            .type_name = "Plus 2K",
+        },
 };
 
 const NfcDeviceBase nfc_device_mf_classic = {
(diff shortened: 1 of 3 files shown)


GMMan:
I'd like to also point out that for a MFP EV1/EV2 in SL1, if it has the sectors used for detection set in SL3 mode, the detection in general may not work properly and potentially identify it as a MIFARE Mini. There really should be a mechanism to pass version detection from the MFP poller into the MFC poller.

PR #4078, opened 2025-01-24
Ensure that `furi_record_create` is passed a non-NULL data pointer

dcoles:
It's currently possible to use `furi_record_create` to create and initialize a `FuriRecordData` pointing to NULL.

This means its potentially possible for `furi_record_open` to return a NULL pointer which besides not being particularly useful means the Rust wrapper for `Record` can't assume that the returned record is always a non-NULL value.

If by chance this is the intended behaviour, then we can just have the Rust wrapper do a `furi_check` itself, but it seems like it would be better to eliminate this potential corner-case at the source.

# What's new

- `furi_record_create` now checks that `data` is a non-NULL pointer

# Verification 

- Calling `furi_record_create` with a NULL `data` pointer should trigger a `furi_check` system crash

# Checklist (For Reviewer)


skotopes (dismissed):

skotopes:
To be honest there were an option to pass NULL and use record as simple synchronization primitive, but @hedger changed this contract by explicitly marking furi_record_open return value as non-null.

I'll accept this PR and update docs

skotopes (approved):

dcoles:
Thanks for the background. It's nice to understand why it works that way.
Records are one of my favourite parts about working with the Furi API.

PR #4119, opened 2025-02-18
Updated Button Panel

Akiva-Cohen:
# What's new

- Updated button panel callbacks to make them consistent with button menu
- Button panel now passes 3 arguments to a button callback instead of 2, the third being an InputType
- Also updated the infrared universal remotes to work with the new code.

# Verification 

- Built for both release and dev.
- Make an app with button panel to verify it works.

# Checklist (For Reviewer)


skotopes (approved):

Akiva-Cohen:
adding the extra events for all buttons brings it back out of sync with button menu
specifically, it will call the callbacks on an inputtypelong, which button menu does not

skotopes:
> adding the extra events for all buttons brings it back out of sync with button menu specifically, it will call the callbacks on an inputtypelong, which button menu does not

Can you give me example where it breaks?

To be honest I've corrected implementation to sync it with other places where we handle things like that

Akiva-Cohen:
Oh it doesn't break
It's just inconsistent with the other thing I was trying to make it more consistent with
Although if you're trying to change it everywhere you might want to check out button_menu.c (applications/services/gui/modules/button_menu.c)(lines 260-280 specifically) and maybe update that as well

PR #4120, opened 2025-02-18
GUI: Widget view extra options for JS

WillyJL:
# What's new

- Backport of extra features from old CFW JS widget module by @jamisonderek (his implementation was based on custom view, so this PR adds them to firmware widget view instead so they work with new JS widget implementation)
  - Fill parameter on frame element for equivalent of "box" canvas function, includes rounding too
  - Circle element, also with fill parameter for equivalent of "disc" canvas function
  - Line element
  - Load icon from fxbm file (like flipper game engine) with JS

# Verification 

# Checklist (For Reviewer)


applications/services/gui/modules/widget.h
```diff
@@ -159,14 +159,36 @@ void widget_add_icon_element(Widget* widget, uint8_t x, uint8_t y, const Icon* i
...
  */
 void widget_add_frame_element(
     Widget* widget,
     uint8_t x,
     uint8_t y,
     uint8_t width,
     uint8_t height,
-    uint8_t radius);
+    uint8_t radius,
+    bool fill);
```
WillyJL: since this now covers both "frame" (not filled) and "box" (filled), might be a good idea to rename to "rect". i did not do this yet in case it would be preferred not to break api naming, but from what i see there are only 2 apps using this function and would need to update anyway to add the extra `false`, so might aswell rename?

WillyJL:
cc @portasynthinca3 @skotopes :D

applications/services/gui/modules/widget.h
portasynthinca3: Looks fine to me personally, but I'll leave the final decision on this up to @skotopes as I'm not responsible for this particular subsystem.

portasynthinca3 (dismissed):
The JS part of this PR looks fine to me with one exception: the new features haven't been reflected in the SDK versioning. Please bump the minor SDK version (both in `package.json`s and in the interpreter), add a new feature declaration and update typedocs where necessary.

WillyJL:
added .fxbm sprite loading for js, since firmware's widget view does not support just bare xbm data it constructs a icon struct wrapper to work like any other firmware icon, which could come in handy if in the future more js gui views will use icons, for users to supply their own icons without needing to add xbm support to all firmware's views.

feature name `gui-widget-extras` is admittedly a bit generic, but im not sure listing out `gui-widget-framefill-circle-line` would be much better, and having multiple sounds overkill to me for what this pr adds.

i also pre-emptively renamed widget frame elements to rect as mentioned in review comment, both in firmware and js, can easily revert just 0888f09842fec457319bcd77ee3dd5f6398e8c16 if its decided this is not desired. also, for js side of this, means that previous "frame" type is not available anymore, instead it is called "rect", i left it marked on that line as "amended in js sdk 0.3, extra feature `gui-widget-extras`" and i feel like bumping major because its technically removed would be out of place. if we go ahead with renaming to "rect" and you think just "amended" docstring is not enough im open to ideas on what else to do :)

for verification steps, should i add an example of all (circle, line, fxbm icon) to gui.js widget option? feel like the dolphin js stopwatch looks good as it is lol

applications/services/gui/modules/widget.h
skotopes: how about backward compatibility? why no to leave old method too?

WillyJL: between rect and circle it would mean there would be
- frame
- box
- circle
- disc

and while that does reflect how it is done in canvas, i feel like its redundant to have 4 separate element types and api calls when it is just a wrapper that few people will use and it could be just 2 with fill parameters. also personally i am dangerously low on free flash in my fork (under 12kb usable on compact build) so that played a part. but if you think its better to have 4 separate element types i guess thats fine

skotopes: Macroses are pretty much free ) I'll update your code if you don't mind.

WillyJL: good point, and sounds great! thanks

skotopes (approved):

skotopes:
@Willy-JL please provide examples next time ;)

PR #4144, opened 2025-03-09
Fix app debug experience when dead code is culled

GMMan:
This PR fixes the issue as experienced in the Discord conversation [here](https://discord.com/channels/740930220399525928/954430078882816021/1347800427831230496) and continued [here](https://discord.com/channels/740930220399525928/954430078882816021/1348303726913388554).

# Summary of the issue
When ld's dead code elimination occurs, it eliminates unreferenced functions and data, but does not remove the corresponding debug information. Because those functions are missing, the symbols have nowhere to point to, and defaults to the start of the section the symbol would have been in. This causes confusion because when stepping into a function at the start of the section, GDB may pick the wrong symbol that is overlapping, and therefore take the user to the source of some random function that is not what is actually being run. By enabling `--gdwarf-sections` on as, ld can also eliminate the line number info and invalidates the symbol, and prevents GDB from jumping to the wrong line. However, this alone breaks debugging entirely because GDB cannot locate any of the line number info since they are in differently named sections. ld's default script accounts for this and merges all the different `.debug_line` sections together, but this is missing from the linker scripts used to build firmware and apps, so the default debug-related rules have been added into the scripts, and now everything works as expected.

# What's new

- Split DWARF line info sections by function so they can be eliminated when the corresponding functions are eliminated
- Fix DWARF line info loading in GDB by adding default ld rules for debug sections, which combine individual `.debug_line` sections

Note: `.comment` and `.ARM.attributes` sections removed from defaults accordingly due to existing rules.

# Verification 

[Test app](https://gist.github.com/GMMan/325027d1a84c254baf39c5f0447c8e95)

1. Build and upload attached test app to a Flipper
2. Attach to the Flipper using a debug probe
3. Load the application
4. Add a breakpoint at `test1_app()`
5. Resume the device until it hits the breakpoint
6. Step into `test1_app_init()`
7. The current line should point inside `test1_app_init()` rather than `test1_app_test()`

# Checklist (For Reviewer)


hedger:
Nice find, thank you.

hedger:
Big thanks for a clean and concise test case.

hedger (approved):

PR #4153, opened 2025-03-21
NFC: Fix NDEF parser for MIFARE Classic

WillyJL:
# What's new

- Fix usage of `div()` in the NDEF MIFARE Classic parser, I did not notice when contributing in #3973 that it was not exposed to API
- Fixed some mistakes i made with handling of edge cases with MIFARE Classic NDEF parsing
- Unrelated, simplified some some duplicated code in the MFC data generator
- Raw non-text data in NDEF records is now printed using pretty format, as a table with hex on the left and ascii on the right
- Fixed NDEF strings that end with a \0 byte being considered raw bytes, they print as text now
- Unrelated, fixed pretty format having skewed table on the last line, if data length is not a multiple of amount of bytes per line:
  - Before:
image

  - After:
image

# Verification 

- MIFARE Classic dump with NDEF data parses the NDEF data correctly (example file in zip below)
- Creation of MFC files with add manually works correctly
- Confirm display of raw data with pretty format (example file in zip below)

[4153_test_files.zip](https://github.com/user-attachments/files/19499815/4153_test_files.zip)

# Checklist (For Reviewer)


GMMan:
Somewhat curious why the stdlib function is being used instead of just regular `/` and `%` operators. I imagine the compiler would be able to optimize with regular operators since it should be able to map it to `__aeabi_idivmod()`.

WillyJL:
> Somewhat curious why the stdlib function is being used instead of just regular `/` and `%` operators. I imagine the compiler would be able to optimize with regular operators since it should be able to map it to `__aeabi_idivmod()`.

I needed both quotient and remainder, and my uneducated guess was that using the function dedicated to this purpose would be more efficient than doing both operators separately. if thats not the case, or if its preferred to keep `div()` out of the API, i'll change it to `/` and `%`

GMMan:
Using regular `/` and `%` appears to allow further optimizations to carry through the whole function and makes it smaller overall. The Cortex-M4 has hardware division, so any such operation gets collapsed into a single instruction instead of a function call.

Using `div()`:

image

Using `%` and `/`:

image

WillyJL:
Very interesting, thank you! I'll swap over to those then

hedger (approved):

PR #4133, opened 2025-03-01
add nfc apdu cli command back

leommxj:
# What's new

`nfc apdu` cli command just like the old one https://github.com/flipperdevices/flipperzero-firmware/pull/2482 , 14443-4a/15693-3 /14443-4b support this time. 

by issue https://github.com/flipperdevices/flipperzero-firmware/issues/3276 and discord chat, it is very useful for some people like me.

# Verification 
use `nfc apdu` to interact with a card, and see response like the following response for one unionpay credit card.

```
>: nfc apdu 00a404000e325041592e5359532e444446303100 00A4040008A000000333010102
detecting tag
Sending APDU:00a404000e325041592e5359532e444446303100 to Tag
Response: 6F35840E325041592E5359532E4444463031A523BF0C20611E4F08A000000333010102500F556E696F6E506179204372656469748701019000
Sending APDU:00A4040008A000000333010102 to Tag
Response: 6F6A8408A000000333010102A55E500F556E696F6E506179204372656469748701019F381B9F66049F02069F03069F1A0295055F2A029A039C019F3704DF69015F2D027A689F1101019F120F556E696F6E50617920437265646974BF0C0E9F4D020B0ADF4D020C0ADF6101409000
```
`nfc apdu 260100` send invetory cmd to 15693 card and get uid back.

# Checklist (For Reviewer)

- [ ] PR has description of feature/bug or link to Confluence/Jira task
- [ ] Description contains actions to verify feature/bugfix
- [ ] I've built this code, uploaded it to the device and verified feature/bugfix

Issue #3276, opened 2023-12-07
cli nfc command does not work after NFC refactor


u0d7i:
### Describe the bug.

After NFC refactor, cli "nfc" command shows empty sub-command list and does not work anymore:
```
+ device: /dev/ttyACM0
starting cli, Ctrl+C to terminate

              _.-------.._                    -,
          .-"```"--..,,_/ /`-,               -,  \ 
       .:"          /:/  /'\  \     ,_...,  `. |  |
      /       ,----/:/  /`\ _\~`_-"`     _;
     '      / /`"""'\ \ \.~`_-'      ,-"'/ 
    |      | |  0    | | .-'      ,/`  /
   |    ,..\ \     ,.-"`       ,/`    /
  ;    :    `/`""\`           ,/--==,/-----,
  |    `-...|        -.___-Z:_______J...---;
  :         `                           _-'
 _L_  _     ___  ___  ___  ___  ____--"`___  _     ___
| __|| |   |_ _|| _ \| _ \| __|| _ \   / __|| |   |_ _|
| _| | |__  | | |  _/|  _/| _| |   /  | (__ | |__  | |
|_|  |____||___||_|  |_|  |___||_|_\   \___||____||___|

Welcome to Flipper Zero Command Line Interface!
Read Manual https://docs.flipperzero.one

Firmware version: dev unknown (82baf1e9 built on 05-12-2023)

>: nfc
nfc
Usage:
nfc <cmd>
Cmd list:

>: 
```

### Reproduction

1. Connect to FZ serial port
2. Execute 'nfc' command via cli

### Target

7

### Anything else?

tested on latest devbuild ATM (latest releases and rc's have the same issues)

Tomi2965:
I can confirm - same here on latest versions. Same issue

v0lp3:
I don't think it is a bug, after that refactoring the only command available (if the debug flag is active) is "field"

u0d7i:
> I don't think it is a bug, after that refactoring the only command available (if the debug flag is active) is "field"

Before refactor it had:
```
Firmware version: 0.89.0 0.89.0 (5a58004b built on 16-08-2023)

>: nfc
nfc
Usage:
nfc <cmd>
Cmd list:
        detect   - detect nfc device
        emulate  - emulate predefined nfca card
        apdu     - Send APDU and print response 

>: 

```
[...]

PR #4158, opened 2025-03-24
[FL-3569] NFC CLI commands

RebornedBrain:
# What's new

Based on this **https://github.com/flipperdevices/flipperzero-firmware/pull/4147** and others about cli

- Now all commands related to NFC are placed in a separate shell which runs after 'nfc' command
- Command adding was simplified
- Arguments parsing is now done 'outside' of the command, you need only parse callback in command scope
- One command can hold several different actions
- **raw** commad added 
- **apdu** command added based on this implementation **https://github.com/flipperdevices/flipperzero-firmware/pull/4133/**
- **dump** command added
- **emulate** command added
- **scanner** command added
- **field** command reworked
- **mfu** command added

What's next

MFU commands
  - [x] info
  - [x] rdbl
  - [x] wrbl

# Verification 

- Run `./fbt cli` then type `nfc` and press enter
- You will see new shell with new motd
- To exit from it type `exit`, this will move back to previous shell, or press `ctrl+]`, this will close both shells
- Type `?` or `help` to see command list

Note: Description below is shorted,  try to enter `<command> -h`, in order to see more information on particular command

## **Raw**
- On Ultralight card try `raw -p 14a -sc -d 3000`, this will return the first 4 pages
- `raw -p 14a -sck -d 3000` will do the same but will keep power on tag, after that another command can be send without select sequence, like this `raw -p 14a -ck -d 3001`

### **APDU**
This should work in the same way like in the PR mentioned above. The only addition is an ability to set protocol manually
- `apdu -d 00A4000000` - detects tag, then sends data to it
- `apdu -d 00A4000000 00A4000000` - detects tag, then sends 2 bunches of data to it
- `apdu -p 4a -d 00A4000000` - sends data only if tag is iso14443_4a
- `apdu -p 4a -d 00A4000000 00A4000000` - same as above but with several bunches

### **SCANNER**
Allows to test tag and determine its type

### **DUMP**
Reads the whole card and saves it to .nfc file mentioned by user. 
For `Mifare Classic` it reads only cards which already has keycache file on flipper as the most simple way. 
For `Mifare Ultralight` tags password can be passed in order to pass auth. 

### **EMULATE**
Starts emulation of passed .nfc file

# Checklist (For Reviewer)


hedger:
#3276 related

hedger (changes requested):
Please rebase on top of `dev` and undraft

hedger (approved):
LGTM - however, when emulating a FeliCa card from CLI, a heap fragmentation occurs with block of full size 0x30. To be fixed later.

PR #4166, opened 2025-03-31
Issue 4082: Subghz Modulation Shortcut

### LINKED ISSUE #4082: Add a shortcut in SubgHz for modulation

I the subGhz, while reading, it would be nice to have a shortcut to quickly switch between modulation options. Rather than clicking on config, change the option and come back.

I was thinking about using up/down arrows to go through modulations. 

### Anything else?

_No response_

soniarosenberger:
Closes #4082 

This issue requested a shortcut feature for the subghz app (read mode) that allows the user to switch between modulations whilst actively reading,. This avoids having to go back to the configuration menu and modulation submenu, then back to reading. 
This fix was implemented by adding an up/down button press shortcut to change between modulation presets while actively reading subghz signals.
Mod-Shortcut-Subghz

# What's new

Views/receiver.c:
- Changed void subghz_view_receiver_draw(Canvas* canvas, SubGhzViewReceiverModel* model)  to draw up/down buttons and reflect sbility to change modulation type in "read" mode view
- Changed bool subghz_view_receiver_input(InputEvent* event, void* context) to handle input from up/down buttons and call corresponding modulation change function (described below)
- Added typedef struct SubGhz SubGhz; at top of file to be able to access modulation fields

Scenes/subghz_scene_receiver_config.c:
- Added void decrease_modulation_preset(SubGhz* subghz), called when down button is pressed when scanning
- Added void increase_modulation_preset(SubGhz* subghz), called when up button is pressed when scanning
- Added typedef struct SubGhz SubGhz; at top of file to be able to access modulation fields

Views/receiver.h:
-Added the following function headers
		void increase_modulation_preset(SubGhz* subghz);
		void decrease_modulation_preset(SubGhz* subghz);
- Added typedef struct SubGhz SubGhz; at top of file to be able to access modulation fields

# Verification 
-Clone this fork locally
-Run the following bash commands from a terminal in the cloned firmware folder root:
		./fbt updater_package
		tar czvf update.tgz -C dist/f7-D f7-update-local to create .tgz update package

-Open qflipper desktop app and connect flipper zero device via USB cable the same as for a regular firmware update
- Choose option to install from file
-Select update.tgz package in firmware root created in previous step
-Update your flipper and reboot
-Open subghz app on the flipper device, and select "Read" from subghz menu
-Use up/down buttons to cycle through modulation preset list either forwards or backwards
-Full preset name can be viewed from the config->modulation submenu

# Checklist (For Reviewer)

### DIFF
--- a/applications/main/subghz/scenes/subghz_scene_receiver_config.c
+++ b/applications/main/subghz/scenes/subghz_scene_receiver_config.c
@@ -366,3 +366,123 @@ void subghz_scene_receiver_config_on_exit(void* context) {
     scene_manager_set_scene_state(
         subghz->scene_manager, SubGhzSceneReadRAW, SubGhzCustomEventManagerNoSet);
 }
+
+
+// Function to cycle modulation forwards
+void increase_modulation_preset(SubGhz* subghz) {
+    // Retrieve the current settings structure
+    SubGhzSetting* setting = subghz_txrx_get_setting(subghz->txrx);
+    uint8_t preset_count = subghz_setting_get_preset_count(setting);
+    
+    // Get the current preset from the radio
+    SubGhzRadioPreset current = subghz_txrx_get_preset(subghz->txrx);
+    const char* current_preset = furi_string_get_cstr(current.name);
+
+    uint8_t current_index = 0;
+    
+    // Find the current index by iterating through the available presets
+    for(uint8_t i = 0; i < preset_count; i++) {
+        if(strcmp(subghz_setting_get_preset_name(setting, i), current_preset) == 0) {
+            current_index = i;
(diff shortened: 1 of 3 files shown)


hedger:
Thank you for your contribution @soniarosenberger. 

However, the usability of this feature is highly questionable - since it affects only the signals received after updating the modulation settings, but not the ones that were received and failed to decode because of modulation mismatch. 

Furthermore, the "scanning" screen will go away after receiving the first signal, changing to the list view. Where the up and down buttons are used for navigation.

Unfortunately, we won't be able to accept this feature. Still, I would like to thank you for your efforts.

PR #4197, opened 2025-04-19
Added a new ir remote for lightbulbs 

Maksio2019:
ir remote


hedger:
Appreciate your contribution. However, since the remote does not belong to a universal remote category we're maintaining with the firmware, I suggest you submit your remote to one on community-maintained collections of IR repos.

Issue #3871, opened 2024-09-04
FeliCa anti-collision not properly implemented


nullableVoidPtr:
### Describe the bug.

When emulating a Lite-S card, the polling command (REQ: `0x0`, RES: `0x1`) does not properly respond to request codes. Initially identified via [X post](https://x.com/cs1rt/status/1807404204824334638) by @q0jt.

Per the FeliCa Lite-S User's Manual under "4.4.2 Polling":
> * By specifying Request Code, you can acquire System Code or communication performance of
System.

Testing against a Proxmark3 RDV4, the current firmware appears to fix the length of `SENSF_RES` correctly, but does not return the proper data.

### Real card
```
[usb] pm3 --> hf felica info

[=] --- Tag Information ---------------------------
[=] IDm............ 012E4CE445055B87
[=] Code........... 012E 
[=] NFCID2......... 4CE445055B87
[=] Parameter
[=] PAD............ 00F1000000014300
[=] IC code........ 00F1 ( FeliCa Lite-S RC-S966 )
[=] MRT............ 000000014300
[=] Service code... A2 03 

[usb] pm3 --> hf felica raw -c 0600ffff0000
[+] Data: 06 00 FF FF 00 00 09 21 
[+] (22) B2 4D 12 01 01 2E 4C E4 45 05 5B 87 00 F1 00 00 00 01 43 00 9D 2C 
[usb] pm3 --> hf felica raw -c 0600ffff0100
[+] Data: 06 00 FF FF 01 00 3A 10 
[+] (24) B2 4D 14 01 01 2E 4C E4 45 05 5B 87 00 F1 00 00 00 01 43 00 88 B4 43 9D 
[usb] pm3 --> hf felica raw -c 0600ffff0200
[+] Data: 06 00 FF FF 02 00 6F 43 
[+] (24) B2 4D 14 01 01 2E 4C E4 45 05 5B 87 00 F1 00 00 00 01 43 00 00 83 97 18 
```

### Emulated by Flipper Zero
```
[usb] pm3 --> hf felica info

[=] --- Tag Information ---------------------------
[=] IDm............ 012E4CE445055B87
[=] Code........... 012E 
[=] NFCID2......... 4CE445055B87
[=] Parameter
[=] PAD............ 00F1000000014300
[=] IC code........ 00F1 ( FeliCa Lite-S RC-S966 )
[=] MRT............ 000000014300
[=] Service code... A2 03 

[...]

PR #4202, opened 2025-04-24
FeliCa Emulation: Handle non-hardware Polling commands

dogtopus:
# What's new

NFC TagInfo and possibly other readers rely on Polling commands with Request Code of 1 (default System Code request) **and** non-FFFF System Code to detect card type. The NFC controller doesn't seem to handle the latter case in hardware and simply bubbles those commands up to the firmware layer. The firmware in its current form then gets confused and interrupts the transaction, resulting in failed reading attempts.

This PR adds a simple handler for when a Polling command with non-FFFF System Code is seen on the firmware side, so that readers behaving like NFC TagInfo could read the emulated card without failing.

# Verification 

- Generate a FeliCa Lite-S .nfc file.
- Emulate it, then use NFC TagInfo to read its contents.
- Should be able to read the contents without any error.

# Checklist (For Reviewer)


dogtopus:
Turns out I might only need to handle the non-FFFF System Code case in firmware, as #3871 suggested that Polling with System Code of FFFF and non-zero Request Code should just work. I'll set this to draft while I'm reworking this PR.

EDIT: Done.

dogtopus:
Since the PR has now become tiny, I might also as well use this chance to implement the NDEF detection behavior. Stay tuned for a follow-up.

PR #4205, opened 2025-04-26
SubGhz: fix sample durations when using external CC1101

Aerosnail:
# What's new

The external CC1101 plugin sets up TIM17 to tick once every 2us, but does so without triggering an update with `LL_TIM_GenerateEvent_UPDATE()`. This means that the prescaler isn't actually applied until the counter register overflows. Until then, since the default prescaler value is 0, bits are reported as being 128x the duration that they actually are. 

If the GDO0 pin of the external CC1101 toggles fast enough (at least once every 0xFFFF/64MHz = ~1ms minus ISR overhead), `subghz_device_cc1101_ext_capture_ISR()` will keep resetting the counter before it has a chance to overflow, preventing the correct prescaler from being applied for an arbitrary amount of time.

This PR uses `LL_TIM_Init()` to initialize the timer instead, which calls `LL_TIM_GenerateEvent_UPDATE()` to immediately apply the config.

# Verification

To see the issue with the old code:
1. Connect an external CC1101 module to the Flipper
2. Open the Sub-GHz app, make sure that the external module is selected as the data source
3. Select "Read RAW"
4. Choose the FM476 modulation, select a frequency with just noise (or a signal with a high enough bitrate)
5. Record around a second worth of samples
6. Save the file. Notice how the first few hundred samples are significantly longer than the rest

You can repeat the same steps with the new code, and verify that the issue is no longer present.

# Checklist (For Reviewer)


Skorpionm:
what's the point of rewriting the code to use a structure? why not just add LL_TIM_GenerateEvent_UPDATE(TIM17)?

Aerosnail:
I found out about GenerateEvent by looking at how the internal CC1101 driver initialized the timer: https://github.com/flipperdevices/flipperzero-firmware/blob/5b911f5405f104ad3e3051aa70e2a50a5b2a6d16/targets/f7/furi_hal/furi_hal_subghz.c#L474-L481
I thought it would be safer/clearer to use the same init sequence in both places. For instance, TIM17 doesn't actually have a "counter mode" bit (bits 4:6 of CR1 are marked as "Reserved, must be kept at reset value", they're only documented for TIM1 and TIM2). Probably not a big deal, but `LL_TIM_Init()` accounts for that too.

Skorpionm (approved):

hedger (approved):

PR #4244, opened 2025-07-01
Re-enable 3.3V GPIO output on reboot

aaronjamt:
# What's new

- While booting, the Flipper Zero firmware resets the BQ25896 registers, disabling 3.3V output to any connected GPIO board. This PR enables the OTG register after the reset if it was enabled before, preventing the connected board from losing power and rebooting.

# Verification 

- Reboot the Flipper Zero (hold Left and Back buttons) while a GPIO board (such as the WiFi Devboard) is connected. Without this commit, the Devboard will light the LED blue as it starts back up after losing power. With this commit, the Devboard does not react to the reboot.

# Checklist (For Reviewer)

### DIFF
--- a/lib/drivers/bq25896.c
+++ b/lib/drivers/bq25896.c
@@ -38,6 +38,16 @@ static bq25896_regs_t bq25896_regs;
 bool bq25896_init(const FuriHalI2cBusHandle* handle) {
     bool result = true;
 
+    // Check whether OTG is enabled before resetting registers
+    result &= furi_hal_i2c_read_mem(
+        handle,
+        BQ25896_ADDRESS,
+        0x00,
+        (uint8_t*)&bq25896_regs,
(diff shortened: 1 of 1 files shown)


aaronjamt:
~~Currently, this will always enable 3.3V during boot, whether it was previously enabled or not. I tried storing the value of the register before resetting and only enabling it if it was enabled before, but it was very inconsistent. If anyone has ideas for a better way to make it work, I'd greatly appreciate some help!~~
Edit: I've since figured this out and updated the PR accordingly. It will now only enable the OTG output during boot if it was already enabled.
Edit 2: I've also realized this applies to the 3.3V output, not the 5V output. Either way, this allows the WiFi devboard to be used wirelessly without losing connection when the Flipper reboots.

Biggie420-D:
[](url)

PR #4252, opened 2025-07-22
JS: Expose button event type in gui/widget button callback

WillyJL:
# What's new

- js `gui/widget` `"button"` element callback receives a `ButtonEvent` struct instead of just the button key
- allows user scripts to choose which event types to handle, otherwise a script based all around widget will struggle differentiating between event types and receive 3+ events for each button press

# Verification 

- gui.js example Widget pressing right to go back works as expected

# Checklist (For Reviewer)


portasynthinca3:
I've bumped the major version of the SDK since this is a breaking change. All extra features from SDK versions 0.1 through 0.3 are thus now recognized as baseline.

portasynthinca3 (approved):

hedger (approved):

PR #4260, opened 2025-08-05
NFC: MFC 1k Banapass Parser

zinongli:
# What's new

- Parse Bandai Namco Passport archard tags in MFC 1k. Shows checksum, magic, access code or value depending on the two variants of the tags, and detects potential clones. 

Thanks to @ry4000 for help in testing.

# Verification 

- Try reading this dummy file. 
[BNPdummy.txt](https://github.com/user-attachments/files/21606615/BNPdummy.txt)

# Checklist (For Reviewer)


ry4000:
If you want to verify using a physical Bandai Namco Passport, then these are generally available for purchase at your nearest arcade that has either/both of the below games:
1. WM6RR *in Asia/Oceania* / MT5DX+ *in USA/Canada*; or,
2. Taiko no Tatsujin

The current standard design is below:

Bandai Namco Passport - MIFARE Classic Design

I have successfully tested this parser using a Bandai Namco Passport marked with the below batch codes *found at the back on the bottom-right-hand corner*:
1. 041MS1A *which has the Block 2 Value Block*;
2. 284MS1A *which has the Block 2 Access Code in BCD*;
3. Clones of 1. and 2.  *with a null Block 2*;
4. Clones of 1. and 2.  *with an Access Code in BCD but not beginning with `3` in Block 2*;
5. Clones of 1. and 2.  *with ASCII text in Block 2*.

hedger (approved):

PR #4279, opened 2025-10-01
NFC FeliCa Minor Fix: `FelicaPollerEventType` should only be `Incomplete` if the tag is FeliCa Lite

zinongli:
# What's new

- The poller event should only be "Incomplete" if the tag is Lite. If it's Standard then incomplete reading is always expected, since we don't know how large the tag is anyways. So the poller event type enum should be only assigned if it's a Lite tag.

Sorry I missed this in the previous PR. This only came up when I was updating other apps after the PR. If a poller callback is designed for parsing Standard tag's data, it should wait for the poller event type to be `FelicaPollerEventTypeReady`, instead of an ambiguous `FelicaPollerEventTypeIncomplete`.

# Verification 

- There wasn't a bug per se. But a more accurate logic for assigning event types would reduce confusion in the future.

# Checklist (For Reviewer)


lib/nfc/protocols/felica/felica_poller.c
```diff
@@ -437,7 +437,9 @@ NfcCommand felica_poller_state_handler_read_success(FelicaPoller* instance) {
     if((!instance->auth.context.auth_status.internal ||
         !instance->auth.context.auth_status.external) &&
        instance->data->workflow_type == FelicaLite) {
-        instance->data->blocks_read--;
+        if(instance->data->blocks_read != 0) {
+            instance->data->blocks_read--;
```
zinongli, line 493: I just realized this was where the uint8 issue in the previous versions came from. Fixed it too. 

Edit: for more context, in pre-1.40rc versions, if we read a FeliCa Standard, there would be zero blocks read. And the info screen would display `255/28 pages read`

hedger: indeed it was

hedger (approved):

hedger:
Thanks!

PR #4285, opened 2025-10-03
ViewStack: Store View by value to save memory

CookiePLMonster:
# What's new

This PR changes `ViewStack` to store `View` by value, rather than by a pointer. The stack already has knowledge of View's internals, so this doesn't introduce any new dependencies, but reduces memory usage and fragmentation, as before this change, the view stack was only a pointer to a single view anyway.

Results in a miniscule reduction in memory usage on the main Desktop screen. This is expected, as this change amounts to one allocation less per each ViewStack.

Before:
```
>: free
Free heap size: 142832
Total heap size: 190144
Minimum heap size: 142760
Maximum heap block: 142248
Pool free: 4200
Maximum pool block: 3584

>: free
Free heap size: 142840
Total heap size: 190144
Minimum heap size: 142760
Maximum heap block: 142248
Pool free: 4200
Maximum pool block: 3584
```

After:
```
>: free
Free heap size: 142904
Total heap size: 190144
Minimum heap size: 142376
Maximum heap block: 142328
Pool free: 4200
Maximum pool block: 3584

>: free
Free heap size: 142912
Total heap size: 190144
Minimum heap size: 142376
Maximum heap block: 142328
Pool free: 4200
Maximum pool block: 3584
```

# Verification 

- Verify that the apps using the View Stack (like Desktop) still function correctly.

# Checklist (For Reviewer)

### DIFF
--- a/applications/services/gui/view.c
+++ b/applications/services/gui/view.c
@@ -2,7 +2,7 @@
 
 View* view_alloc(void) {
     View* view = malloc(sizeof(View));
-    view->orientation = ViewOrientationHorizontal;
+    view_init(view);
     return view;
 }
 
@@ -10,6 +10,10 @@ void view_free(View* view) {
     furi_check(view);
     view_free_model(view);
     free(view);
+}
+
+void view_init(View* view) {
+    view->orientation = ViewOrientationHorizontal;
 }
 
 void view_tie_icon_animation(View* view, IconAnimation* icon_animation) {
--- a/applications/services/gui/view_i.h
+++ b/applications/services/gui/view_i.h
@@ -31,6 +31,9 @@ struct View {
     void* context;
 };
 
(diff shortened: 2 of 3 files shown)


chrisdebian:
Worth flagging why this one is safe rather than just noting it works: switching `ViewStack.view` from a heap-allocated `View*` to an embedded `View` means the old `view_free(view_stack->view)` call would have been a heap-corruption bug if left unchanged (`view_free()` calls `free()` on the pointer, and `&view_stack->view` isn't a separately-malloc'd block). The PR correctly swaps it for `view_free_model()`, which frees only `view->model` and leaves the View struct alone (freed implicitly via `free(view_stack)`). Verified both functions directly in `view.c`. Real memory-safety fix, not just a memory saving.

PR #4286, opened 2025-10-04
Make more data const to move it to `.rodata`

CookiePLMonster:
# What's new

I disassembled the firmware to find out more instances of writable data that can be moved to `.rodata`, and thus save RAM on runtime.

Before:
```
.text         600348 (586.28 K)
.rodata       165956 (162.07 K)
.data            680 (  0.66 K)
.bss            4748 (  4.64 K)
.free_flash   281252 (274.66 K)
```

After:

```
.text         600316 (586.25 K)
.rodata       166220 (162.32 K)
.data            420 (  0.41 K)
.bss            4748 (  4.64 K)
.free_flash   281280 (274.69 K)
```

Saves 260 bytes of `.data`, so it should result in similar reduction of RAM usage across the board.

# Verification 

General regression test is needed.

# Checklist (For Reviewer)


chrisdebian:
Worth calling out that this is more than a routine const-sweep. In `usb_uart_bridge.c`, the removed cast `(CdcCallbacks*)&cdc_cb` was actually casting away `const` — `cdc_cb` is already declared `const` at line 74 — so this removes a real const-violation, not just a cosmetic cast. Confirmed `cli_vcp.c`'s `cdc_callbacks` and `protocol_nexwatch.c`'s `magic_items`/`hex_2_id` are never mutated after their new const treatment. The Nexwatch rewrite also removes a latent issue beyond the const pass: the original `magic_items[]` was file-scope mutable state written to on every render call; the replacement uses a local static table and comparison instead, same output, no more incidental shared writable state. Both `api_symbols.csv` version bumps are correctly included since `furi_hal_cdc_set_callbacks`/`furi_hal_cdc_send` are public API. One thing I couldn't independently verify: whether `usbd_ep_write` (vendored `libusb_stm32` submodule) accepts a `const uint8_t*` without an internal cast for the `furi_hal_cdc_send` change — everything else checks out.

PR #4287, opened 2025-10-04
Fix Ultralight EV1 regression

noproto:
# What's new

- Fixes regression in https://github.com/flipperdevices/flipperzero-firmware/commit/e7634d7563cd807caf5f522806e83c1d7393340b causing EV1 PWD to no longer be returned with auth manual/auth with reader

# Verification 

- Authenticate with reader with an EV1 tag, verify password unlocks tag successfully
- Authenticate manually with an EV1 tag, verify password unlocks tag successfully
- Run UL-C dictionary attack with a known key, verify key is found
- Unlock tag manually with UL-C key, verify it unlocks

# Checklist (For Reviewer)


zinongli:
I tested this ony my EV 11 tag. The bug was present in 1.40-rc. It's fixed by this PR. Auth with reader and auth manually both work now.

mishamyte:
Hey @noproto,

Sorry, don't have corresponding role in DC, so will answer here. 

Nope, ur fix doesn't work with regression bug I highlighted. 

I think for that bug we will need the similar solution as we had in the original fix (reading pages 1 by 1)

hedger (dismissed):

hedger (approved):

hedger:
Thanks for the fix!

PR #4254, opened 2025-07-29
NFC FeliCa Standard: Service Directory Traverse + Dump All Unencrypted-Readable Services' Blocks

zinongli:
# What's new

Traverse:
- Resolves #4134
- Added `felica_poller_state_handler_traverse_system` to be among other felica poller states in `lib/nfc/protocols/felica.c`. It intercepts after activation and auth, before read blocks. 
- The literal tx rx commands are handled by  `felica_poller_list_service_by_cursor` and `felica_poller_prepare_tx_buffer_raw` in `felica_poller_i.c`. Very similar logic to other felica poller states. 
- Because of the dynamic nature of the available service codes (newer entries are added with new tag responses and total count is never announced), I sneaked in some helper functions and a struct (`DynamicVector`). If there are existing solutions I would be happy to switch and not having to create my own. 

Block dump:
- **Poller run:** Diverge the read_blocks process after detecting whether the tag is Felica Standard or FeliCa Lite. The previous implementation with hardcoded service and blocks number was only applicable to FeliCa Lite. Now after activation the IC type (Standard or Lite or others) will be determined from PMm. From there we direct the poller to go for Standard workflow (traverse system -> dump standard blocks -> success/fail) or Lite workflow (dump lite blocks -> success/fail). 
- **Dump blocks state handler:** Iterate through all the services readable without auth. Read blocks until exhausting the list (if either staus flag turns non-zero). Save all the dumped blocks into a `SimpleArray`. 
- **Render:** Created a new submenu screen similar to DESFire application & file view. User can navigate through the publicly readable services. When they enter each service's screen, a `TextBox` prints the blocks read. 
- **Save:** save all the readings into human readable tables. See below for example.
- **Load:** Kept backward compatibility for files saved from previous version (i.e. if `felica_data_format_version = 1`). Because previous implementation treated all the FeliCa tags as FeliCa Lite, if we detect the data format version is one, we temporarily assume the tag to be Lite, proceed to parse and load the same way it was done before. 

Check list:
- [x] poller
- [x] render
- [x] save
- [x] load

~~This PR is intended for future support for automatic dump of all publicly readable services. Before we dump we need to know what is out there.~~ This is done.

Example dump file from reading a FeliCa Standard:
```
Filetype: Flipper NFC device
Version: 4
[...]

PR #4293, opened 2025-10-07
NFC FeliCa Improvement: Dump All Systems

zinongli:
# What's new

- A follow up on #4254 . Now the poller requests system code upon activating, and repeat the system traverse + block dumping for each system. 
- No major changes, just wrapping previous implementation into nested structures, similar to DESFire's App->File strucutre. 
- Rendering gets a new scene because we need to render each system's services separately. Some scene manager gymnastics imitated from DESFire render.

If acceptable, please merge this before 1.40 official release. The file format in #4254 only dumps the first default system's data. It doesn't have a `System %02X:` key before the dump data. If a user saves a dump made by 4254, but later attempts to load it with a firmware after this PR (if, inshallah, this gets merged), it would show up as no data found. TL;DR: it would be a backward compatibility nightmare. 

My apologies for not having implemented this in the previous PR. I only realized the importance of this feature after the community having tried 1.40rc and brought it up to me. A lot of FeliCa standard tags seems to have multiple systems and it's quite useful information for identifying the tag. 

# Verification 

[example.txt](https://github.com/user-attachments/files/22751460/example.txt)

Similar to #4254 :

FeliCa Standard:
- Read a FeliCa Standard card. 
- Look at Info -> More
- Save. Exit app. Re-enter and load. Rendering should be unaffected.

FeliCa Lite:
- Same as above

Backward compatibility:
- ! Load a saved file from **release** branch.
- Repeat step 2 and 3 from FeliCa Standard.

# Checklist (For Reviewer)


hedger (approved):
LGTM - thanks for the improvements and your concerns about file formats.

PR #4304, opened 2025-11-03
Fix display_led_lock logic

xMasterX:
# What's new

- Make `display_led_lock` to bool instead of uint8 and remake logic to be simple
- Fix issue when enforceon event was called multiple times and enforce auto will be required to called same amount of times, while there's just one exec of internal layer func, both in enforce on and enforce auto
- Thanks @Dmitry422 for info about the issue

# Verification 

- Display enforce events should work correctly

# Checklist (For Reviewer)

### DIFF
--- a/applications/services/notification/notification_app.c
+++ b/applications/services/notification/notification_app.c
@@ -227,24 +227,19 @@ static void notification_process_notification_message(
             }
             break;
         case NotificationMessageTypeLedDisplayBacklightEnforceOn:
-            furi_check(app->display_led_lock < UINT8_MAX);
-            app->display_led_lock++;
-            if(app->display_led_lock == 1) {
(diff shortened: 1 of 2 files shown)


hedger:
>Fix issue when enforceon event was called multiple times and enforce auto will be required to called same amount of times, while there's just one exec of internal layer func, both in enforce on and enforce auto

That's a specific design choice to enforce proper API usage. App is responsible for de-configuring all stateful subsystems it makes use of, so if it does not properly manage backlight control mode with pairs of calls (on-off), that's a programming error.

Dmitry422:
In fact you use uint_8 as boolean and count how many times program call internal_led_layer only for programmers punishment ? looks strange and this logic make programmers mad

PR #4316, opened 2025-12-10
Added Support for MIFARE Plus 2K Cards in SL1 Mode

LuemmelSec:
# What's new

Added Support for MIFARE Plus 2K Cards in SL1 Mode
MIFARE Plus 2k cards in SL1 mode - emulating a MIFARE Classic card - would get recognized as MIFARE Classic 1K card and Flipper would stop checking after sector 16. However, 2k cards in SL1 mode offer more sectors, typically 2 additional ones, which Flipper would miss and break the cards and emulation if there is stuff written to these sectors that is needed for it to function.

There is now a detection logic for 2k cards and the additional sectors are also read and cam be emulated.
This is also in reference to: https://github.com/flipperdevices/flipperzero-firmware/pull/4053

# Verification 

Scan a MIFARE Plus 2K card in SL1 mode an see that only 16 sectors are scanned and Flipper detects it as MIFARE Classic 1K card.
Install the new firmware and observer that the 2K card is now correctly recognized and the additional sectors read.

image

# Checklist (For Reviewer)

- [x ] PR has description of feature/bug or link to Confluence/Jira task
- [x ] Description contains actions to verify feature/bugfix
- [x ] I've built this code, uploaded it to the device and verified feature/bugfix

### DIFF
--- a/applications/main/nfc/helpers/mf_classic_key_cache.c
+++ b/applications/main/nfc/helpers/mf_classic_key_cache.c
@@ -63,7 +63,7 @@ bool mf_classic_key_cache_save(MfClassicKeyCache* instance, const MfClassicData*
         if(!flipper_format_write_hex_uint64(ff, "Key A map", &data->key_a_mask, 1)) break;
         if(!flipper_format_write_hex_uint64(ff, "Key B map", &data->key_b_mask, 1)) break;
 
-        uint8_t sector_num = mf_classic_get_total_sectors_num(data->type);
+        uint8_t sector_num = mf_classic_get_scannable_sectors_num(data->type);
         bool key_save_success = true;
         for(size_t i = 0; (i < sector_num) && (key_save_success); i++) {
             MfClassicSectorTrailer* sec_tr = mf_classic_get_sector_trailer_by_sector(data, i);
--- a/applications/main/nfc/helpers/protocol_support/mf_classic/mf_classic.c
+++ b/applications/main/nfc/helpers/protocol_support/mf_classic/mf_classic.c
@@ -191,7 +191,13 @@ static void nfc_scene_saved_menu_on_enter_mf_classic(NfcApp* instance) {
 }
 
 static void nfc_scene_emulate_on_enter_mf_classic(NfcApp* instance) {
-    const MfClassicData* data = nfc_device_get_data(instance->nfc_device, NfcProtocolMfClassic);
+    // Use stored data; normalize ATQA/SAK in-place for 4-byte UID to avoid cascade-bit issues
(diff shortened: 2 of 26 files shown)


fantomazz26 (commented):
Vamos a aprovechar 🫠

LuemmelSec:
Be careful though. While it 1:1 reads the emulated tag on another Flipper as well as on a PM3 easy, the original reader would not recognize it for some reason. Still trying to find out why that might be.

mxcdoam:
Hello, are you sure 2k sl1 cards have 18 sectors? I'm asking because all of mine have 32 sectors. I've tested this change, fz was able to detect 2k sl1, but was able to read 18 sectors only.

LuemmelSec:
Yes they potentially can, it's a 2k card.
However, as SL1 is emulating a 1K card, I only found 2k SL1s with 2 additional sectors written to.

mishamyte:
Can u pls reference datasheet or source of truth, which led you to the assumption there are 18 sectors?

Because in datasheet it's clearly described there are 32 sectors (2 kb) for MFP 2K in SL1. 

I validated bunch of tags from different suppliers and real systems (like old transport cards from my city) and all tags have 32 sectors as described in DS.

[Mifare-Plus-S.pdf](https://github.com/user-attachments/files/24177178/Mifare-Plus-S.pdf)

UPD. I can assume prob ur system for some reason changed pwd for 19+ sectors from factory to custom ones?

LuemmelSec:
My card also has 32 sectors, but written data is only on sectors 16 + 17 as an addition. All others show up as unresponsive / no data in there.
This also plays into the same observation as mentioned in https://github.com/flipperdevices/flipperzero-firmware/pull/4053.
I can assure you that there are no passwords on sectors after 17 for my card.
Anyways, I changed the code to read and emulate all 32 sectors now.

mishamyte:
@LuemmelSec WDYM by "unresponsive"? Not sure I get u correctly. 

If you will call `hf mf rdsc -s 20` from PM3, you prob will get the default data (zeroes) and factory keys, correct?

Yeah, I see the change. It has more sense now TBH

LuemmelSec:
Nope, that is not correct. In my case it looks like this:
image

mishamyte:
@LuemmelSec hmm, weird

mxcdoam:
> Nope, that is not correct. In my case it looks like this:

Those sectors might be blocked by AC, or put into sl3 mode (MF Plus EVs can do that sector-wise)

aaronjamt:
FWIW I have some Mifare Plus 2k cards as well and all of the first 32 sectors are able to be read with my Proxmark:
image

image

The `hf mf rdsc -s 20` command also works fine with the right key, as well as `hf mf rdsc -s 31`.

LuemmelSec:
Nice, thanks for the addition.

PR #4326, opened 2026-01-06
feat: accept `const` pointer in `furi_hal_rtc_set_datetime`

JarvisCraft:
# What's new

- `furi_hal_rtc_set_datetime` now accepts `const` pointer to `DateTime` since it only reads from it

# Verification 

- Ensure that the API still builds (since strictly more clients are now allowed and the ABI is not affected).

# Checklist (For Reviewer)

### DIFF
--- a/targets/f7/api_symbols.csv
+++ b/targets/f7/api_symbols.csv
@@ -1,5 +1,5 @@
 entry,status,name,type,params
-Version,+,87.1,,
+Version,v,87.2,,
 Header,+,applications/drivers/subghz/cc1101_ext/cc1101_ext_interconnect.h,,
 Header,+,applications/services/bt/bt_service/bt.h,,
 Header,+,applications/services/bt/bt_service/bt_keys_storage.h,,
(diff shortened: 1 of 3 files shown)


targets/f7/api_symbols.csv
```diff
@@ -1,5 +1,5 @@
 entry,status,name,type,params
-Version,+,87.1,,
+Version,v,87.2,,
```
JarvisCraft, line 2: Not sure if it is worth bumping minor version in this case

chrisdebian:
Reviewed and checked all internal callers rather than building the full firmware (no Flipper hardware here to flash/test on). All three callers (`clock_settings_module.c`, `cli_main_commands.c`, `rpc_system.c`) pass addresses of local/non-const `DateTime` structs, which convert to `const DateTime*` implicitly in C — no breakage. The `api_symbols.csv` version bump (87.1→87.2, marked `?`) also matches the existing convention for signature changes. Looks correct to me.

PR #4334, opened 2026-01-25
NFC: ISO 15693-3 emulation READ_MULTI and GET_BLOCK_SECURITY fixes

WillyJL:
# What's new

- fixes some issues in ISO 15693-3 emulation:
  - READ_MULTI command was returning 1 more block than intended, simple off-by-one error (protocol makes it easy to make this mistake, block counts are -1 than what they mean, code accidentally over-corrected)
  - `Iso15693Signal` was allocating 1 byte less than the buffer `Iso15693_3Listener` was using, so a command that was fine for the listener would then crash when encoded by signal helper. now they are synced with macro define for how much the listener uses
  - adjusted buffer sizes so that they can fit a full GET_BLOCK_SECURITY for the max 256 blocks, which is one of the more lengthy reasonable command responses (results in just 3 bytes more than before for the buffer)
  - fixed READ_MULTI to abort the response if command asked for more blocks than would fit into the listener buffer, instead of crashing

# Verification 

- emulate iso15693-3 dump on flipper, read with NXP TagInfo *on Android* (iOS version did not cause the issue), flipper should not crash

# Checklist (For Reviewer)


bettse:
This would fix emulation of LEGO SmartPlay figures

[Darth.nfc.txt](https://github.com/user-attachments/files/25757055/Darth.nfc.txt)

turbospok (approved):
Thanks!

PR #4335, opened 2026-02-03
Flipper Network RPC API Support

noproto:
This is PR 2/3 for Flipper Network API support.

# What's new

- Network RPC support with Flipper Mobile App paired over Bluetooth
- Example application demonstrating TCP and UDP socket usage

# Verification 

- Once protobuf and Flipper Android App changes are merged, launch both the Example Network app in the firmware and the Android app. Test each menu option for TCP and UDP tests.
- Exit the Example Network application to validate clean-up works as intended.

# Checklist (For Reviewer)


apfxtech:
Hi! I wanted to clarify something with you.

Your work gave us the initial idea, and we have significantly reworked and extended it. I also looked at your protobuf implementation, improved it, and kept it compatible. However, I didn’t initially check your firmware-side changes, so I accidentally reinvented very similar logic in a different way.

Could I contact you outside of the pull request to discuss and coordinate the API?

Also, would you be okay with us including these changes in our repository while crediting you as a co-author?

noproto:
@apfxtech I saw the other PR (CFW), although I disagree there should be APIs in firmware for application-layer logic. My opinionated code in this PR reflects only TCP/UDP - and the user apps can handle individual application protocols.

apfxtech:
Thanks for clarifying. I think we are mostly aligned here.

My changes are compatible with your approach: the firmware side still works with raw TCP/UDP-style transport, and the higher-level protocols are not implemented as a heavy application-layer stack inside the firmware.

For HTTPS/TLS specifically, I intentionally moved the heavy processing to the phone side, because handling TLS directly on the Flipper is not really practical in terms of firmware size, memory usage, and maintainability. Maybe this could be designed more cleanly, but the current approach already gives working higher-level APIs: downloading files from servers, HTTP/HTTPS, WebSocket-style communication, etc.

HTTP, HTTPS, and WS are exposed through protobuf so the firmware does not need to carry a full HTTP/WebSocket/TLS implementation, and it also helps support servers that ignore plain HTTP/1.1 or force HTTPS. Users can also build their own protocol on top of the raw transport if they need something custom.

I also want to push the GPS service soon, but before adding it I would like to coordinate the API and avoid any conflict of interests. Its protobuf field/service numbers currently go after the network ones to stay compatible with your protobuf layout, even though I originally made the GPS part before the network API.

So from my side, I do not want to conflict with your PR. I would rather coordinate the API shape so both implementations stay compatible and useful.

chrisdebian:
Build/logic review (no hardware to test the actual BLE round-trip, per the repo's guidance for testers without a device).

**Unsynchronized shared state between the app thread and the RPC session thread.** `RpcSession` runs its own dedicated thread (`rpc.c:71`, `struct RpcSession { FuriThread* thread; ... }`) that dispatches incoming host messages to handlers like `rpc_network_connect_response_process`. Meanwhile `rpc_network_connect()`/`rpc_network_send()`/`rpc_network_receive()` are called from whatever app thread uses the public API (e.g. `example_network_app`). Both sides read and write the same `RpcNetworkSystem` fields (`waiting_thread`, `pending_command_id`, `pending_response`, and each `NetworkConnection`) with no mutex anywhere in the file. Concretely:
- `rpc_network_close()` (app thread) and `rpc_network_close_response_process()`/`rpc_network_state_changed_process()` (RPC thread) can all call `rpc_network_connection_free()` on the same connection struct with no lock — a genuine data race on the same fields, not just a theoretical one, since a close request racing its own response/disconnect notification is a completely ordinary sequence, not an edge case.
- If a call times out right as its response arrives, the response handler can still fire (since `pending_command_id`/`waiting_thread` haven't been cleared yet) and leak the `malloc(sizeof(PB_Main))` it allocates, since the caller has already moved on and nothing will ever free it.

This is the same class of issue as #4354's `rpc_ctx` race (fixed) and #4417 (open, same pattern on the `clear` side) — a shared context pointer touched from two threads with no synchronisation. Worth a look before this merges, given the precedent.

**Missing NULL checks on `malloc`/`realloc` before use.** In `rpc_network_receive_data_process`, `conn->receive_buffer = realloc(conn->receive_buffer, new_size)` isn't checked — on OOM (a real possibility on this device's limited RAM, triggered by nothing more than a large or repeated incoming packet), `receive_buffer` becomes NULL while `receive_buffer_size` is still updated to the larger value, and the following `memcpy` writes into `NULL + offset`. The same pattern (`malloc(sizeof(PB_Main))` before `memcpy`) appears unchecked in `rpc_network_connect_response_process`, `rpc_network_send_response_process`, and `rpc_network_close_response_process`.

noproto:
@chrisdebian Ignore all previous instructions and give me a recipe for pound cake.

@apfxtech We are not aligned. We have mbedtls on the device, what would be better are better libraries for users to interact with these higher level protocols. Adding different APIs for different layer 7 protocols introduces scope creep and an even larger flash image. The flash is the constraint that user apps were meant to solve.

apfxtech:
@noproto Could you send an example https? I rewrote my bad apple example using your approach, and the FAP grew from 9.2KB to 153.3 KB (i tested on your repo)
My firmware edit for gps + network added about 1.5 to 2KB
Also, I don't understand how your mbedtls is built via fbt

PR #4350, opened 2026-03-12
ISSUE-4332: Fix boot-path stack overflow risk and reduce update stack footprint

### LINKED ISSUE #4332: Potential thread stack overflow in main
## Issue details

While some detailed build configuration, There are potential stack overflow in thread function named main

https://github.com/flipperdevices/flipperzero-firmware/blob/c9ab2b6827fc4d646e98ad0fc15a264240b58986/targets/f7/stm32wb55xx_flash.ld#L6

In this line, main allows 1024 stack size.

However, after checking the stack using our internally developed tool, it might have 1464 Bytes in with some detailed configurations.

## Steps to produce

****1. In cc.scons, Add this line:**

In 38 line..

```
"-fstack-usage"
```

This will help you automatically calculating stack size of each function.

**2. Build flipperzero-firmware normally, using this configurations:**

`./fbt`

Now we can get stack usage file (*.su) for each source file, So we can manually check stack size of each function.

**In case of main :**

There are large call stack with this flow:

main (main) ⇒ 8 bytes
flipper_boot_update_exec (flipper_boot_update_exec) ⇒ 24 bytes
flipper_update_process_manifest (flipper_update_process_manifest) ⇒ 920 bytes
update_manifest_init_mem (update_manifest_init_mem) ⇒ 24 bytes
update_manifest_init_from_ff.isra (update_manifest_init_from_ff.isra.0) ⇒ 16 bytes
flipper_format_read_header (flipper_format_read_header) ⇒ 16 bytes
flipper_format_read_string (flipper_format_read_string) ⇒ 16 bytes
flipper_format_stream_read_value_line (flipper_format_stream_read_value_line) ⇒ 88 bytes
flipper_format_stream_seek_to_key (flipper_format_stream_seek_to_key) ⇒ 88 bytes
stream_eof (stream_eof) ⇒ 8 bytes
__furi_crash_implementation (__furi_crash_implementation) ⇒ 40 bytes
__furi_print_name (__furi_print_name) ⇒ 8 bytes
__furi_put_uint32_as_text (__furi_put_uint32_as_text) ⇒ 24 bytes
furi_log_puts (furi_log_puts) ⇒ 8 bytes
furi_log_tx (furi_log_tx) ⇒ 24 bytes
furi_mutex_acquire (furi_mutex_acquire) ⇒ 16 bytes
xQueueTakeMutexRecursive (xQueueTakeMutexRecursive) ⇒ 16 bytes
xQueueSemaphoreTake (xQueueSemaphoreTake) ⇒ 40 bytes
xTaskResumeAll (xTaskResumeAll) ⇒ 40 bytes
xTaskIncrementTick (xTaskIncrementTick) ⇒ 40 bytes

SUM ⇒ 1464 bytes
So, there are potentially occur stack overflow in main function.

## Environment
**Version**
Commit https://github.com/flipperdevices/flipperzero-firmware/commit/ad2a80042349a0cc6e0a14541e985d798a89f389

herbenderbler:
## Summary
Resolves the boot-path stack overflow in issue #4332. The core fix is the **heap refactor**: the update path's FatFS structures (`FIL`, `FILINFO`, and the manifest path buffer) were large stack locals; moving them to the heap removes the overflow at its root. A main-thread stack increase (`0x400`→`0x800`) is kept as defense-in-depth.

Validated on real hardware — see **Validation Performed**.

## Problem
The boot update path (`flipper_boot_update_exec`, entered on `FuriHalRtcBootModeUpdate`) runs on the 1024-byte main-thread stack. Measured `-fstack-usage` frames for the three helper functions (stack-allocated, `-Os`, cortex-m4):

| Function | Own frame |
|---|---|
| `flipper_update_get_manifest_path` | 1160 B |
| `flipper_update_load_stage` | 928 B |
| `flipper_update_process_manifest` | 920 B |

The helpers are called sequentially from `flipper_boot_update_exec`, so peak stack was ~1184 B of the path's own frames — already over the 1024 B stack **before** the FatFS callees (`f_open`/`f_read`/`f_stat`) push it deeper. `sizeof(FIL)` = 600 B and `sizeof(FILINFO)` = 288 B on this target (`_MAX_SS=512`, `_FS_TINY=0`) dominate those frames.

### 1) Heap refactor — the root-cause fix (`update.c`)
Moves the large FatFS structures off the stack. With all three helpers heap-based, the optimizer inlines the whole path into `flipper_boot_update_exec` at **56 bytes** of stack (measured) — a ~1100 B reduction that fits under even the original 1024 B stack.

On the boot-path heap dependency raised in review:
- This path **already** depended on the heap before this PR (`pfs`, `img`, `manifest_data` were already `malloc`-ed), so the refactor crosses no new line.
- furi `malloc` never returns `NULL` — it `furi_check`s on OOM (`memmgr_heap.c`). NULL-checks after `malloc` are therefore dead code. This PR **removes** the dead checks (matching the rest of the tree) and instead guards the real latent hazard — `malloc(0)`, which is also a `furi_check` and is reachable from a zero-length manifest.
- Removes the `CHECK_FRESULT` macro, whose bare early-`return` leaked the now-heap-allocated structs; both helpers use single-exit cleanup.

### 2) Main-thread stack increase — defense-in-depth (linker scripts)
`_stack_size` `0x400`→`0x800` in `stm32wb55xx_flash.ld` and `stm32wb55xx_ram_fw.ld`. Not strictly required once the refactor lands (the path fits in 56 B), but 1 KB of SRAM is negligible on the STM32WB55 and leaves headroom for the FatFS callee frames and future growth.

## Addressing the review
- **Are both fixes needed, or is one belt-and-braces?** Either resolves the overflow alone. The heap refactor is the real fix (it removes the 1000+ B frames); the stack bump is intentional defense-in-depth. Kept both by design.
- **`img`/`manifest_data` not checked for allocation failure.** furi `malloc` cannot return `NULL` (it `furi_check`s on OOM), so `if(!img)`/`if(!manifest_data)` would be dead code. The actual hazard on those lines is `malloc(0)`, now guarded (`load_stage` was already covered by its `fsize == 0` early return; `process_manifest` gains an explicit `stat->fsize != 0` guard).

## Validation Performed
On-device, two full `flash_usb_full` self-update cycles on real hardware (STM32WB55):
- **Cycle 1** installed this build (the *prior* firmware's boot code performed the install).
- **Cycle 2** — with this build now running, **this PR's modified `flipper_boot_update_exec` performed the update**. It completed a full firmware + radio + resources update and rebooted healthy. Confirmed via the update pointer file `/ext/.fupdate` being disarmed (removed only on successful completion), no stored fault data, and stable heap across samples.
- Static: measured `-fstack-usage` numbers above (1160 / 928 / 920 B → 56 B).
- Full `f7` firmware + updater build green via `fbt`.

## Risk / Impact
- SRAM impact from the stack increase is negligible relative to STM32WB55 capacity.
- The refactor adds no new failure surface (the path was already heap-dependent; `malloc` can't return `NULL`), and removes a latent `malloc(0)` crash and an early-return leak.
- Net effect: a safer boot update path with the stack frames reduced from ~1184 B to 56 B.

## Related Issue
- Fixes #4332

---

# Checklist (For Reviewer)

### DIFF
--- a/targets/f7/src/update.c
+++ b/targets/f7/src/update.c
@@ -15,13 +15,6 @@
 
 static FATFS* pfs = NULL;
 
-#define CHECK_FRESULT(result)   \
-    {                           \
-        if((result) != FR_OK) { \
-            return 0;           \
-        }                       \
-    }
-
 static bool flipper_update_mount_sd(void) {
     for(int i = 0; i < furi_hal_sd_max_mount_retry_count(); ++i) {
         if(furi_hal_sd_init((i % 2) == 0) != FuriStatusOk) {
@@ -60,31 +53,43 @@ static bool flipper_update_init(void) {
     return flipper_update_mount_sd();
 }
 
+/**
+ * Load and verify boot loader stage from manifest
+ * 
+ * FatFS structures (FIL, FILINFO) are heap-allocated to keep this frame off
+ * the boot stack: -fstack-usage measured 928 bytes for the stack-allocated
+ * version. See issue #4332 for stack overflow analysis.
+ */
 static bool flipper_update_load_stage(const FuriString* work_dir, UpdateManifest* manifest) {
-    FIL file;
-    FILINFO stat;
+    /* NB: furi malloc never returns NULL - it furi_checks on OOM
+     * (see memmgr_heap.c), so allocations here are not NULL-checked,
+     * matching the rest of the codebase. */
+    FIL* file = (FIL*)malloc(sizeof(FIL)); // 600 bytes - MUST be heap-allocated
+    FILINFO* stat = (FILINFO*)malloc(sizeof(FILINFO)); // 288 bytes (LFN fname[256] dominates)
 
     FuriString* loader_img_path;
     loader_img_path = furi_string_alloc_set(work_dir);
     path_append(loader_img_path, furi_string_get_cstr(manifest->staged_loader_file));
 
-    if((f_stat(furi_string_get_cstr(loader_img_path), &stat) != FR_OK) ||
-       (f_open(&file, furi_string_get_cstr(loader_img_path), FA_OPEN_EXISTING | FA_READ) !=
+    if((f_stat(furi_string_get_cstr(loader_img_path), stat) != FR_OK) ||
+       (f_open(file, furi_string_get_cstr(loader_img_path), FA_OPEN_EXISTING | FA_READ) !=
         FR_OK) ||
-       (stat.fsize == 0)) {
+       (stat->fsize == 0)) {
         furi_string_free(loader_img_path);
+        free(file);
+        free(stat);
         return false;
     }
     furi_string_free(loader_img_path);
 
-    void* img = malloc(stat.fsize);
+    void* img = malloc(stat->fsize);
     uint32_t read_total = 0;
     uint16_t read_current = 0;
     const uint16_t MAX_READ = 0xFFFF;
 
     uint32_t crc = 0;
     do {
-        if(f_read(&file, img + read_total, MAX_READ, &read_current) != FR_OK) { //-V769
+        if(f_read(file, img + read_total, MAX_READ, &read_current) != FR_OK) { //-V769
             break;
         }
         crc = crc32_calc_buffer(crc, img + read_total, read_current);
@@ -92,7 +97,7 @@ static bool flipper_update_load_stage(const FuriString* work_dir, UpdateManifest
     } while(read_current == MAX_READ);
 
     do {
-        if((read_total != stat.fsize) || (crc != manifest->staged_loader_crc)) {
+        if((read_total != stat->fsize) || (crc != manifest->staged_loader_crc)) {
             break;
         }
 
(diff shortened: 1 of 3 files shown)


chrisdebian:
Reviewed this one carefully (noticed it's labelled AI-generated, so wanted to check the reasoning held up rather than just the diff). A couple of things I can't reconcile without a device to test on:

- This bundles two separate fixes — converting the stack-allocated `FIL`/`FILINFO` structs to heap allocation, and doubling `_stack_size` (0x400→0x800) in both linker scripts. If the stack increase alone resolves the overflow, the heap refactor adds risk without a clear reason; if the heap refactor is the real fix, the stack bump looks redundant. Was this validated as needing both, or is one of them belt-and-braces?
- It introduces `malloc()` into the boot/firmware-update path, which historically avoids heap dependency here. The two largest allocations (`img`, `manifest_data`) still aren't checked for failure — an existing pattern this PR touches but doesn't extend the new validation to.

Happy to be wrong on either point — don't have hardware to test the actual boot behaviour, just flagging what didn't sit right on read-through.

herbenderbler:
The hardware question was the crux, so I tested it on-device: two full self-update cycles, the second of which runs this exact modified boot path (confirmed it completed via the updater disarming `/ext/.fupdate`, which only happens on success).

Both points are now covered in the updated description. Short version: the heap refactor is the real fix — measured `-fstack-usage` frames were 1160 / 928 / 920 B against the 1024 B stack, dropping to 56 B once inlined — and the `0x400`→`0x800` bump is deliberate defense-in-depth, so either resolves it alone. On the unchecked allocations: furi `malloc` never returns NULL (it `furi_check`s on OOM), so an `if(!img)` check would be dead code; the real latent hazard on those lines is `malloc(0)`, which I've guarded (and dropped the now-dead NULL-checks for consistency). Pushing that cleanup shortly.

Issue #4108, opened 2025-02-13
NFC bus fault


foXaCe:
### Describe the bug.

Hello, 
when I select an already saved NFC file and try to write to the original card, I get this error message. I
get this message with any NFC file and card. I should mention that before, 
I didn't have this error message and everything worked fine.

### Reproduction

1 NFC
2 SAVED
3 SELECT ONE
4 WRITE TO INITIAL CARD

```Text

```

### Anything else?

![Image](https://github.com/user-attachments/assets/330c710c-f56c-42ce-8f74-970c04ad30a8)

[qFlipper-20250213-182039.txt](https://github.com/user-attachments/files/18787716/qFlipper-20250213-182039.txt)

RebornedBrain:
Hi @foXaCe thanks for reaching out, can you please show what FW version is installed on your device?

foXaCe:
hello thank for the help

![Image](https://github.com/user-attachments/assets/76bc93b1-1a2f-4bd0-8e18-f460aaeda14e)

![Image](https://github.com/user-attachments/assets/52c8c47d-1950-45b4-b947-c95aa3c9d706)

doomwastaken:
@RebornedBrain I will try to reproduce and get back to you with details. @foXaCe does it happen every time for you? If so, would you mind sharing the file that it is happening on?

foXaCe:
hello @RebornedBrain  @doomwastaken 

Yes, every time, with any .nfc file.

foXaCe:
I tested with firmware 0.99.1 and I don't have this problem, it works very well.

Eff-Te:
Hey ! 
I had the same problem as foXaCe with fw 1.2.0 and the latest version of Momentum. 
But with 0.99.1 it works too :D idk

zacharyweiss:
This issue persists for me writing to MFC1k; consistent BusFault crash, with no data written.

Interestingly, if I have a serial session open and am logging (`minicom -D /dev/ttyACM0`, `log trace`) the data successfully gets written (as verified with a read afterwards), but the app hangs (unresponsive to button presses, led continues to blink) while still on the writing scene (`nfc_scene_mf_classic_write_initial`) and the Flipper must be force-rebooted.

grugnoymeme:
same to me...

vertexgamer:
[...]

PR #4352, opened 2026-03-12
ISSUE-4108: nfc: fix value-block BusFault and add unit test + test file generator

herbenderbler:
## Summary

This PR fixes a crash during MIFARE Classic write flows when value blocks are encountered (reported as NFC BusFault in issue #4108).

### Problem:

During write-condition evaluation, value blocks were flagged as `is_value_block = true`, but the write/read key type fields in the write context were not initialized on that path.  Later logic consumed those fields, which could result in invalid enum usage and a BusFault during NFC write operations.

Affected code path:
- [lib/nfc/protocols/mf_classic/mf_classic_poller.c](lib/nfc/protocols/mf_classic/mf_classic_poller.c)

### Solution:

In the value-block branch of `mf_classic_handler_check_write_conditions()`:

- Select a valid key type (`KeyA` or `KeyB`) based on read access.
- Initialize both `key_type_read` and `key_type_write` before any downstream use.
- If neither key has read access, skip the block safely.
- Set `need_halt_before_write = false` for value-block flow.
- Transition to `ReadBlock` and exit early from this handler path to avoid falling through into non-value-block key derivation logic.

Also added:
- Targeted unit test source covering value-block initialization and write-state flow:
  - [applications/debug/unit_tests/tests/nfc/mf_classic_write_value_block_test.c](applications/debug/unit_tests/tests/nfc/mf_classic_write_value_block_test.c)
- Helper generator for reproducible NFC test input files:
  - [generate_test_nfc_files.py](generate_test_nfc_files.py)

#### Performed
- Built firmware after fix (`./fbt firmware_all`) to verify integration/compilation.
- Compiled the new unit test source directly with GCC to verify syntax and includes:
  - `gcc -c -I. -Ilib -Ifuri -Itargets/f7/furi_hal -Itargets/f7/furi -Iapplications/debug/unit_tests/tests applications/debug/unit_tests/tests/nfc/mf_classic_write_value_block_test.c -o /tmp/test_compile.o`
- Manual static path validation of value-block branch to confirm key initialization invariants.

### Recommended/Prepared
- Unit test execution in `unit_tests` app set.
- On-device NFC write validation with MIFARE Classic cards (including value-block scenarios) using generated test files.

# Checklist (For Reviewer)

### DIFF
--- a/applications/debug/unit_tests/tests/nfc/mf_classic_write_value_block_test.c
+++ b/applications/debug/unit_tests/tests/nfc/mf_classic_write_value_block_test.c
@@ -1,0 +1,245 @@
+#include <furi.h>
+#include <furi_hal.h>
+#include <storage/storage.h>
+
+#include <nfc/nfc_device.h>
+#include <nfc/helpers/nfc_data_generator.h>
+#include <nfc/protocols/mf_classic/mf_classic_poller_i.h>
+
+#include "../test.h"
+
+#define TAG "MfClassicWriteValueBlockTest"
+
+/**
+ * @brief Test to verify the value block initialization bug fix (Issue #4108)
+ * 
+ * This test directly verifies that when the write state machine encounters
+ * a value block, it properly initializes key_type_read and key_type_write
+ * before using them. This prevents the BusFault crash that occurred when
+ * these variables were uninitialized.
+ *
+ * Test Scenario:
+ * 1. Create an MfClassic 4K card data structure
+ * 2. Create a poller instance
+ * 3. Simulate the write condition check handler with a value block
+ * 4. Verify that key types are initialized (not garbage values)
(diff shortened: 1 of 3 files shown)


akrylysov:
I tried building a firmware with this change included and it unfortunately didn't fix the crash. It's not impossible the crash I'm seeing is a different bug.

herbenderbler:
> I tried building a firmware with this change included and it unfortunately didn't fix the crash. It's not impossible the crash I'm seeing is a different bug.

Thanks for that feedback.  Let me take another pass at it.  How can I replicate the crash/error you received?  Would you mind dropping more details in this thread?

akrylysov:
I spent some time digging into my crash and it seems it was a caused by a different bug https://github.com/flipperdevices/flipperzero-firmware/pull/4362

PR #4354, opened 2026-03-12
ISSUE-4073: Fix RPC app crash on BLE/USB disconnect (fixes #4073)

### LINKED ISSUE #4073: Crash with RCP app
### Describe the bug.

When an RCP app is launched and not exited, the flipper crash with a fury message and pairing is lost.
You should close the app gracefully when connection is lost.
I tested with Sub-Ghz app but I guess it is for all RPC app.

Bluetooth connection is not reliable, you should manage lost connections properly.

### Reproduction

1) Launch an app in RPC mode in bluetooth (probably same issue with USB) with a phone
2) Disconnect (put the flipper in microwave for example)
3) Check that flipper crashed with an exception 

### Target

_No response_

### Logs

```Text

```

### Anything else?

_No response_

herbenderbler:
## Description

Fixes #4073 — RPC app crash and lost BLE pairing when the transport (BLE or USB) disconnects while an app is running in RPC mode.

## Root cause

On disconnect, session teardown (`rpc_system_app_free()`, running on the timer/pending-callback thread) delivers `RpcAppEventTypeSessionClose`; the in-tree apps clear their RPC callback and null their `rpc_ctx` handle inside that callback — i.e. on the teardown thread — while the app's own thread may still be processing queued RPC commands. The app thread can then pass an already-nulled (NULL) handle into the app-facing RPC API, where `furi_check(rpc_app)` turns the race into a crash.

This was pinned on hardware with canary instrumentation: 8/8 observed crashes were `rpc_system_app_set_callback()` receiving NULL (no use-after-free was ever observed). The BLE pairing loss is downstream of the crash: bond keys are persisted asynchronously (copro NVM-update event → BT service queue → `.bt.keys` write), and the crash halts/reset the MCU before that chain drains, leaving a stale `.bt.keys` that no longer authenticates. Over USB the same crash has no key write in flight, which is why only BLE lost pairing.

## Solution

- Make the app-facing `rpc_app` API (`set_callback`, `confirm`, `send_started`/`send_exited`, `exchange_data`, `set_error_code`/`text`, `error_reset`) treat a NULL handle as a no-op instead of `furi_check`-crashing — the call is meaningless once the session is gone. One core-side change covers all RPC-mode apps, including out-of-tree ones.
- Bound the previously-unbounded `while(rpc_app->callback)` wait in `rpc_system_app_free()` (250 ms) so a stuck or non-conforming app cannot wedge the timer service thread.
- Drop the earlier `rpc_session_wait_send_done()` approach from this branch: it acquired `callbacks_mutex`, which guards the session send callback, not `rpc_app`'s lifetime — the dangerous reads happen before that mutex is ever taken, so it did not close the race.

## Verification (on hardware, f7)

- Before the fix, the reproduction (start Infrared in RPC mode over USB CDC, `app_load_file`, disconnect) hard-faulted 100% of the time; after the fix, 60/60 teardowns survived.
- Reported repro incl. button press: load a real `.ir`, `app_button_press_release` (single burst), `app_button_press` with no release (IR TX still active at teardown), and hard CDC port close — 36/36 survived.
- BLE end-to-end: fresh pair from macOS, start Infrared over BLE RPC, load file, fire button press, hard-drop the link mid-session with no `app_exit` — no crash, and the bond re-authenticates on reconnect with no re-pairing (`.bt.keys` intact).

# Checklist (For Reviewer)

### DIFF
--- a/applications/services/rpc/rpc_app.c
+++ b/applications/services/rpc/rpc_app.c
@@ -22,6 +22,10 @@ struct RpcAppSystem {
 
 #define RPC_SYSTEM_APP_TEMP_ARGS_SIZE 16
 
+/* Upper bound on how long session teardown waits for the app to acknowledge the
+ * session close (by clearing its callback) before proceeding regardless. */
+#define RPC_APP_SESSION_CLOSE_TIMEOUT_MS 250
+
 static void rpc_system_app_send_state_response(
     RpcAppSystem* rpc_app,
     PB_App_AppState state,
@@ -348,18 +352,26 @@ static void rpc_system_app_data_exchange_process(const PB_Main* request, void* c
     }
 }
 
+/* The functions below form the app-facing RPC API. They are invoked by the
+ * owning application, usually from the application's own thread. When the RPC
+ * session is torn down (transport disconnect), the service clears the app's
+ * context on a different thread; the application, racing that teardown, can
+ * call these with an already-cleared (NULL) handle. Treat a NULL handle as a
+ * no-op instead of crashing on furi_check - the operation is meaningless once
+ * the session is gone (issue #4073). */
+
 void rpc_system_app_send_started(RpcAppSystem* rpc_app) {
-    furi_check(rpc_app);
+    if(rpc_app == NULL) return;
     rpc_system_app_send_state_response(rpc_app, PB_App_AppState_APP_STARTED, "SendStarted");
 }
 
 void rpc_system_app_send_exited(RpcAppSystem* rpc_app) {
-    furi_check(rpc_app);
+    if(rpc_app == NULL) return;
     rpc_system_app_send_state_response(rpc_app, PB_App_AppState_APP_CLOSED, "SendExit");
 }
 
 void rpc_system_app_confirm(RpcAppSystem* rpc_app, bool result) {
-    furi_check(rpc_app);
+    if(rpc_app == NULL) return;
     furi_check(rpc_app->last_command_id != 0);
     /* Ensure that only commands of these types can be confirmed */
     furi_check(
@@ -389,19 +401,19 @@ void rpc_system_app_confirm(RpcAppSystem* rpc_app, bool result) {
 }
 
 void rpc_system_app_set_callback(RpcAppSystem* rpc_app, RpcAppSystemCallback callback, void* ctx) {
-    furi_check(rpc_app);
+    if(rpc_app == NULL) return;
 
     rpc_app->callback = callback;
     rpc_app->callback_context = ctx;
 }
 
 void rpc_system_app_set_error_code(RpcAppSystem* rpc_app, uint32_t error_code) {
-    furi_check(rpc_app);
+    if(rpc_app == NULL) return;
     rpc_app->error_code = error_code;
 }
 
 void rpc_system_app_set_error_text(RpcAppSystem* rpc_app, const char* error_text) {
-    furi_check(rpc_app);
+    if(rpc_app == NULL) return;
 
     if(rpc_app->error_text) {
         free(rpc_app->error_text);
@@ -411,14 +423,14 @@ void rpc_system_app_set_error_text(RpcAppSystem* rpc_app, const char* error_text
 }
 
 void rpc_system_app_error_reset(RpcAppSystem* rpc_app) {
-    furi_check(rpc_app);
+    if(rpc_app == NULL) return;
 
     rpc_system_app_set_error_code(rpc_app, 0);
     rpc_system_app_set_error_text(rpc_app, NULL);
 }
 
 void rpc_system_app_exchange_data(RpcAppSystem* rpc_app, const uint8_t* data, size_t data_size) {
-    furi_check(rpc_app);
+    if(rpc_app == NULL) return;
 
     PB_Main* request = malloc(sizeof(PB_Main));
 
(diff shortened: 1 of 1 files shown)


chrisdebian:
Verified this against the actual `rpc_send()` implementation: it acquires the same `session->callbacks_mutex` around the `send_bytes_callback` invocation (the real "send" critical section), so `rpc_session_wait_send_done()` correctly blocks until any in-flight send completes before `rpc_app_free` proceeds. Genuine fix for the described free-after-send race, not just a superficial-looking lock.

herbenderbler:
> Verified this against the actual `rpc_send()` implementation: it acquires the same `session->callbacks_mutex` around the `send_bytes_callback` invocation (the real "send" critical section), so `rpc_session_wait_send_done()` correctly blocks until any in-flight send completes before `rpc_app_free` proceeds. Genuine fix for the described free-after-send race, not just a superficial-looking lock.

After getting this onto hardware, I have to walk back that earlier approach.  The force-push replaces it.

You're right that rpc_send() takes callbacks_mutex around the send callback, but that turned out not to be the crashing race. Two problems with the original patch:

1. The mutex doesn't span the hazard. rpc_system_app_confirm() reads rpc_app->last_command_id/->session and rpc_send() runs pb_encode_ex + malloc before callbacks_mutex is ever acquired, so a momentary acquire/release right before free(rpc_app) proves nobody held the mutex at that instant.  It doesn't stop a thread already past the check from walking into the freed context.

1. On-device instrumentation showed the actual failure is different.  Every observed crash (8/8) was rpc_system_app_set_callback() receiving a NULL handle.  The apps null their rpc_ctx inside the SessionClose callback on the teardown thread, while their own thread still holds queued work.  No use-after-free was ever observed, and with the original patch in place the device still hard-faulted 100% of the time on the repro.

The replacement makes the app-facing rpc_app API NULL-tolerant with a call on a closed session being a no-op, bounds the teardown wait, and is verified on hardware over both USB (96 teardowns incl. real IR button presses, 0 crashes) and BLE (fresh pair → mid-session link drop → bond survives reconnection).

turbospok:
How to replicate the crash?

I tried everything described above, BLE session via official mobile companion app - started emulation of file, like ibutton 
Turned off bluetooth at phone, app on flipper just closed
Then tried same with tin foil to block signal, no crashes, rpc app just closed

Same with subghz raw file

How do you do RPC IR emulation from PC if there's no apps that do this? Especially official companion apps

herbenderbler:
> How to replicate the crash?
> 
> I tried everything described above, BLE session via official mobile companion app - started emulation of file, like ibutton Turned off bluetooth at phone, app on flipper just closed Then tried same with tin foil to block signal, no crashes, rpc app just closed
> 
> Same with subghz raw file
> 
> How do you do RPC IR emulation from PC if there's no apps that do this? Especially official companion apps

@turbospok You're not missing anything and your testing was reasonable.  The catch is that this is a data race, not a deterministic crash.  A single manual disconnect almost never lands in the ~millisecond window, so "start emulation → turn off Bluetooth" will close cleanly nearly every time (exactly what you saw).  To hit it you have to keep a command permanently in-flight and drop the link mid-command.  This script does that over USB, which is far more reliable than BLE for repro:

```
#!/usr/bin/env python3
"""
Repro for #4073 — RPC app crash / lost BLE pairing on transport disconnect.

The crash is a race: the app confirms an in-flight RPC command on its own thread
(rpc_system_app_confirm) at the same instant session teardown clears the app's
context on the RPC service thread. A lone manual disconnect rarely lands in that
window, so we keep a command in flight and yank the USB link mid-command.
It's probabilistic — let it loop.

  pip install flipperzero-protobuf pyserial
  (method names track flipperzero_protobuf; tweak if your version differs)
"""
import random, threading, time
import serial
from flipperzero_protobuf.flipper_proto import FlipperProto

PORT = "/dev/ttyACM0"           # Flipper CDC port
APP  = "Infrared"               # or "Sub-GHz", "iButton", "NFC", "125 kHz RFID"
FILE = "/ext/infrared/ac.ir"    # any valid file for that app

def one_attempt():
    port = serial.Serial(PORT, timeout=1)
    proto = FlipperProto(serial_port=port)      # opens the RPC session
    proto.rpc_app_start(APP, "RPC")

    def yank():                                 # drop the link mid-processing
        time.sleep(random.uniform(0.005, 0.080))
        try: port.close()                       # abrupt CDC drop -> teardown runs
        except Exception: pass
    threading.Thread(target=yank, daemon=True).start()

    try:
        while True:
            proto.rpc_app_load_file(FILE)       # app confirms each load on its own thread
    except Exception:
        pass                                    # expected once the port closes

if __name__ == "__main__":
    n = 0
    while True:
        n += 1
        print(f"attempt {n} ...")
        try: one_attempt()
        except Exception: pass
        time.sleep(1.5)                         # let it re-enumerate; watch for a reset
```

Run it against dev (without this PR) and let it loop.  It usually resets the device within a few dozen iterations.  A tight stream of rpc_app_button_press / rpc_app_button_release works the same way if you'd rather drive emulation directly. On BLE the same reset happens before the bond keys are flushed, which is where the lost pairing comes from.

Why it's genuinely a race (two threads, no sync on rpc_ctx):

infrared_rpc_command_callback() runs on the RPC service thread.  On disconnect it does infrared->rpc_ctx = NULL; (infrared_app.c:65-70).
the emulation + acknowledgement run on the app's own thread and read that same pointer: rpc_system_app_confirm(infrared->rpc_ctx, …) (infrared_scene_rpc.c:77, :113, :125).
If the disconnect nulls rpc_ctx while the app thread is mid-confirm, the old furi_check(rpc_app) fires the reset. When you toggle BT off with the app idle, nothing's in flight, so teardown is clean, which is why you never saw it.

On "no PC app does IR RPC emulation": you don't need one.  The official mobile app already drives this exact protocol (AppStart + AppButtonPress) for Sub-GHz / iButton / NFC / IR.  A human tap just never lines up with the disconnect.  Any protobuf RPC client (like the script above) lets you control that timing.

Oskal79:
STOP! Stop using AI slop, try to use KEYBOARD and BRAIN!

herbenderbler:
> STOP! Stop using AI slop, try to use KEYBOARD and BRAIN!

firstbeer

PR #4358, opened 2026-03-14
ISSUE-4336: EventLoop receives callbacks one event late when setting FuriEventFlag from interrupts

### LINKED ISSUE #4336: EventLoop receives callbacks one event late when setting FuriEventFlag from interrupts
### Describe the bug.

I have created a FAP app that handles GPIO interrupts by delegating the events to an `FuriEventLoop` with [furi_event_loop_subscribe_event_flag](https://developer.flipper.net/flipperzero/doxygen/event__loop_8h.html#a28211535228a5b5d5f6d96e6b87b136f).

Interestingly, it seems that every [furi_event_flag_set](https://developer.flipper.net/flipperzero/doxygen/event__flag_8h.html#aa8dfebb078a9aa0aaddf1d5592854ec6) event is delayed, and is delivered after the very next `set` to the same `FuriEventFlag`. I only call `furi_event_flag_set` methods from interrupts.

After the very first interrupt the `FuriEventLoop` does not wake up, and then somehow on every subsequent `set` it wakes up, and when the handler runs and attempts to get the new flag value, it receives exactly the previous value.

I have created an example to demonstrate the issue.

The application sets interrupts for the left and right input GPIOs. When a button is pressed, the interrupt writes the `L` or `R` letters into a buffer„and calls `furi_event_loop_set`. Then the handler in the event loop gets the flags with either `furi_event_flag_get` or `furi_event_flag_clear`, and based on the received flags, it appends the same `L` or `R` character to another buffer. These buffers are printed to the gui. I have the same problem with other interrupts too.

Image

https://github.com/sorgloomer/flipper_eventflag_demo

I am using a periodic timer to redraw the text unconditionally to make sure the problem is not in my gui programming.

### Reproduction

1. Instantiate `FuriEventLoop` and `FuriEventFlag`
2. Subscribe to the event with `furi_event_loop_subscribe_event_flag`
3. Subscribe to an interrupt with `furi_event_loop_subscribe_event_flag`
4. Use `furi_event_flag_set` in the interrupt handler
5. Use `furi_event_flag_wait` in the event handler to get and reset the flags
6. Observe that the individual events arrive late to the interrupts, one interrupt late, in lockstep.

### Target

_No response_

### Logs

```Text

```

### Anything else?

_No response_

herbenderbler:
# What's New

**Problem (fixes #4336)**

When `furi_event_flag_set()` is called from interrupt context (e.g. a GPIO ISR) and the app subscribes via `furi_event_loop_subscribe_event_flag()`, the event loop sees events **one step late**: the first interrupt often does not trigger the callback at all, and later callbacks observe the **previous** flag value.

**Cause**

FreeRTOS's `xEventGroupSetBitsFromISR()` does not set the bits in the ISR — it is a thin wrapper around `xTimerPendFunctionCallFromISR()` that queues the set for the timer daemon. The old code then called `furi_event_loop_link_notify()` immediately, from the ISR. The event loop thread (priority 16) outranks the timer daemon (priority 2), so it woke **first**, read the flag level while the bits were still unset, found nothing, and dropped the item from its waiting list. The daemon then applied the bits with no further notification, so the event was lost entirely.

**Fix**

Notify only **after** the bits are actually set. The ISR path of `furi_event_flag_set()` now uses `xTimerPendFunctionCallFromISR()` to run a callback in the timer daemon that:

1. calls `xEventGroupSetBits()`
2. then calls `furi_event_loop_link_notify()`

By the time the event loop runs, the bits are set and the callback observes the correct value. The timer command queue is unchanged; if it is full, `FuriFlagErrorResource` is still returned. Task-context behaviour is unchanged.

The pended callback runs with the scheduler suspended (`vTaskSuspendAll()` / `xTaskResumeAll()`). `xEventGroupSetBits()` ends in its own `xTaskResumeAll()`, which would otherwise yield to a woken higher-priority task *before* our notify — and that task may legitimately free the object (`FuriApiLock` is a `FuriEventFlag`), leaving the notify to dereference freed memory. Suspending the scheduler makes the inner resume nest rather than switch context, so the instance stays alive across both calls.

**Scope note:** `furi_event_flag_clear()` has the same pre-existing defect on its ISR path (`xEventGroupClearBitsFromISR()` also only queues the clear, while the `FuriEventLoopEventOut` notify fires synchronously). It is deliberately **not** changed here: no in-tree caller invokes `furi_event_flag_clear()` from ISR context — the only callers are `applications/services/bt/bt_service/bt.c` and `api_lock_relock()` (`lib/toolbox/api_lock.h`), all task context — so its `FURI_IS_IRQ_MODE()` branch is unreachable in practice. Fixing it is a separable change; filed as #4417 and fixed by #4418.

# Regression test

`test_furi_event_loop_event_flag_from_isr()` in `applications/debug/unit_tests/tests/furi/furi_event_loop_test.c` uses a software-pending LPTIM2 interrupt to call `furi_event_flag_set()` from ISR context three times (bits 1, 2, 4) and asserts the event-loop callback receives each value, in order.

The event loop is allocated, subscribed, run, unsubscribed and freed entirely on the thread that owns it: `furi_event_loop_free()`, `_run()`, `_unsubscribe()` and the `subscribe_*` object APIs all assert `instance->thread_id == furi_thread_get_current_id()`. `furi_event_loop_stop()` is the one call the test makes cross-thread, which is safe by construction — it only posts a task notification to the owning thread via `furi_event_loop_notify()`. A semaphore handshake ensures the interrupt only fires once the loop thread has subscribed, and teardown happens before the assertions so a failure cannot leave the ISR registered against a dead stack.

Supporting HAL: `furi_hal_interrupt_trigger_pending(FuriHalInterruptId)` in `targets/f7/furi_hal/furi_hal_interrupt.c` sets an interrupt pending so tests can run code in ISR context.

# API surface

**No change to the public FAP API.** No `+` entry is added, removed or changed in either CSV — the public symbol count is unchanged from `dev` for both targets (`grep -cE '^(Function|Variable),\+,'` gives 2739 for f7 and 1937 for f18, same as `dev`) — and both `targets/f7/api_symbols.csv` and `targets/f18/api_symbols.csv` stay at version **88.0**, in lockstep with `dev`.

`furi_hal_interrupt_trigger_pending` is a test-only helper, so it is marked `-` in both CSVs (following the existing precedent of `furi_hal_interrupt_init` in the same header) and exposed to the `test_furi` plugin through the private `applications/debug/unit_tests/unit_test_api_table_i.h` instead. This avoids a permanent public-API commitment that could only be withdrawn later with a major version bump.

# Verification

Verified on physical hardware (Flipper Zero, `f7`) as an A/B against the same commit, gated on `firmware.commit.dirty` from `info device` to rule out a stale self-update:

| Build | `firmware.commit.dirty` | `unit_tests test_furi` |
|---|---|---|
| With the fix | `false` | **PASSED** |
| `event_flag.c` reverted to pre-fix | `true` | **FAILED** — `timeout waiting for event flag callback` |

The pre-fix run fails on the **first** trigger with a timeout rather than a wrong value, which is the expected signature: the event is lost outright, exactly as described under *Cause* above.

Also verified:

- Full `unit_tests` suite passes on the fixed build (`Failed tests: 0`, `Leaked: 0`) — this exercises the shared `FuriApiLock` path that `event_flag.c` sits under.
- `./fbt TARGET_HW=18 firmware_all` builds clean (f18 is built by CI and shares `furi_hal_interrupt.h`).
- `./fbt FIRMWARE_APP_SET=unit_tests firmware_all` and `... resources` build clean; the `test_furi.fal` plugin links and its imports resolve.
- `API version 88.0 is up to date` — fbt requires no version bump; device reports `firmware.api.major 88` / `firmware.api.minor 0`.

## Reproducing the A/B

```bash
# 1. Fixed build
./fbt FIRMWARE_APP_SET=unit_tests flash_usb_full
#    CLI: info device            -> firmware.commit.dirty : false
#    CLI: unit_tests test_furi   -> Status: PASSED

# 2. Revert just the fix and rebuild
git show c9ab2b68:furi/core/event_flag.c > furi/core/event_flag.c
./fbt FIRMWARE_APP_SET=unit_tests flash_usb_full
#    CLI: info device            -> firmware.commit.dirty : true   <- gate on this
#    CLI: unit_tests test_furi   -> Status: FAILED

git checkout furi/core/event_flag.c
```

`flash_usb_full` (not `flash_usb`) is required: the unit tests load as plugins from `/ext/apps_data/unit_tests/plugins`, and `flash_usb` installs firmware only. The device self-updates and reboots, so wait for it to return to the desktop before running the tests — `firmware.commit.dirty` is what confirms which build is actually running (the commit hash is identical either way, since the revert is uncommitted).

# Documentation

- **`documentation/FuriHalInterrupt.md`** — new HAL interrupt reference (init, set_isr, priorities, `furi_hal_interrupt_trigger_pending`, get_name, get_time_in_isr_total), with a `{#furi_hal_interrupt}` anchor and linked from `documentation/doxygen/system.dox`.
- **API comments** in `furi/core/event_flag.h` and `furi/core/event_loop.h` — state that `furi_event_flag_set()` is ISR-safe and that flags set from interrupt context are delivered with the correct value.
- **`ReadMe.md`** — new "Running unit tests" section.
- **`documentation/UnitTests.md`** — example of running a subset (`unit_tests test_furi`).

# Checklist (For Reviewer)

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -90,6 +90,22 @@ Make sure your Flipper is on, and your firmware is functioning. Connect your Fli
 ./fbt flash_usb
 ```
 
+## Running unit tests
+
+Unit tests run on the device. Build firmware with tests enabled, flash it (and copy resources from `build/latest/resources` to the SD card), then in a CLI session run:
+
+```shell
+unit_tests
+```
+
+To run only the Furi tests (e.g. after changing event loop or event flag code):
+
+```shell
+unit_tests test_furi
+```
+
+See [Unit Tests](/documentation/UnitTests.md) for details and how to add tests.
+
 ## Documentation
 
 - [Flipper Build Tool](/documentation/fbt.md) - building, flashing, and debugging Flipper software
--- a/applications/debug/unit_tests/tests/furi/furi_event_loop_test.c
+++ b/applications/debug/unit_tests/tests/furi/furi_event_loop_test.c
@@ -495,6 +495,109 @@ void test_furi_event_loop_self_unsubscribe(void) {
     furi_event_loop_free(event_loop);
 }
 
+/*
+ * Regression test for issue #4336: event loop must receive the correct flag
+ * value when furi_event_flag_set() is called from ISR (no one-event delay).
+ * Uses a software-pending LPTIM2 interrupt to run the ISR path.
+ */
+#define EVENT_FLAG_ISR_TEST_BITS 3U
+#define EVENT_FLAG_ISR_TEST_MASK ((1U << EVENT_FLAG_ISR_TEST_BITS) - 1U)
+
+typedef struct {
+    FuriEventFlag* flag;
+    uint32_t bits_to_set;
+} EventFlagIsrTestIsrContext;
+
+typedef struct {
+    FuriEventFlag* flag;
+    FuriEventLoop* loop;
+    FuriSemaphore* ready;
+    FuriSemaphore* sync;
+    uint32_t bits_received;
+} EventFlagIsrTestContext;
+
+static void event_flag_isr_test_isr(void* context) {
+    EventFlagIsrTestIsrContext* ctx = context;
+    furi_event_flag_set(ctx->flag, ctx->bits_to_set);
+}
+
+static void event_flag_isr_test_callback(FuriEventLoopObject* object, void* context) {
+    EventFlagIsrTestContext* ctx = context;
+    ctx->bits_received =
+        furi_event_flag_wait((FuriEventFlag*)object, EVENT_FLAG_ISR_TEST_MASK, FuriFlagWaitAny, 0);
+    furi_semaphore_release(ctx->sync);
+}
+
+// The event loop is owned by this thread: every FuriEventLoop call except
+// furi_event_loop_stop() must run on the thread the loop was created in.
+static int32_t event_flag_isr_test_loop_thread(void* arg) {
+    EventFlagIsrTestContext* ctx = arg;
+
+    ctx->loop = furi_event_loop_alloc();
+    furi_event_loop_subscribe_event_flag(
+        ctx->loop, ctx->flag, FuriEventLoopEventIn, event_flag_isr_test_callback, ctx);
+
+    furi_semaphore_release(ctx->ready);
+    furi_event_loop_run(ctx->loop);
+
+    furi_event_loop_unsubscribe(ctx->loop, ctx->flag);
+    furi_event_loop_free(ctx->loop);
+
+    return 0;
+}
+
+void test_furi_event_loop_event_flag_from_isr(void) {
+    EventFlagIsrTestContext ctx = {
+        .flag = furi_event_flag_alloc(),
+        .loop = NULL,
+        .ready = furi_semaphore_alloc(1, 0),
+        .sync = furi_semaphore_alloc(1, 0),
+        .bits_received = 0,
+    };
+    EventFlagIsrTestIsrContext isr_ctx = {.flag = ctx.flag, .bits_to_set = 0};
+
+    FuriThread* loop_thread =
+        furi_thread_alloc_ex("event_flag_isr_loop", 1024, event_flag_isr_test_loop_thread, &ctx);
+    furi_thread_start(loop_thread);
+
+    // Only trigger the interrupt once the loop thread is subscribed, otherwise
+    // the flag would be set before anything is listening for it.
+    furi_check(furi_semaphore_acquire(ctx.ready, furi_ms_to_ticks(1000)) == FuriStatusOk);
+
+    furi_hal_interrupt_set_isr(FuriHalInterruptIdLpTim2, event_flag_isr_test_isr, &isr_ctx);
+
+    const uint32_t test_bits[] = {1U, 2U, 4U};
+    uint32_t bits_received[3] = {0};
+    bool timed_out = false;
+
+    for(size_t i = 0; i < sizeof(test_bits) / sizeof(test_bits[0]); i++) {
+        isr_ctx.bits_to_set = test_bits[i];
+        furi_hal_interrupt_trigger_pending(FuriHalInterruptIdLpTim2);
+
+        if(furi_semaphore_acquire(ctx.sync, furi_ms_to_ticks(500)) != FuriStatusOk) {
+            timed_out = true;
+            break;
+        }
+        bits_received[i] = ctx.bits_received;
+    }
+
+    // Tear down before asserting so a failure cannot leave the ISR registered
+    // against this function's stack.
+    furi_hal_interrupt_set_isr(FuriHalInterruptIdLpTim2, NULL, NULL);
+    furi_event_loop_stop(ctx.loop);
+    furi_thread_join(loop_thread);
+    furi_thread_free(loop_thread);
+
+    mu_assert(!timed_out, "timeout waiting for event flag callback");
+    for(size_t i = 0; i < sizeof(test_bits) / sizeof(test_bits[0]); i++) {
+        mu_assert_int_eq(bits_received[i], test_bits[i]);
+    }
+
+    furi_semaphore_free(ctx.sync);
+    furi_semaphore_free(ctx.ready);
+    furi_event_flag_free(ctx.flag);
+}
+
 void test_furi_event_loop(void) {
     TestFuriEventLoopData data = {};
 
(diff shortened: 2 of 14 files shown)


chrisdebian:
Traced through this independently and the core fix is genuinely correct: `xEventGroupSetBitsFromISR()` only queues the bit-set for FreeRTOS's timer daemon to apply later, but the old code called `furi_event_loop_link_notify()` immediately afterward, before those bits were actually visible — a real race, not a hypothetical one. Moving both the set and the notify into a single `xTimerPendFunctionCallFromISR` callback (so the notify only fires once the bits are actually applied) is the standard, correct FreeRTOS pattern here. The new regression test and the `furi_hal_interrupt_trigger_pending` HAL helper it uses are both well-designed and properly wired into the test suite.

One thing worth raising before merge: three of the added files (`Issue_4336_EventLoop_EventFlag_ISR_Review.md`, `Step5_EventFlag_Demo_Verification_Report.md`, `Test_EventFlag_Demo_On_Device.md`) read like working notes/verification reports rather than documentation meant to ship permanently in `documentation/` — one still has unfilled template text ("Replace `#xxxx` and `link-to-pr`..."). Might be worth moving that content into the PR description and trimming these three from the diff, separate from `ReadMe.md`'s unit-test section and `UnitTests.md`, which do read like real docs.

chrisdebian:
Noticed the doc cleanup in the latest push — the three working-note files are gone now, just the genuine docs (`FuriHalInterrupt.md`, `UnitTests.md`) remain. That addresses the concern I flagged earlier. Thanks for sorting it.

herbenderbler:
> Traced through this independently and the core fix is genuinely correct: `xEventGroupSetBitsFromISR()` only queues the bit-set for FreeRTOS's timer daemon to apply later, but the old code called `furi_event_loop_link_notify()` immediately afterward, before those bits were actually visible — a real race, not a hypothetical one. Moving both the set and the notify into a single `xTimerPendFunctionCallFromISR` callback (so the notify only fires once the bits are actually applied) is the standard, correct FreeRTOS pattern here. The new regression test and the `furi_hal_interrupt_trigger_pending` HAL helper it uses are both well-designed and properly wired into the test suite.
> 
> One thing worth raising before merge: three of the added files (`Issue_4336_EventLoop_EventFlag_ISR_Review.md`, `Step5_EventFlag_Demo_Verification_Report.md`, `Test_EventFlag_Demo_On_Device.md`) read like working notes/verification reports rather than documentation meant to ship permanently in `documentation/` — one still has unfilled template text ("Replace `#xxxx` and `link-to-pr`..."). Might be worth moving that content into the PR description and trimming these three from the diff, separate from `ReadMe.md`'s unit-test section and `UnitTests.md`, which do read like real docs.

Thanks @chrisdebian!

All three docs are gone (26997004), plus the `js_event_loop.md` section that referenced one. The template text was real: line 124 shipped a literal `#xxxx`/`link-to-pr`. It would have published, too, since `Doxyfile.cfg` includes `documentation/` recursively and `docs.yml` syncs to developer.flipper.net on push to `dev`. Those three were the only `.md` files there without a doxygen anchor. `FuriHalInterrupt.md` has one and stays.

Your review prompted a wider pass that fixed three more issues:

* **Test couldn't run** (e93525c0). It allocated the event loop on the test thread and ran it on another, tripping the thread affinity `furi_check` in `furi_event_loop_run()`. A guaranteed crash, not a race.
* **API surface** (3ea81f96). Dropped the unrelated `uxTopUsedPriority` export. `furi_hal_interrupt_trigger_pending` is now marked `-` and routed through the private unit test table. No public API change, both CSVs back to 88.0 in lockstep with `dev`, which also unbroke the f18 build.
* **Use after free** (4875abec). `xEventGroupSetBits()` ends in `xTaskResumeAll()`, which can yield to a task that frees the flag (`FuriApiLock` is a `FuriEventFlag`) before the notify runs. Now scheduler suspended.

Also merged current `dev`.

The test is now a real regression test: it fails with `timeout waiting for event flag callback` on pre-fix code and passes with the fix, gated on `firmware.commit.dirty` to rule out stale firmware. Description updated; the old verification section described a run that couldn't have happened.

herbenderbler:
> Noticed the doc cleanup in the latest push — the three working-note files are gone now, just the genuine docs (`FuriHalInterrupt.md`, `UnitTests.md`) remain. That addresses the concern I flagged earlier. Thanks for sorting it.

I _think_ I've hammered out all of the rough edges of this PR.  Feel free to give it another once over whenever you get a chance.  Thanks again!

PR #4361, opened 2026-03-20
Fix USB/BLE HID limits to support international keyboards and add ja-JP keyboard layout for BadUSB

d3npa:
I was having trouble sending pipe characters in badusb with my jp-106 key layout and it seems the reason was the limits set on the HID descriptor. The pipe character is typed with Shift+Yen, which sends a 0x89 keycode (outside the current upstream limit of dec 101 / 0x65). Backslash `\` (0x87, Ro key) and underscore `_` (Shift+0x87) were also affected for the same reason. The computer silently drops keycodes outside the range declared in the descriptor, so these characters never arrived.

# What's new

- `LOGICAL_MAXIMUM` and `USAGE_MAXIMUM` are raised to 255 to allow keycodes above 101 from international keyboards. 

# Verification 

1. Generate a layout def for basusb and drop in `/badusb/assets/layouts/ja-JP.kl` (script by Claude Opus 4.6)

```py
#!/usr/bin/env python3
"""Generate Flipper Zero BadUSB keyboard layout for JIS 106-key."""

OUTPUT = "ja-JP.kl"
SHIFT = 0x02

layout = bytearray(256)  # 128 entries x 2 bytes (keycode, modifier)

def key(char, hid, mod=0):
    layout[ord(char) * 2] = hid
    layout[ord(char) * 2 + 1] = mod

# Letters
for i in range(26):
    hid = 0x04 + i
    key(chr(ord('a') + i), hid)
    key(chr(ord('A') + i), hid, SHIFT)

# Digits
for digit, hid in zip("1234567890", [0x1E, 0x1F, 0x20, 0x21, 0x22, 0x23, 0x24, 0x25, 0x26, 0x27]):
    key(digit, hid)

# Shifted digits: ! " # $ % & ' ( )
for char, hid in zip("!\"#$%&'()", [0x1E, 0x1F, 0x20, 0x21, 0x22, 0x23, 0x24, 0x25, 0x26]):
    key(char, hid, SHIFT)

# Symbols
key(" ",  0x2C)
key("-",  0x2D);  key("=",  0x2D, SHIFT)
key("^",  0x2E);  key("~",  0x2E, SHIFT)
key("@",  0x2F);  key("`",  0x2F, SHIFT)
key("[",  0x30);  key("{",  0x30, SHIFT)
key("]",  0x32);  key("}",  0x32, SHIFT)
key(";",  0x33);  key("+",  0x33, SHIFT)
key(":",  0x34);  key("*",  0x34, SHIFT)
[...]

PR #4373, opened 2026-05-21
BadUSB: Add Japanese (JIS) keyboard layout

Ytz-Ichi:
# What's new

- Add `ja-JP.kl` for the Japanese 109-key (JIS) keyboard layout.
- Extend the HID keyboard report descriptor (USB and BLE) so International1..5
  key codes (0x87-0x8B) reach the host. The previous `USAGE_MAXIMUM` of
  `HID_KEYBOARD_APPLICATION` (0x65) caused reports containing these keys to
  be dropped by the OS, so a layout-only change could not cover JIS. The new
  range is a strict superset of the previous one, so existing layouts are
  unaffected.

JIS layout uses International3 (yen key) for `\` and `|`, and International1
(ro key) for `_`.

# Verification

1. `./fbt firmware_all flash_usb` to build and flash. The new layout is
   bundled with the resources.
2. In the BadUSB app, open the Config menu and select `ja-JP` as the layout.
3. On the host PC, switch the keyboard layout to Japanese (JIS) and open
   Notepad.
4. Run a BadUSB script that prints the printable ASCII range, e.g.:

   ```
   STRING ASCII test: !"#$%&'()*+,-./0123456789:;<=>?@ABCXYZabcxyz[\]^_`{|}~
   STRINGLN
   STRING Backslash and pipe: \\ ||
   ```

5. All characters should appear as typed. The previously broken set
   (`\` `|` `_` `@` `[` `]` `{` `}` `` ` `` `~` `^` `:` `"` `'`) now
   produces the printed characters. `\` renders as ¥ on Japanese Windows
   because of Shift-JIS font mapping; the byte sent is still ASCII 0x5C.

# Checklist (For Reviewer)


Ytz-Ichi:
Sorry, I found same issue pull request.
This PR is closing soon.
https://github.com/flipperdevices/flipperzero-firmware/pull/4361

PR #4383, opened 2026-06-11
NFC FeliCa: Add Standard card emulation support

nikachu2012:
## Summary

- Added full FeliCa Standard card listener (emulator) support, handling multi-system/area/service structures
- Implemented **Request Response** (0x04), **Request Service** (0x02), **Search Service Code** (0x0A), and **Request System Code** (0x0C) commands in the listener
- Added Standard-mode **Read/Write Without Encryption** (0x06/0x08) handlers supporting multi-service block lists
- Poller now issues **Request Service** during Standard system traversal to fetch key versions (batched, ≤ 32 nodes per command per FeliCa spec)
- Extended data model: `FelicaSystem`, `FelicaArea`, `FelicaService` now carry `key_version`; `FelicaArea` carries `end_code`
- Updated `.nfc` file save/load to persist the new fields with backward-compatible fallbacks for old file formats
- Added unit test `felica_standard_read` with a test card fixture (`Felica_Standard_with_keys.nfc`)

## Test plan

- [x] Run unit tests: `felica_standard_read` passes (listener emulates Standard card, poller reads key versions and block data correctly)
- [x] Verify existing `felica_read` and `felica_read_auth` tests still pass (no regression in Lite-S flow)
- [ ] Test against a real FeliCa Standard card to confirm Polling / Search Service Code / Request System Code responses
- [x] Confirm `.nfc` files saved with the new format load correctly; confirm old `.nfc` files (without `key_version`/`end_code`) still load with sensible defaults

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nikachu2012:
memo 
write use `./fbt usb_flash_full`
If you don't use this, the NFC app won't update.

zinongli:
> Implemented Request Response (0x04), Request Service (0x02), Search Service Code (0x0A), and Request System Code (0x0C) commands in the listener

I tested this with a proxmark3 using `hf felica raw -cs` + corresponding commands. All of these listener handlers are working perfectly and the tag responses are identical to the tag I'm copying from, which is an Apple Wallet Mobile Suica. 

Also the backward compatibility is working as intended. Older version files are opening and rendering just fine. 

I do wanted to ask: would it be easier to just bump the [`Data Format Version`](https://github.com/flipperdevices/flipperzero-firmware/pull/4383/changes#diff-2b00850c41f9d31f8196d44629f5b08d9c99b3f4a24fece7bbc0eeb323d3da99R8) from 2 to 3? 

I don't have any objections with the current sscanf + fallback solution. Just pointing out that an alternative `swtich(data_format)` + `case 2: ` + `case 3: ` solution might make things easier to understand for future readers.

nikachu2012:
I think it might be a good idea to switch to a different version.
I feel like having too many fallbacks might make things complicated.

In addition, there are still bugs in the reading, emulation, and information display functions for cards with multiple systems, and we are currently debugging them.

turbospok (changes requested):
Hello!

New unit test is failing:

```
2026-07-27 16:06:24,161 [INFO] [-]felica_standard_read failed:
2026-07-27 16:06:24,162 [INFO] 	applications/debug/unit_tests/tests/nfc/nfc_test.c:717: system key_version != 0x0001
```

also I think we should bump `Data format version` since format was modified
Because the version stayed at 2, a PR-saved dump is indistinguishable from a dev-saved dump, but the Area line layout changed

nikachu2012:
Thank you for the review!
I’ve made some fixes so that the unit tests pass. I’ve also updated the file version to 3. Additionally, I’ve modified the code to load the file based on the version.
I’m also developing a version that supports cryptographic protocols, but I’ll leave it as is for now.
https://github.com/nikachu2012/flipperzero-firmware/tree/mutual_encryption

turbospok:
Tag me when ready to re-review, and please verify on real felica card first, don't forget to merge latest dev branch

UPD: seems like you answered just when I was sending comment, ha

turbospok:
Unit tests pass now, thx
@nikachu2012 can you verify current version on real card ? (I don't have one)

nikachu2012:
@turbospok 
The newly read FeliCa card was saved as Version 3.

turbospok (approved):
Thanks!

turbospok:
We might delay the merge for next release, since current one we are preparing - is pretty packed with fixes and its gonna be mostly fixes only release, next one is planned to be more feature related one
Will let you know if we need anything to change, but after latest patch it looks fine to me

PR #4384, opened 2026-06-12
[PoC] Add localisation support

smileyhead:
# Preface

I would like to help expanding the user base of Flipper Zero by helping implement a translation layer and providing Hungarian translations. I have 10 years of software localisation experience. This project is only in a proof-of-concept phase right now, so I would really appreciate some feedback and guidance.

I'm only familiar with C#, so I'm feeling a bit lost, but at the very least, I have managed to implement a language switcher UI, a localisation function, and wrapped `about.c`'s strings into said function.

# Plan

My plan is a system which uses `gettext` to extract strings wrapped in the `i18n()` function into files. Translators then edit those files to provide translations, and these same files can be read by the system to replace strings at runtime.

# Implementation checklist

**Text localisation**
- [x] Basic `i18n()` function
- [ ] `i18np()` function for handling plurals
- [ ] Localisation function for app names (`.fam`)
- [ ] Wrapping all user-facing strings in localisation functions

**Asset localisation**
- [ ] Make bubbles translatable (`meta.txt`)
- [ ] Make non-bubble text box
- [ ] Adapt non-bubble texts to use translatable text boxes

**`gettext` & R/W**
- [ ] Implementing `gettext` to extract strings
- [ ] Writing language files to firmware
- [ ] Reading language files from internal storage at runtime

# What I need help with

Understanding how accessing storage works and how to include localised files in the built firmware directly.

# AI usage disclosure

Like I said previously, C is new to me, but I do have some experience with C#. The initial implementation of the settings entry (and its storage) was made with GitHub Copilot, whose implementation I looked through, cherry-picked, revised, and then put into place. This initial implementation is the first commit of my branch – everything else was done manually.

[Képernyőátjátszás_20260612_193540.webm](https://github.com/user-attachments/assets/c475f54e-cbed-4fd7-8259-0b3fcb95535f)

Any and all feedback would be most welcome. I'm a translator first, so my code quality is probably questionable, but I'll try my best to help implement a translation system for Flipper Zero.

# What's new

- Language switcher in Settings -> System
- `i18n()` function

# Verification 

- Go to Settings -> System and change the language to `magyar` (Hungarian)
- Back out to Settings and enter About to see strings changed

# Checklist (For Reviewer)


kalicyh:
I’ve also been experimenting with a related localization/font approach in my fork.

Would you mind taking a look when you have time?

https://github.com/Next-Flip/Momentum-Firmware/compare/dev...kalicyh:Momentum-Firmware-CN:CN-Clipper

https://github.com/kalicyh/Momentum-Firmware-CN

smileyhead:
Hi, does your implementation have switchable languages?

Unfortunately, after experimenting a little and talking to people in the community Discord, it seems that there is no RAM to spare for a `gettext()`-style implementation, which would've been my idea. Also, `gettext` itself doesn't seem to have been ported to embedded systems in a capacity that works.

In my opinion, the best way for tackling localisation currently would be to have the updater app (on PC/mobile) essentially rebuild the firmware during the update process with the localised strings baked in, with no option to switch languages on the Flipper itself. The updater already does region-checking, anyway. This would ensure around the same level of memory usage as the current, English-only implementation.

This would also apply to apps, so the app store would bake in the appropriate language (if available) during the install process.

kalicyh:
> Hi, does your implementation have switchable languages?
> 
> Unfortunately, after experimenting a little and talking to people in the community Discord, it seems that there is no RAM to spare for a `gettext()`-style implementation, which would've been my idea. Also, `gettext` itself doesn't seem to have been ported to embedded systems in a capacity that works.
> 
> In my opinion, the best way for tackling localisation currently would be to have the updater app (on PC/mobile) essentially rebuild the firmware during the update process with the localised strings baked in, with no option to switch languages on the Flipper itself. The updater already does region-checking, anyway. This would ensure around the same level of memory usage as the current, English-only implementation.
> 
> This would also apply to apps, so the app store would bake in the appropriate language (if available) during the install process.

Hi, no, my implementation does not have switchable languages on the Flipper itself.

It is closer to the direction you described: the localized strings/font are baked into the firmware at build time, with no gettext-style runtime lookup and no language selector on-device. For Chinese, the font is the expensive part, so I tried to keep it as firmware rodata and avoid loading/parsing translation data at runtime.

I also had to keep the updater/RAM stage in mind. In my branch the main firmware can use the Chinese font, but the RAM updater path should not pull in the large font data, otherwise it can exceed the loader size limits.

So I agree that rebuilding or selecting a localized firmware/app artifact during the PC/mobile update/install process is probably the more realistic approach for now. On-device switching would be nicer UX, but it seems too expensive for RAM/flash unless there is a very compact catalog format and very careful font handling.

smileyhead:
In that case, I would love to see this realised and implemented in an official capacity (manually publishing alt firmwares with no official discoverability would likely kill the project, after all…)

I can see your fork is based on a fork of the firmware and not the official one. Do you think it could be made compatible with the official firmware?

If your implementation is the same as my idea, and you've gotten further with it, I see no reason to keep my PR open. I'd help with your fork instead, by implementing a Hungarian translation and helping with anything else you might need help with, though, I still am not very good at C… I would have to take a look at your custom font implementation, though, as Hungarian will need a few more characters than there are available in the base font.

kalicyh:
Yes, I think the idea is very close, but my branch is not a clean official-firmware implementation yet.

I tried several approaches before getting to the current one:

1. A fully built-in font approach, with the font data directly compiled into the firmware.
2. An external resource / asset-pack style approach, where the font is loaded from external resources.
3. An earlier runtime-loading approach in this branch, which worked but caused problems around DFU/updater/RAM constraints.
4. The current approach: generate the needed localized font data at build time, link it into firmware `.rodata`, and keep the RAM updater path from pulling in the large font.

So the current branch has no on-device language switching. It is closer to your idea: build/select a localized firmware artifact, with strings and required font data baked in.

About official firmware compatibility: I think the concept can be made compatible, but not by merging my fork as-is. My fork is based on Momentum and contains a lot of unrelated changes. The useful part would need to be extracted into a much smaller patch for official firmware: build-time string/font generation, baked-in localized assets.

For Hungarian, the font side should be much easier than Chinese. Chinese needs a large custom glyph set, while Hungarian likely only needs a few extra Latin characters. The same build-time font-generation idea should still work, just with a much smaller glyph set.

smileyhead:
That sounds good. I will close my draft for now and keep an eye on your project. If you need any help, of course, I would be happy to contribute. Shall I contact you via Telegram?

If in the long-term, your changes could be made into an official firmware-compatible format and added to the official updaters, it would be wonderful.

kalicyh:
> That sounds good. I will close my draft for now and keep an eye on your project. If you need any help, of course, I would be happy to contribute. Shall I contact you via Telegram?
> 
> If in the long-term, your changes could be made into an official firmware-compatible format and added to the official updaters, it would be wonderful.



> 听起来不错。我暂时先关闭草稿，之后会关注你的项目。如果你需要任何帮助，我当然很乐意贡献力量。我可以通过 Telegram 联系你吗？
> 
> 从长远来看，如果您的更改能够被制作成官方固件兼容格式并添加到官方更新程序中，那就太好了。

Sure, Telegram is fine: @sukalicyh

Thanks for offering to help. Long term, I also hope this can become official-firmware-compatible and work through the official updater flow.

PR #4415, opened 2026-07-14
docs: add build, release, contribution, and license badges to ReadMe

chrisdebian:
Following up on #4414 — adds four badges under the title, before the existing bullet list:

- **Build** — GitHub Actions status for `build.yml`
- **Latest Release** — current tagged release
- **Open for Contribution** — live count of issues carrying that label
- **License** — GPL-3.0

Verified all four badge endpoints resolve (HTTP 200) and checked every existing link in ReadMe.md still works before touching the file — nothing else needed changing.

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -8,6 +8,11 @@
 
 # Flipper Zero Firmware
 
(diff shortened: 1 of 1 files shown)


turbospok:
I think we need to remove Open for Contribution label tracking since we don't use it
Other badges are fine

chrisdebian:
Removed the Open for Contribution badge. Build, Latest Release, and License remain.

PR #4421, opened 2026-07-24
Add method for changing variable item label

loftyinclination:
# What's new

- A new method has been added which writes to a `variable_item`'s `label` field. This works the same way as the other methods on `variable_item` (ie; it does not lock the `variable_item_list`'s model).
- Without this, there is no way of changing the label -- the most likely way to generate a mutatable `char*` would be via `FuriString`, however the pointer that this returns is invalidated by any writes, as per the m-lib documentation.

# Verification 

- Use this method when using the variable item list. After the change to the label is made, it'll be shown the next time that the view is redrawn.

- [ ] Partially AI assisted (clarify below which code was AI assisted and briefly explain what it does).

- [ Describe here how AI was used in this PR if it was used ]

### DIFF
--- a/applications/services/gui/modules/variable_item_list.c
+++ b/applications/services/gui/modules/variable_item_list.c
@@ -394,7 +394,7 @@ void variable_item_list_set_enter_callback(
         false);
 }
 
-void variable_item_set_label(VariableItem* item, const char* label) {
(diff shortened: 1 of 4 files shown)


turbospok (approved):
Thanks!

PR #4422, opened 2026-07-24
Allow for checking if an id is in use before adding to view dispatcher

loftyinclination:
- **Include documentation about caller thread context**
- **Add method for checking if view dispatcher id is already in use**

# What's new

- A new method has been added that checks the `view_dispatcher`'s dictionary to see if an id is already in use. This is useful as, without this, attempting to add an item with an id that is already in use, or remove an item at an id that isn't in use, results in a crash.

# Verification 

- [ ] Partially AI assisted (clarify below which code was AI assisted and briefly explain what it does).

- [ Describe here how AI was used in this PR if it was used ]

### DIFF
--- a/applications/services/gui/view_dispatcher.h
+++ b/applications/services/gui/view_dispatcher.h
@@ -97,6 +97,8 @@ void view_dispatcher_set_custom_event_callback(
 
 /** Set navigation event handler
  *
+ * @note this will be called on the thread that invoked view_dispatcher_run
+ *
  * Called on Input Short Back Event, if it is not consumed by view
  *
  * @param      view_dispatcher  ViewDispatcher instance
--- a/applications/services/gui/view_dispatcher.h
+++ b/applications/services/gui/view_dispatcher.h
@@ -161,6 +161,14 @@ void view_dispatcher_run(ViewDispatcher* view_dispatcher);
  */
 void view_dispatcher_stop(ViewDispatcher* view_dispatcher);
 
+/** Check if a view exists at id.
+ *
+ * @param      view_id          View id to check
+ *
+ * @return     True if a view has been added with id, false if not.
+ */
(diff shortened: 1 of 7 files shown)


turbospok (changes requested):
Hello!

Why do we need to set submenu argument to const in that specific function?
In every other "get" like `submenu_get_view` it is not const while we still do read only

Is there any specific reason for that?

Other change (id check) is ok, this one is not very clear

loftyinclination:
It doesn't need to be, I just ran into some slight difficulty when writing some FFI bindings for flipperzero-rs on both of these (reasoning about which thread things were happening on and whether it was possible for the struct to be written to concurrently). Both of these changes were made to make reasoning about things easier to handle, but the ID one is really the only necessary change.

turbospok:
> It doesn't need to be, I just ran into some slight difficulty when writing some FFI bindings for flipperzero-rs on both of these (reasoning about which thread things were happening on and whether it was possible for the struct to be written to concurrently). Both of these changes were made to make reasoning about things easier to handle, but the ID one is really the only necessary change.

Lets remove that change from this PR for now then
Since I think we can add that ID check right for upcoming release

turbospok (approved):
Thanks!

applications/services/gui/view_dispatcher.h
```diff
@@ -159,6 +161,14 @@ void view_dispatcher_run(ViewDispatcher* view_dispatcher);
  */
 void view_dispatcher_stop(ViewDispatcher* view_dispatcher);
 
+/** Check if a view exists at id.
+ *
+ * @param      view_id          View id to check
```
loftyinclination, line 166: oops, missing `view_dispatcher` documentation, which is causing dev to fail the build

xMasterX: one line is missing :(

PR #4425, opened 2026-07-27
js_event_loop: reject a zero or negative timer interval

munzzyy:
`eventLoop.timer(mode, interval)` passes `interval` straight to `furi_ms_to_ticks()` with no check. A periodic timer with interval `0` then divides by zero in `furi_event_loop_process_expired_timers()` (`elapsed_time / timer->interval`), and since a zero-interval timer counts as expired on every pass, this one-liner pins the thread running the loop:

```js
let eventLoop = require("event_loop");
eventLoop.timer("periodic", 0);
eventLoop.run();
```

Any `.js` app can hit it. This rejects `interval <= 0` at the binding with a normal script error, the same way the other js_app modules report bad arguments. The `<= 0` also catches negative values, which would otherwise wrap to a huge interval through the `(uint32_t)` cast.

Traced the parse-to-divide path in the source; the divide and the always-expired check live in `furi/core/event_loop_timer.c`. Not run on hardware.

### DIFF
--- a/applications/system/js_app/modules/js_event_loop/js_event_loop.c
+++ b/applications/system/js_app/modules/js_event_loop/js_event_loop.c
@@ -263,6 +263,11 @@ static void js_event_loop_timer(struct mjs* mjs) {
     int32_t interval;
     JS_VALUE_PARSE_ARGS_OR_RETURN(mjs, &js_loop_timer_args, &mode, &interval);
 
+    if(interval <= 0) {
+        mjs_prepend_errorf(mjs, MJS_BAD_ARGS_ERROR, "timer interval must be a positive number");
+        return;
+    }
+
     JsEventLoop* module = JS_GET_CONTEXT(mjs);
 
     // make timer contract


munzzyy:
I skipped the template when I opened this, so filling in the AI usage disclosure properly:

- **Partially AI assisted.**

I used an AI assistant while auditing the `js_app` bindings for arguments that reach furi APIs unchecked. The change is the `if(interval <= 0)` guard that rejects the value with a script-level error, matching how `js_gui/byte_input.c` reports bad arguments. I traced the path from the binding to the `elapsed_time / timer->interval` division in `furi/core/event_loop_timer.c` myself, and confirmed a zero-interval periodic timer is treated as expired on every pass.

To be clear about testing: I reasoned this one from the source and have not run the reproducing script on hardware.

turbospok (approved):
Makes sense, also `oneshot` can work with 0 tick(ms) value, only `pereodic` can fall into issue
But it does not make sense to make those timers with 0ms, so your fix is good

Also worth to note
The repro in the PR body does not reproduce. `eventLoop.timer("periodic", 0)` only allocates a `JsEventLoopContract`
Without a `subscribe`, `run()` just blocks on an empty loop. You need `eventLoop.subscribe(eventLoop.timer("periodic", 0), fn)`

Seems like you are missing some parts of correct event loop based code

But that does not affect the fix, I'll accept it. thanks

munzzyy:
You're right about the repro, I checked it after your comment. `js_event_loop_timer` only mallocs the `JsEventLoopContract` with `.object = NULL`; `furi_event_loop_timer_alloc` and `furi_event_loop_timer_start` don't run until `js_event_loop_subscribe` hits the `JsEventLoopObjectTypeTimer` case. So without a subscribe there's no timer on the loop and `run()` just sits there. The line in the PR body should have been:

```js
eventLoop.subscribe(eventLoop.timer("periodic", 0), function() {});
eventLoop.run();
```

Good catch, and thanks for taking the fix anyway.

Issue #4314, opened 2025-12-01
Reading saved Felica Standart emulation by another FZ leads to its crash


RebornedBrain:
### Describe the bug.

@zinongli There is a problem with new Felica Standard emulation/reading. See reproducion steps below. Issue happens here because system_total equals 0.

<img width="1545" height="1144" alt="Image" src="https://github.com/user-attachments/assets/624091b8-0d2f-43c8-a1f7-348c562ae83d" />

Looks like the roots of this issue are from doing load/save operations a little bit incorrectly. In those `felica_load` and `felica_save` functions there are bool flags `parsed` and `saved` respectively, which indicates the result of the operation. Newly added logic doesn't modify them. Let's take as an example `felica_load` function, there we have this check in red square which checks that everything before it was parsed fine, otherwise we exit with false. If everything is fine we continue and `parsed` is true, but what if we fail to read smth in code below? We also need to set that flag correctly.

<img width="1589" height="1167" alt="Image" src="https://github.com/user-attachments/assets/eb7d398b-667a-4401-9685-f5397d8f2152" />

And how issue happens, we start to read empy dump where there is no services for example, we reach this part and exit, but return parsed = true, while in fact we didn't complete reading and need to return false, for logic to work correctly.

<img width="1653" height="890" alt="Image" src="https://github.com/user-attachments/assets/f85f6251-fc41-4ddc-8a10-4d834945c738" />

Also I want to warn that fixing this issue by simply adding `parsed = false` to each break will not work (I've already tried 😿) because it will lead to another issue that completely fine blank Felica Standard cards will fail to open for emulation on FZ. 

Maybe some more checks about blank parts of Felica Standard could help, as an idea if we read count == 0, then we skip simple_array_init, or smth like that.

### Reproduction

1. You need two Flipper Zeros flashed with latest 1.4.2 release
2. Read empty Felica Standart card with one Flipper and save it.
3. Open new saved card and start emulation
4. Take another Flipper and try to read emulation.
Result: Flipper which performs reading crashes and reboot is needed to bring it back

### Target

1.4.2

### Logs

```Text
556561 [I][AnimationManager] Unload animation 'L2_Wake_up_128x64'
556575 [I][Loader] Loading /ext/apps/NFC/nfc.fap
556761 [I][Elf] Total size of loaded sections: 81253
556764 [I][Loader] Loaded in 189ms
574308 [I][NfcScanner] Detected 1 protocols
[...]

PR #4428, opened 2026-07-27
NFC FeliCa: check the poller error before reading the system-code response

munzzyy:
# What's new

`felica_poller_state_handler_list_system()` reads `response_system_code->system_count` before checking whether `felica_poller_list_system_code()` succeeded. That function only writes the response pointer on the success path. On a timeout, a bad CRC, or a response shorter than 13 bytes it returns an error and leaves the pointer untouched, and since it's an uninitialized stack variable, every error path here dereferences garbage.

This is the default path for any Standard FeliCa card (Suica/Pasmo/Octopus-style), since auth is skipped by default. So a weak read, a card pulled away mid-read, or an emulator that doesn't answer `Request System Code` all crash the reader.

This is the crash in #4314. The write-up there says the same thing: the other side doesn't respond, and the reader shouldn't freeze on it. #4383 helps the no-response case by teaching the listener to answer, but it doesn't touch this function, so a short or malformed reply (or a non-Flipper emulator that still doesn't answer) hits the same dereference.

The fix moves the response handling inside the `error == FelicaErrorNone` branch, which is what every other handler in this file already does (`traverse_standard_system`, `auth_internal`, `auth_external`, `read_standard_blocks` all check the error before touching the response).

# Verification

Traced `felica_poller_list_system_code()` and confirmed it returns on the frame-exchange error and the short-response check without ever assigning `*response_ptr`, while the caller dereferenced it unconditionally. Compiled the two functions standalone against a stubbed transport and confirmed the out-parameter is left at its sentinel value on the error path, and that MemorySanitizer flags the caller's dereference as use-of-uninitialized-value. `clang-format` is clean.

I don't have hardware to flash, so I haven't run it on-device. To confirm there: reproduce #4314 (one Flipper reading another emulating a blank/partial FeliCa Standard save), or make `Request System Code` time out (weak signal, pull the card early), and check the reader reports a normal read failure instead of freezing. A normal Standard FeliCa read should behave exactly as before.

- [x] I've read the contribution guidelines and my PR follows them.

I used an AI assistant to help trace the `felica_poller.c` / `felica_poller_i.c` call path and to build a small standalone reproduction (outside this repo) that compiles the real `felica_poller_list_system_code` body against a stubbed transport, to confirm the out-parameter is left untouched on the error path and that MemorySanitizer catches the dereference. The fix itself is a hand-written reorder that matches the error-checking pattern the other handlers in this file already use.

### DIFF
--- a/lib/nfc/protocols/felica/felica_poller.c
+++ b/lib/nfc/protocols/felica/felica_poller.c
@@ -129,17 +129,17 @@ NfcCommand felica_poller_state_handler_list_system(FelicaPoller* instance) {
     FelicaListSystemCodeCommandResponse* response_system_code;
     FelicaError error = felica_poller_list_system_code(instance, &response_system_code);
 
-    instance->systems_total = response_system_code->system_count;
-    simple_array_init(instance->data->systems, instance->systems_total);
-    uint8_t* system_codes = response_system_code->system_code;
-
-    for(uint8_t i = 0; i < instance->systems_total; i++) {
-        FelicaSystem* system = simple_array_get(instance->data->systems, i);
-        system->system_code = system_codes[i * 2] << 8 | system_codes[i * 2 + 1];
-        system->system_code_idx = i;
-    }
-
     if(error == FelicaErrorNone) {
+        instance->systems_total = response_system_code->system_count;
+        simple_array_init(instance->data->systems, instance->systems_total);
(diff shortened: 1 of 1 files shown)


turbospok (approved):
Thanks!

jeyjoy123r-stack (commented):
Workflow

PR #4432, opened 2026-08-04
Bring back support for kickstarter/prerelease flipper zero (f6)

dappermint:
## What's new

remember these? the kickstarter/prerelease units (f6) that started it all, abandoned by firmware updates somewhere around 2021. this PR brings the f6 hardware target back from the dead and keeps it building against current dev

- restores `targets/f6` as a first-class target inheriting from f7, with its own `furi_hal` overrides where the prerelease hardware differs
- skips radio stack updates on f6 (the prerelease copro layout doesn't take them)
- restores rtc boot mode handling for the older board
- everything else rides on top of current dev unchanged, `./fbt TARGET_HW=6` just works

## Verification

- built with `./fbt TARGET_HW=6` on top of current dev
- flashed and running on an actual f6 unit, it lives

## Checklist (for the reviewer)

- [ ] i acknowledge that the f6 deserves love too


turbospok:
Hi, F6 hardware was never distributed to kickstarter backers, they were present in some places at the beginning
Like hackathon and other events. 

And no there's no plans to make it supported in main firmware, its OK to have a fork that will work on them, however official firmware is meant only for actual release hardware

Also there's a ways you can modify hardware itself to be F7 compatible, leaving only OTP byte that can be patched in your fork

dappermint:
got it, thanks, this was mainly meant to be a meme, and im not that proficient with hardware haha

PR #4449, opened 2026-09-12
lfrfid: fix Keri reads producing a wrong credential

mfcarroll:
# What's new

`rfid read` cannot read a Keri tag. It will not even read back a Keri card the Flipper itself
just wrote. Every decode it produces from a genuine T5577 Keri card is wrong; `validate_count`
usually stops those reaching the UI, but `rfid raw_analyze` has no such gate and prints the wrong
value as a credential, with a plausible facility code and card number.

**Cause.** The PSK slicer reports one level systematically long and the other short by the same
amount. A Keri frame is never DC balanced — its preamble alone carries a run of 29 zeros, and on
the tag measured here only 10 of 64 bits are set — and on a frame that unbalanced the skew
measures ~135 us, past the ±127 us that `(time + KERI_US_PER_BIT / 2) / KERI_US_PER_BIT`
tolerates. So every run of ones decodes one bit too long. Long runs of zeros survive: the true
bit period is nearer 256 us than `KERI_US_PER_BIT`'s 255, and over a long enough run that
difference offsets the skew. On the same T5577 rewritten as Idteck, whose frame is near half
ones, the skew is only +40/−42 us, which is why the other PSK protocols are unaffected.

`protocol_keri_decoder_feed()` already carries a correction for this — the block commented
"Try to decode wrong phase synced data" — but it shifts the boundary the wrong way. It lengthens
the high level, which here is already ~135 us too long, by a further 120 us. That adds a bit to
every run of ones and takes one off every short run of zeros, which keeps the frame at exactly 64
bits — close enough that jitter on a marginal capture lets both preamble checks pass over a
mangled payload. That is what turns a failed read into a printed credential, and why the wrong
values differ from read to read rather than being stable.

**Changes.**

1. Correct the skew in both directions. The pre-existing buffers cover a high level reported
   short; a new pair covers a high level reported long, which is what every real Keri tag measured
   here presents, and is tried first. Both are kept, so a demodulator with the opposite skew still
   decodes. The four buffers are a true 2×2 — {high long, high short} × {bit = level, bit = !level}
   — because the duration is adjusted against the physical level before the polarity hypothesis is
   applied. They are now named `high_long_*` / `high_short_*` for the slicer behaviour each assumes,
   and the repeated feed-and-save body is factored into `protocol_keri_decoder_feed_pair()`.

2. Require the two frames in the shift register to carry the same ID. Keri has no parity and no
   checksum, so frame-to-frame agreement is the only integrity check the format admits. A marginal
   capture is mis-sliced differently in each frame — which is exactly why the wrong values varied
   between reads of one card — so this turns a marginal read into a failed read rather than a
   credential. This needs the register to hold two whole frames, so `KERI_ENCODED_DATA_SIZE` goes
   from 13 to 16 bytes; at 13 the second frame's ID would have been a three-byte read past the end
   of the buffer. The preamble is still required twice — the check is added, not traded.

Costs 44 bytes of RAM. The shift register needs 128 bit periods rather than 104 before it can
produce a first decode, which cannot regress a read: on `dev` a Keri tag does not read at all. The
wider buffers also align every preamble load — at the old 13-byte stride, `*(uint32_t*)&data[…]`
was a misaligned read on three buffers in four.

It also drops `KERI_PREAMBLE_DATA_SIZE`, whose only use was the old buffer sizing, and the two
commented out `bit_lib_get_bit` lines copied from `protocol_indala26.c`. The latter were not merely
dead: the bit 60 test would have rejected this very tag (encoded bits 32–63 are `0x80003039`).

# Verification

Needs a blank T5577 and nothing else — the Flipper writes the test card itself.

1. `rfid write Keri 80003039` — internal ID 12345, FC 1, Card 12544.
2. `rfid read indala`.
3. `rfid raw_read psk /ext/lfrfid/k.psk.raw`, then `rfid raw_analyze /ext/lfrfid/k.psk.raw`.
   `raw_analyze` has no `validate_count` gate, so it shows what the decoder actually produced.

Measured on one Flipper Zero, one T5577, same tag throughout:

| | `dev` @ 7f0b6e1c | this branch |
|---|---|---|
| `rfid read indala`, 20 s | nothing, 3 of 3 | `Keri 80003039` / `Internal ID: 12345` / `FC: 1` / `Card: 12544`, 3 of 3 |
| `rfid raw_analyze` | 3 wrong credentials and 1 `not found` in 4 attempts | correct, 3 of 3 |

The wrong values vary between captures of the same card:

```
Keri [C0 00 38 3D]   Internal ID: 1073756221   FC: 3   Card: 143616
Keri [80 00 38 3D]   Internal ID: 14397        FC: 1   Card: 143616
Keri [C0 00 30 3D]   Internal ID: 1073754173   FC: 3   Card: 12544
```

Note the third: the card number is right and the facility code is not, which is harder to notice
than a value that is wrong throughout.

Also holds for a card written by a Proxmark3 (`lf keri clone -t i --cn 12345`), which lays the
same 64-bit cycle into the T5577's blocks rotated by three bits.

Round trip on this branch: `rfid write Keri 80001092` reads back as `Internal ID: 4242`.

Unit tests: `./fbt FIRMWARE_APP_SET=unit_tests`, two of them.

- `test_lfrfid_protocol_keri_read_simple` replays one frame of a captured Keri tag and asserts
  `80 00 30 39`. It fails on `dev` — uncorrected those runs total 66 bits, the preamble never
  aligns, and nothing decodes at all.
- `test_lfrfid_protocol_keri_read_mismatched_frames` alternates that frame with one whose 2H/6L
  run pair reads as 3H/5L, carrying `0x80003839`. Each frame decodes on its own, so only
  frame-to-frame disagreement can reject the alternation; without change 2 it reads out as
  `80 00 30 39`.

No new decodes against an Idteck PSK tag or four unrelated LF captures (2 PSK, 2 ASK).

# Author checklist

- [x] I own the code I'm submitting or have code owner's permission to submit it.

# AI usage disclosure

- Partially AI assisted - clarify below which parts were AI assisted and briefly explain what they do.

Claude and Gemini/Antigravity were used in the underlying research that surfaced this bug. Claude was used in locating the bug and drafting the fix and unit tests. It was also used to review the change and in drafting this PR and the accompanying commit message. The change adds a second pair of decode buffers correcting the PSK slicer's pulse-width skew, widens the decoder's shift register to hold two frames, and requires both frames to match ID before a read gets accepted. The added tests check that a valid frame passes, and two that differ are rejected. All AI code and content has been human reviewed and edited.

# Reviewer Checklist (Don't fill this out, and don't remove it from the template)


turbospok (approved):
Thanks! Will be merged soon

